PR #489, opened 2025-02-18
Add PoC under dotnet/samples that uses AIO SDK without code generation

koepalex:
Adding sample that simulates process control (write of simple type, call process control action) dynamically registered at runtime.


dotnet/samples/NoCodeGenServer/NuGet.config
ryanwinter, line 1: I dont think we need this file, the dependencies should be pulled locally or from the default Nuget feed.

dotnet/samples/NoCodeGenServer/NoCodeGenServer.csproj
```diff
@@ -0,0 +1,33 @@
...
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\Azure.Iot.Operations.Mqtt\Azure.Iot.Operations.Mqtt.csproj" />
+    <ProjectReference Include="..\..\src\Azure.Iot.Operations.Protocol\Azure.Iot.Operations.Protocol.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <!--PackageReference Include="Azure.Iot.Operations.Protocol" Version="0.8.0" /-->
```
ryanwinter, line 16: remove the commented out lines

dotnet/samples/NoCodeGenServer/Program.cs
```diff
@@ -0,0 +1,136 @@
...
+};
+
+var sessionClientOptions = new MqttSessionClientOptions
+{
+    EnableMqttLogging = true,
+};
+
+var mqttClient = new MqttSessionClient(sessionClientOptions);
+
+var connectionString = configuration.GetConnectionString("Default");
```
ryanwinter, line 37: We have moved away from connection string in our samples.

Let align with using the .env file: https://github.com/Azure/iot-operations-sdks/blob/main/dotnet/samples/README.md#run-a-sample

and then load from environment variables:
```
        MqttConnectionSettings settings = MqttConnectionSettings.FromEnvVars();
```

dotnet/samples/NoCodeGenServer/appsettings.json
```diff
@@ -0,0 +1,14 @@
...
+  "Logging": {
+    "LogLevel": {
+      "Default": "Debug",
+    },
+    "Console": {
+      "FormatterName": "simple",
+      "singleLine": "true"
+    }
+  },
+  "ConnectionStrings": {
```
ryanwinter, line 11: Can remove this once using FromEnvVars

ryanwinter (changes requested):
Can you create a short readme? Just containing:
1. One sentence on what this samples does / is for
2. How to run it, which looks like you run the program, and then use the await-response.sh and send-requests.sh scripts, or maybe these files are unrelated?
3. What to look for to understand if the step above was successful or not.

PR #607, opened 2025-03-17
Add ADR for large message chunking in MQTT protocol

maximsemenov80:
ADR for Large Message Chunking

[Transcript March 24, 2025](https://microsoft-my.sharepoint.com/:w:/p/maximsemenov/EQgpnmTY9fVIrfMA1at0u8kBjWKD_efmQe_HDqVgzcvgAA?e=PKb4Ea)

### DIFF
--- a/doc/dev/adr/0023-large-message-chunking.md
+++ b/doc/dev/adr/0023-large-message-chunking.md
@@ -14,7 +14,7 @@ We will implement sdk-level message chunking as part of the Protocol layer to tr
 
 **The chunking mechanism will**:
 
-- Be enabled/disabled by a configuration setting.
+- Be part of the protocol starting some version. [^1]
 - Use standardized user properties for chunk metadata. The `__chunk` user property will contain a colon-separated string with chunking metadata: ```<messageId>:<chunkIndex>:<totalChunks>:<checksum>```. The string will include:
   - `messageId` - UUID string in the 8-4-4-4-12 format, present for every chunk.
   - `chunkIndex` - unsigned 32 bit integer in decimal format, present for every chunk;
@@ -76,6 +76,12 @@ Chunking will use SHA-256 for checksum calculation.
 ### Compatibility
 
 - Non-chunking-aware clients will receive individual chunks as separate messages. Chunks reassembly could be implemented on the application side, given described above chunking implementation is known to the application author.
+
+[^1]: The group decided that supporting a specific protocol version should imply mandatory support for chunking, with no option to opt out, to simplify compatibility and avoid the complexity of feature flags or negotiation mechanisms.
+This approach ensures that all clients and servers claiming compliance with a given protocol version can reliably handle chunked messages, reducing ambiguity and potential errors.
+It was noted that making chunking always available (though not always used) streamlines the design and avoids the need for additional feature negotiation logic.
+The team acknowledged that while this is a stricter stance, it keeps the protocol simpler, especially for constrained devices, and avoids layering new features on top of feature negotiation.
+There was recognition that this approach means any implementation supporting the new protocol version must handle chunking, and rejecting chunked messages would make it non-compliant.
 
 ## Appendix
 


doc/dev/adr/0020-large-message-chunking.md
```diff
@@ -0,0 +1,100 @@
...
+The MQTT protocol has inherent message size limitations imposed by brokers and network constraints. Azure IoT Operations scenarios often require transmitting payloads that exceed these limits (e.g., firmware updates, large telemetry batches, complex configurations). Without a standardized chunking mechanism, applications must implement their own fragmentation strategies, leading to inconsistent implementations and interoperability issues.
+
+## Decision
+We will implement sdk-level message chunking as part of the Protocol layer to transparently handle messages exceeding the MQTT broker's maximum packet size.
+
+**The chunking mechanism will**:
+- Be enabled/disabled by a configuration setting.
+- Use standardized user properties for chunk metadata:
+   - The `__chunk` user property will contain a JSON object with chunking metadata.
+   - The JSON structure will include:
```
timtay-microsoft, line 16: I remember getting some pushback from @vaavva about introducing a JSON library dependency into our protocol layer. 

To avoid that, it may be easier to just set these 5 fields as 5 MQTT user properties rather than putting them all in one MQTT user property

doc/dev/adr/0020-large-message-chunking.md
```diff
@@ -0,0 +1,100 @@
...
+
+**Implementation layer**:
+- Chunking will be implemented as middleware in the Protocol layer between serialization and MQTT client.
+ ```
+ Application → Protocol Layer (Serialization) → Chunking Middleware → MQTT Client → Broker
+ ```
+- This makes chunking transparent to application code and compatible with all serialization formats.
+
+- Sending Process:
+  - When a payload exceeds the maximum packet size, the midlware intercepts it before transmission
```
timtay-microsoft, line 41: nit

```suggestion
  - When a payload exceeds the maximum packet size, the middleware intercepts it before transmission
```

doc/dev/adr/0020-large-message-chunking.md
```diff
@@ -0,0 +1,100 @@
...
+  - Error propagation to application code
+- **Performance Optimization:**
+  - Concurrent chunk transmission
+  - Efficient memory usage during reassembly
+- **Security:**
+  - Validate message integrity across chunks and prevent chunk injection attacks (see Checksum Algorithm Options for MQTT Message Chunking in the Appendix)
+
+# Appendix
+
+## Chunk Timeout Mechanism Options
```
timtay-microsoft, line 68: Is the idea that we would offer both of these options? Or is this a forum for deciding on which of the two to offer?

doc/dev/adr/0020-large-message-chunking.md
timtay-microsoft: To clarify, the pushback I mentioned was about an orthogonal change (error modeling in headers).

I'm fine with using JSON here because System.Text.Json isn't a hard dependency for us to take, but double check with Val that the rust folks are on board. If they aren't, we may need to use some other serialization method here.

doc/dev/adr/0020-large-message-chunking.md
timtay-microsoft: Approach 1 seems fine to me, then. I don't think the average chunking user cares about the time between chunks, but does care about the overall time to receive the re-assembled message

doc/dev/adr/0020-large-message-chunking.md
```diff
@@ -0,0 +1,100 @@
...
+
+2. Sliding Timeout Window
+   - Reset the timeout each time a new chunk arrives
+   - Only expire the chunked message if there's a long gap between chunks
+   - **Pros**: Tolerates varying network conditions and delivery rates
+   - **Cons**: Could keep resources allocated for extended periods
+
+## Checksum Algorithm Options for MQTT Message Chunking
+
+1. MD5
```
timtay-microsoft, line 84: We could consider asking users to inject their algorithm of choice so that our SDK doesn't choose one for them

doc/dev/adr/0020-large-message-chunking.md
```diff
@@ -0,0 +1,100 @@
...
+Proposed
+
+## Context
+The MQTT protocol has inherent message size limitations imposed by brokers and network constraints. Azure IoT Operations scenarios often require transmitting payloads that exceed these limits (e.g., firmware updates, large telemetry batches, complex configurations). Without a standardized chunking mechanism, applications must implement their own fragmentation strategies, leading to inconsistent implementations and interoperability issues.
+
+## Decision
+We will implement sdk-level message chunking as part of the Protocol layer to transparently handle messages exceeding the MQTT broker's maximum packet size.
+
+**The chunking mechanism will**:
+- Be enabled/disabled by a configuration setting.
```
cartertinney, line 13: Is this actually necessary? It seems like it could be automatic. If the, say, the `TelemetryMessage` payload is too large - just chunk. What's the harm in that? Removes an entire class of conditional error, and allows us to assume input correctness.

What is the argument for toggling it on and off?

doc/dev/adr/0020-large-message-chunking.md
cartertinney: This would all just end up being internally used, right? If so, I don't care - we can just use the JSON library internally. I would only have hesitation if it crossed API boundary, which I can't see why it would.

However, even then, given the pre-defined structure, we could just deserialize it to a struct that mirrors the JSON 1:1.

doc/dev/adr/0020-large-message-chunking.md
```diff
@@ -0,0 +1,100 @@
...
+- Use standardized user properties for chunk metadata:
+   - The `__chunk` user property will contain a JSON object with chunking metadata.
+   - The JSON structure will include:
+     ```json
+     {
+       "messageId": "unique-id-for-chunked-message",
```
cartertinney, line 19: How is this ID generated? UUID I assume?

doc/dev/adr/0020-large-message-chunking.md
```diff
@@ -0,0 +1,100 @@
...
+       "timeout" : "00:00:10",
+       "totalChunks": 5,
+       "checksum": "message-hash"
+     }
+     ```
+   - `messageId, chunkIndex, timeout` - present for every chunk; `totalChunks, checksum` - present only for the first chunk.
+
+**Chunk size calculation**:
+- Maximum chunk size will be derived from the MQTT CONNECT packet's Maximum Packet Size.
+- A static overhead value will be subtracted from the Maximum Packet Size to account for MQTT packet headers, topic name, user properties, and other metadata.
```
cartertinney, line 30: 1) I don't like this as a configuration option.
2) We can't guarantee correctness even with the option, and I don't think we should lock out sufficiently large user properties by design.
3) Header size could fluctuate, so a static value seems insufficient.

While difficult, it would seem to me that we should attempt to do it dynamically. A static value to start with might be a good first step, but I don't like baking it into the config options and setting it for the life of the envoy.

There might also be an argument for handling the edge case of header offset entirely reactively in the name of only handling this (unlikely) case lazily in order to reduce computation overhead in the general case (assuming there is client-side validation).

doc/dev/adr/0020-large-message-chunking.md
```diff
@@ -0,0 +1,100 @@
...
+- The overhead size will be configurable, large enough to simplify calculations while ensuring we stay under the broker's limit.
+
+**Implementation layer**:
+- Chunking will be implemented as middleware in the Protocol layer between serialization and MQTT client.
```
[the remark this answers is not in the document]
cartertinney: I would hope this is internal and entirely abstracted. I think "middleware" overstates the complexity, or at least is overly prescriptive of the design.

doc/dev/adr/0020-large-message-chunking.md
```diff
@@ -0,0 +1,100 @@
...
+ ```
+ Application → Protocol Layer (Serialization) → Chunking Middleware → MQTT Client → Broker
+ ```
+- This makes chunking transparent to application code and compatible with all serialization formats.
+
+- Sending Process:
+  - When a payload exceeds the maximum packet size, the middleware intercepts it before transmission
```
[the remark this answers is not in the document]
cartertinney: Why would we want a chunk-timeout that's different from the other timeouts for messages? Seems like conjuring new rules and edge cases out of nowhere. Why wouldn't a chunk be valid for the time the underlying MQTT message is valid for?

doc/dev/adr/0020-large-message-chunking.md
```diff
@@ -0,0 +1,100 @@
...
+  - The message is split into fixed-size chunks (with potentially smaller last chunk)
+  - Each chunk is sent as a separate MQTT message with the same topic but with chunk metadata.
+  - Effort should be made to minimize user properties copied over to every chunk: first chunk will have full set of original user properties and the rest only thoses that are neccessary to reassamble original message (ex.: ```$partition``` property to support shared subscriptions:).
```
cartertinney, line 44: Even more reason not to use a static header subtraction size across all chunks then

doc/dev/adr/0020-large-message-chunking.md
```diff
@@ -0,0 +1,100 @@
...
+  - Error propagation to application code
+- **Performance Optimization:**
+  - Concurrent chunk transmission
+  - Efficient memory usage during reassembly
+- **Security:**
+  - Validate message integrity across chunks and prevent chunk injection attacks (see Checksum Algorithm Options for MQTT Message Chunking in the Appendix)
+
+# Appendix
+
+## Chunk Timeout Mechanism Options
```
cartertinney, line 68: Again, why would a chunk timeout differently from the underlying MQTT message? Seems unnecessarily complex.

doc/dev/adr/0020-large-message-chunking.md
```diff
@@ -0,0 +1,100 @@
...
+**The chunking mechanism will**:
+- Be enabled/disabled by a configuration setting.
+- Use standardized user properties for chunk metadata:
+   - The `__chunk` user property will contain a JSON object with chunking metadata.
+   - The JSON structure will include:
+     ```json
+     {
+       "messageId": "unique-id-for-chunked-message",
+       "chunkIndex": 0,
+       "timeout" : "00:00:10",
```
vaavva, line 21: what is the format of this? Should we just have this be an int that represents seconds or milliseconds instead of having something to parse?

doc/dev/adr/0020-large-message-chunking.md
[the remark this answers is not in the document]
cartertinney: Right, but each chunk has its own MQTT timeout, so what's the need to introduce a full timeout for the entire chunks? 

Suppose we're using an RPC command invocation, which has a timeout value. The command in total, regardless of how many chunks it has, is invalid after this set amount of time. This value then gets used as the message expiry interval for MQTT (arguably it shouldn't, but it does). Why then, would you ever want to set a chunk timeout that's any different from that value? The Command Request timeout already represents the timeout of the Command, and the Command is made up 1..n chunks. The number of chunks does not change when the command times out.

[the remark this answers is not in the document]
maximsemenov80: Carter is right:
_[MQTT-3.3.2-6] | The PUBLISH packet sent to a Client by the Server MUST contain a Message Expiry Interval set to the received value minus the time that the message has been waiting in the Server._

we should than use this value as time left for the rest of the chunks to be delivered.

doc/dev/adr/0020-large-message-chunking.md
timtay-microsoft: > so the subscriber should be the SDK as well

We can't assert that. If the receiver doesn't support chunking, then it will receive gibberish when the sender chunks (as opposed to a clear error saying "message too large to send"). And if the sender chunks without the user knowing, that error would be very confusing. I'd generally rather the user opt in to chunking.

doc/dev/adr/0020-large-message-chunking.md
```diff
@@ -0,0 +1,142 @@
...
+- The overhead size will be configurable, large enough to simplify calculations while ensuring we stay under the broker's limit.
+
+**Chunk Timeout Mechanism**
+
+> [MQTT-3.3.2-6] | The PUBLISH packet sent to a Client by the Server MUST contain a Message Expiry Interval set to the received value minus the time that the message has been waiting in the Server.
+
+The receiving client uses the Message Expiry Interval from the first chunk as the timeout period for collecting all remaining chunks of the message.
+
+Edge case:
+- Since the Message Expiry Interval is specified in seconds, chunked messages may behave differently than single messages when the expiry interval is very short (e.g., 1 second remaining). For a single large message, the QoS flow would complete even if the expiry interval expires during transmission. However, with chunking, if the remaining expiry interval is too short to receive all chunks, the message reassembly will fail due to timeout.
```
vaavva, line 47: for other message expiry calculations, we always round partial seconds up, never down - I imagine doing something similar here should maintain acceptable behavior.

To your statement about `the QoS flow would (not) complete` for the chunking scenario, what exactly do you mean by this? Just that the message might not be received by the end application if all chunks aren't delivered in time, or is there some ramification about the QoS flow not completing - we would definitely still need to ack all messages for the chunking scenario

doc/dev/adr/0020-large-message-chunking.md
```diff
@@ -0,0 +1,142 @@
...
+        Note over Receiver: All chunks received
+        Receiver->>Receiver: Verify checksum
+        Note right of Receiver: SHA-256 or<br/>custom algorithm
+        Receiver->>Receiver: Reassemble message
+        Note over Receiver: Process complete message
+    else Failure: Timeout
+        Note over Receiver: Message Expiry Interval exceeded
+        Receiver->>Receiver: Timeout occurred
+        Receiver->>Receiver: Cleanup buffers
+        Note over Receiver: Notify application:<br/>ChunkTimeoutError
```
vaavva, line 128: I think for these error cases, we would just log the error and ignore the received message, similar to any other invalid message that we've received. There's no action the application can take

vaavva, line 128: we will need some new rpc errors for communicating chunking issues to the invoker in the command response

doc/dev/adr/0020-large-message-chunking.md
```diff
@@ -0,0 +1,142 @@
...
+  - The reconstructed message is then processed as a single message by the application.
+- Receiving Failures:
+  - Message timeout interval ended before all chunks received.
+  - Reassembly buffer size limit reached before all chunks received.
+  - Calculated checksum does not match checksumm from chunk metadata.
+
+**Configuration settings:**
+- Enable/Disable
+- Overhead size
+- Reassembly buffer size limit
```
vaavva, line 74: Is this a buffer size per message being reassembled, or is this a size limit for all messages being reassembled? I understand why we'd want to minimize the amount of space taken up on a device waiting for all chunks to come in, but it seems like this will lead to _very_ unpredictable behavior about when messages do or don't get received and make it seem like we aren't upholding our message loss guarantees.
If it's for a single message, we might be better off having a limit on the number of chunks that are allowed, that way we can abort after receiving the first message instead of waiting for the buffer to fill up (maybe that's effectively what we have here already)

This raises another interesting question as well: is the goal of this document to get around the MQTT packet size restriction, or is this intended for supporting constrained devices? If the goal is just the first, then I'd almost argue that there shouldn't be a buffer size limit at all

doc/dev/adr/0020-large-message-chunking.md
vaavva: If this config stays, would it be per envoy? Per message seems excessive, and anything larger in scope doesn't make sense, since different receivers might be configured differently.
Second, if this is an option, I would say it should only be an option on the sending side - the receiving side will either not know what to do with the chunk messages (older version of the SDK or receiver that doesn't use the SDK), or it knows what a chunk message means and might as well reassemble it instead of not getting the message at all

doc/dev/adr/0020-large-message-chunking.md
maximsemenov80: it means it does not matter if the single message expires during QoS flow execution it will complete (even if it involves whole message resend), but in case of chunks at the very end of expiration interval it is possible some tail chunks would expire before their transfer started thus creating effect of the whole (original) message expires midflight and transfer canceled.

doc/dev/adr/0023-large-message-chunking.md
```diff
@@ -0,0 +1,120 @@
...
+
+The MQTT protocol has inherent message size limitations imposed by brokers and network constraints. Azure IoT Operations scenarios often require transmitting payloads that exceed these limits (e.g., firmware updates, large telemetry batches, complex configurations). Without a standardized chunking mechanism, applications must implement their own fragmentation strategies, leading to inconsistent implementations and interoperability issues.
+
+## Decision
+
+We will implement sdk-level message chunking as part of the Protocol layer to transparently handle messages exceeding the MQTT broker's maximum packet size.
+
+**The chunking mechanism will**:
+
+- Be enabled/disabled by a configuration setting.
```
jspaith, line 17: What is the default if not set?  I think it should be disabled for compat concerns esp. if a non-SDK is receiving the PUBLISH messages.

doc/dev/adr/0023-large-message-chunking.md
```diff
@@ -0,0 +1,120 @@
...
+- Use standardized user properties for chunk metadata. The `__chunk` user property will contain a colon-separated string with chunking metadata: ```<messageId>:<chunkIndex>:<totalChunks>:<checksum>```. The string will include:
+  - `messageId` - UUID string in the 8-4-4-4-12 format, present for every chunk.
+  - `chunkIndex` - unsigned 32 bit integer in decimal format, present for every chunk;
+  - `totalChunks` - unsigned 32 bit integer in decimal format, present only for the first chunk.
+  - `checksum` - SHA-256 hash in hexadecimal format (64 characters long), present only for the first chunk.
+
+**Chunk size calculation:**
+
+- Maximum chunk size will be derived from the MQTT CONNECT packet's Maximum Packet Size.
```
jspaith, line 26: * I like having the maximum size based off CONNACK.  
  * Lower prio - it maybe OK to punt in v1 - but you should think whether you want this to also be configured via a knob, where the size would be min(CONNACK calculation, knob setting).
  * The scenario is that we could imagine say a low-end device that only had some small amount of RAM, where we only want to send is smaller chunks even if the MQTT broker allowed larger sizes.

doc/dev/adr/0023-large-message-chunking.md
```diff
@@ -0,0 +1,120 @@
...
+
+We will implement sdk-level message chunking as part of the Protocol layer to transparently handle messages exceeding the MQTT broker's maximum packet size.
+
+**The chunking mechanism will**:
+
+- Be enabled/disabled by a configuration setting.
+- Use standardized user properties for chunk metadata. The `__chunk` user property will contain a colon-separated string with chunking metadata: ```<messageId>:<chunkIndex>:<totalChunks>:<checksum>```. The string will include:
+  - `messageId` - UUID string in the 8-4-4-4-12 format, present for every chunk.
+  - `chunkIndex` - unsigned 32 bit integer in decimal format, present for every chunk;
+  - `totalChunks` - unsigned 32 bit integer in decimal format, present only for the first chunk.
```
jspaith, line 21: * Do we care about the streaming scenario, where if an initiator is sending a video it won't know `totalChunks` until its done?
  * I've gone back and forth on this when I was thinking about it in 2022.  You need an end-marker for when stream is done.  And it may complicate the expiry logic, since the expiry is unknown.
  * At minimum I think you should at least see "How can I add streaming to this protocol?" even if we don't break it.

doc/dev/adr/0023-large-message-chunking.md
```diff
@@ -0,0 +1,120 @@
...
+The MQTT protocol has inherent message size limitations imposed by brokers and network constraints. Azure IoT Operations scenarios often require transmitting payloads that exceed these limits (e.g., firmware updates, large telemetry batches, complex configurations). Without a standardized chunking mechanism, applications must implement their own fragmentation strategies, leading to inconsistent implementations and interoperability issues.
+
+## Decision
+
+We will implement sdk-level message chunking as part of the Protocol layer to transparently handle messages exceeding the MQTT broker's maximum packet size.
+
+**The chunking mechanism will**:
+
+- Be enabled/disabled by a configuration setting.
+- Use standardized user properties for chunk metadata. The `__chunk` user property will contain a colon-separated string with chunking metadata: ```<messageId>:<chunkIndex>:<totalChunks>:<checksum>```. The string will include:
```
jspaith, line 18: * Speaking of versions - should this include a version in it of the chunking version?
  * How does mRPC command/response handle versions of the mRPC protocol layer itself (which is different than the customer mRPC client/servers having the version field)?  Can we leverage those concepts?
  * As one case I'm imagining, we could imagine(!) a v1 of this protocol requires sender to know size up front but then we relax in a v2 to support streaming.

doc/dev/adr/0023-large-message-chunking.md
```diff
@@ -0,0 +1,120 @@
...
+  - `messageId` - UUID string in the 8-4-4-4-12 format, present for every chunk.
+  - `chunkIndex` - unsigned 32 bit integer in decimal format, present for every chunk;
+  - `totalChunks` - unsigned 32 bit integer in decimal format, present only for the first chunk.
+  - `checksum` - SHA-256 hash in hexadecimal format (64 characters long), present only for the first chunk.
```
jspaith, line 22: * Does anywhere else in mRPC use a `checksum`?
  * If yes, then makes sense to keep here I guess.  
  * If no, do we need it here?  We rely on TCP to do checksum checks for us.  And this checksum would still only check the payload, the headers could (in theory) get corrupted still.

doc/dev/adr/0023-large-message-chunking.md
```diff
@@ -0,0 +1,120 @@
...
+  - When all chunks for a message ID are received, they are reassembled in order and message checksum verified (see Checksum Algorithm Options for MQTT Message Chunking).
+  - The reconstructed message is then processed as a single message by the application.
+- Receiving Failures:
+  - Message timeout interval ended before all chunks received.
+  - Calculated checksum does not match checksumm from chunk metadata.
+
+**Configuration settings:**
+- Enable/Disable
+
+### Implementation Considerations
```
jspaith, line 62: *Outside the scope of this ADR and potentially post-2510 even.  But it's on my mind so recording here.*

* We need to think about who the customers of this can/should be.
  If SDKs do most of work it should be an easy discussion.  But on my mind:
  * Schema Registry team - maybe.  Even on a tiny configuration size, max message is still 4MB which I think should cover vast majority of schemas.  Though maybe not - OPC UA schemas get big.
  * Tinykube | WASM - 100%!  These WASM modules can be really large - like dozens of MB.  And since both the Tinykube server and client are owned by Microsoft, should be really easy to add this.
  * Dataflows - ??.  I don't know actually since I don't have a sense of how large messages we're getting.
  * OPC UA / connectors - ??.  It depends on what we do with Dataflows as Dataflows is one of their main customers.

doc/dev/adr/0023-large-message-chunking.md
```diff
@@ -0,0 +1,120 @@
...
+
+- Non-chunking-aware clients will receive individual chunks as separate messages. Chunks reassembly could be implemented on the application side, given described above chunking implementation is known to the application author.
+
+## Appendix
+
+### Message Flow Diagram
+
+```mermaid
+sequenceDiagram
+    participant Sender as Sending Client
```
jspaith, line 86: * Retry on the sender?
  * It looks like the sending client has no way of knowing whether its upload was actually processed by the receiver.
  * This is the same as any MQTT PUBLISH / telemetry of course - getting a PUBACK only tells PUBLISHER so much.
  * The question is - is that OK for this case?  Or do we want to let clients be more robust?  
  * So some sort of mRPC type response-topic construct - maybe sent in the 1st message - where the caller can indicate success|failure?
  * We should understand scenarios though - we may indeed not need this.  
    * Like for Tinykube, a client will be initiating "please TK server download for me this giant WASM module" and then the TK client will listen for chunked-response.  
    * So if chunked-response doesn't come in time, then TK client would quest retry the "give me WASM module" rather than relying on it to retry.
  * So something to think about, and if we don't think we need it I think its worth calling out in design that we don't need it & why.

doc/dev/adr/0023-large-message-chunking.md
```diff
@@ -0,0 +1,120 @@
+﻿# ADR 23: Large Message Chunking in MQTT Protocol
```
jspaith, line 1: Nit - it looks like JD has an ADR 23 out for PR so this will need to be bumped at least by 1.

doc/dev/adr/0023-large-message-chunking.md
maximsemenov80: I agree, integrity check that travels with the data it's protecting can be spoofed by an attacker who can modify both.

doc/dev/adr/0023-large-message-chunking.md
maximsemenov80: I heard word streaming, but no details at all. My thoughts are: in the stream with unknown end each piece of data is meaningful like telemetry or live video feed and doesn't have to be assembled into the bigger object, as soon as you require it to be part of the whole you have to know what is the whole and then it is regular chunks with known full size.

jspaith: That makes sense.  It might be worth calling this out as a non-goal *for this protocol* though, but otherwise don't think you need changes here.

doc/dev/adr/0023-large-message-chunking.md
jspaith: There's always possibility of timeout - I think MQ sets messages with 24 hour timeouts by default though I may be misremembering that.  

Or we could imagine a client saying "timeout = 2 minutes" for say the entire operation, since maybe if this fails they want to retry or at least signal an error to the caller / user in a more timely manner?  For that I think you need an message to a response-topic on initiator.

maximsemenov80: I'm a bit confused after our last meeting (now when chunking is unconditional and automatic) say I know my message is not going to be chunked then QOS give me all I need, on the other hand if I know that my message will be chunked, RPC with underlying chunking MQTT client gives desired level of the delivery control on the sender side (probably with some extended information on specific chunking failure in the InnerException):

1. RPC calls mqttClient.PublishAsync(completeMessage)
2. Chunking layer splits into chunks and sends them
3. Something goes wrong with chunking (partial failure)
4. RPC only gets the final result: success or failure of the entire message

timtay-microsoft: If I understand the question correctly, the answer is that the executor side's chunking layer would time out waiting for the final chunk of the RPC call and would not notify the user that any RPC was invoked. The RPC layer would only "count" an RPC call that has received all chunks successfully

PR #643, opened 2025-03-25
Test loading action directly

ryanwinter:


ryanwinter:
Not the most durable solution, I'm currently using a PAT, but this will unblock our pipeilnes while I investigate another solution.

avishekpant (approved):

cartertinney (approved):

PR #649, opened 2025-03-26
[Codegen] Target .NET 9

timtay-microsoft:


dotnet/samples/Protocol/TestEnvoys/Memmon/GetRuntimeStatsResponsePayloadSerialization.g.cs
```diff
@@ -17,11 +17,11 @@ public partial class GetRuntimeStatsResponsePayload
     {
       ""name"": ""diagnosticResults"",
       ""type"": {
-        ""type"": ""map"",
-        ""values"": {
-          ""type"": ""string""
-        },
-        ""default"": {}
+""type"": ""map"",
```
clecompt-msft, line 20: nit: What happened to the indentation here?

clecompt-msft (approved):

dotnet/samples/Protocol/TestEnvoys/Memmon/GetRuntimeStatsResponsePayloadSerialization.g.cs
timtay-microsoft: I'm not entirely sure, but this happened after running code gen so it is fine to be checked in

PR #650, opened 2025-03-27
Review dotnet package install instructions to use NuGet and ADO feeds

ryanwinter:


timtay-microsoft:
That preview feed is still used by customers using the connector package. Are you sure we should remove it?

ryanwinter:
> That preview feed is still used by customers using the connector package. Are you sure we should remove it?

Perhaps have two options? We also have separate directions in the azure-iot-operations-preview repo.

PR #654, opened 2025-03-28
[Rust][MQTT] ConnectionSettings tests + fixes

cartertinney:
* Reimplemented `MqttConnectionSettingsBuilder::from_file_mount()` to conform to Builder Pattern expectations
* Reduced number of allocations in `MqttConnectionSettingsBuilder::from_file_mount()`
* Re-enabled unit-tests for `MqttConnectionSettingsBuilder::from_file_mount()` without the use of `unsafe` blocks
* Added missing unit-tests for `MqttConnectionSettingsBuilder::from_environment()`
* Fixed a bug where `MqttConnectionSettingsBuilder::from_environment()` and `MqttConnectionSettingsBuilder::from_file_mount()` set default values incorrectly
* Updated existing tests to check that `MqttConnectionSettingsBuilder::from_environment()` and `MqttConnectionSettingsBuilder::from_file_mount()`  set the correct default values for non-provided values.
* Added additional test helpers to facilitate use of filesystem in unit-tests
* Added additional warning logs indicating likely misconfiguration
* Fixed a bug where use of `MqttConnectionSettingsBuilder::from_file_mount()` set the wrong default value for `clean_start`
* Fixed a bug where `cert_file` could be provided without `key_file` on `MqttConnectionSettingsBuilder`
* Simplified `MqttConnectionSettingsBuilder::validate()`


rust/azure_iot_operations_mqtt/src/connection_settings.rs
```diff
@@ -330,140 +338,165 @@ fn string_from_environment(key: &str) -> Result<Option<String>, String> {
...
+        }
+
+        fn path(&self) -> &Path {
+            self.dir.path()
+        }
+    }
 
     #[test]
-    fn test_connection_settings_empty_hostname() {
+    fn minimum_configuration() {
```
vaavva, line 393: I think we usually start test functions with `test_` to help distinguish them from test helper functions

rust/azure_iot_operations_mqtt/src/connection_settings.rs
```diff
@@ -330,140 +338,165 @@ fn string_from_environment(key: &str) -> Result<Option<String>, String> {
...
-        let connection_settings_builder_result = MqttConnectionSettingsBuilder::default()
+        // The client id still must be provided if clean_start is true
+        let result = MqttConnectionSettingsBuilder::default()
+            .hostname("test_host".to_string())
+            .clean_start(true)
+            .build();
+        assert!(result.is_err());
+
+        // But an empty client_id is allowed if clean_start is true
+        // NOTE: Not sure why though. Perhaps this is undesirable.
```
vaavva, line 435: I think this might be a holdover from before we required client_id across the languages - before we were just ensuring that rumqttc wouldn't give us a problem with the client_id that was provided. If we now always want to require a client_id across the languages, it should always be non-empty

vaavva (approved):

rust/azure_iot_operations_mqtt/src/connection_settings.rs
cartertinney: Yeah, I'm trying to move away from that - it's very redundant in the test output. The macro on the top should be sufficient.

rust/azure_iot_operations_mqtt/src/connection_settings.rs
cartertinney: As per our discussion, I have elected to remove carve-out logic and handle client id consistently - it is required, and it must not be empty string.

rust/azure_iot_operations_mqtt/src/connection_settings.rs
```diff
@@ -4,6 +4,7 @@
 //! Generic MQTT connection settings implementations
 
 use std::env::{self, VarError};
+use std::path::{Path, PathBuf};
```
olivakar, line 7: why path buf

rust/azure_iot_operations_mqtt/src/connection_settings.rs
```diff
@@ -181,145 +211,120 @@ impl MqttConnectionSettingsBuilder {
...
+                // NOTE: See the warning section father below in the function for an
+                // explanation of why this isn't an error.
+                log::warn!("AEP_CONFIGMAP_MOUNT_PATH is not set in environment");
+                (None, None, None, None)
             }
         };
 
-        // Return builder with settings
+        // --- Mount 2: BROKER_SAT_MOUNT_PATH ---
+        // NOTE: This will be moved to be part of Mount 1 in the future.
```
olivakar, line 261: is this decided ?

RLeclair (approved):
LGTM!

olivakar (approved):
I have some questions - we can chat later. For now LGTM.

rust/azure_iot_operations_mqtt/src/connection_settings.rs
cartertinney: `PathBuf` can be a source of truth that we can use slice references to (Both `Path`s and `str`s) that reduce the number of allocations necessary. Before, all the `String` concatenations and `format!`s were requiring allocations.

PR #656, opened 2025-03-28
Add pull_request_target so the workflows are run successfully in fork

ryanwinter:
…d not the fork to access the action secret.


cartertinney (approved):
LGTM

PR #663, opened 2025-03-31
[.Net] Add ADR Client.

maximsemenov80:
Autogen ADR client and add/implement public interface for generated ADR service client


dotnet/src/Azure.Iot.Operations.Services/AssetAndDeviceRegistry/IAdrServiceClient.cs
```diff
@@ -0,0 +1,55 @@
...
+
+    Task<CreateDetectedAssetResponseSchema?> CreateDetectedAssetAsync(string aepName,
+        CreateDetectedAssetRequestPayload requestPayload,
+        TimeSpan? commandTimeout = null,
+        CancellationToken cancellationToken = default);
+
+    Task<CreateDiscoveredAssetEndpointProfileResponseSchema?> CreateDiscoveredAssetEndpointProfileAsync(string aepName,
+        CreateDiscoveredAssetEndpointProfileRequestPayload requestPayload,
+        TimeSpan? commandTimeout = null,
+        CancellationToken cancellationToken = default);
```
timtay-microsoft, line 54: I think these APIs shouldn't be exposed to users. Creating assets/asset endpoint profiles is something that only the operator should be doing in this context. 

The only reason they are in the DTDL is so that the service implementation can use codegen as well.

dotnet/src/Azure.Iot.Operations.Services/AssetAndDeviceRegistry/AdrServiceClient.cs
```diff
@@ -0,0 +1,196 @@
...
+    {
+        cancellationToken.ThrowIfCancellationRequested();
+        ObjectDisposedException.ThrowIf(_disposed, this);
+
+        var additionalTopicTokenMap = new Dictionary<string, string> { { "aepName", aepName } };
+        return (await _assetEndpointProfileServiceClient.CreateDiscoveredAssetEndpointProfileAsync(requestPayload, null, additionalTopicTokenMap,
+            commandTimeout ?? _defaultTimeout, cancellationToken)).CreateDiscoveredAssetEndpointProfileResponse;
+    }
+
+    public event Func<string, AssetEndpointProfile?, Task>? OnReceiveAssetEndpointProfileUpdateTelemetry
```
timtay-microsoft, line 185: Make sure you define these events in the ```IAdrServiceClient``` interface as well.

dotnet/src/Azure.Iot.Operations.Services/Azure.Iot.Operations.Services.csproj
```diff
@@ -27,4 +27,18 @@
     <DelaySign>false</DelaySign>
   </PropertyGroup>
 
+  <ItemGroup>
+    <InternalsVisibleTo Include="Azure.Iot.Operations.Services.UnitTests" />
+    <InternalsVisibleTo Include="Azure.Iot.Operations.Connector.UnitTests" />
+  </ItemGroup>
```
timtay-microsoft, line 33: This may be a relic of a merge issue. These internalsVisibleTo declarations live elsewhere now

dotnet/src/Azure.Iot.Operations.Services/AssetAndDeviceRegistry/AdrServiceClient.cs
```diff
@@ -0,0 +1,196 @@
...
+    public async ValueTask DisposeAsync()
+    {
+        if (_disposed) return;
+        await _assetServiceClient.DisposeAsync().ConfigureAwait(false);
+        await _assetEndpointProfileServiceClient.DisposeAsync().ConfigureAwait(false);
+        GC.SuppressFinalize(this);
+        _disposed = true;
+    }
+
+    public async Task<NotificationResponse> ObserveAssetEndpointProfileUpdatesAsync(string aepName, Func<string, AssetEndpointProfile?, Task> onAssetEndpointProfileUpdateEventReceived, CancellationToken cancellationToken)
```
timtay-microsoft, line 27: It is a bit unorthodox to pass in a func to add to the event like this. The typical pattern for .NET here would look like:

```csharp
public static void main()
{
  var adrClient = new();
  adrClient.OnReceiveAssetUpdateEventTelemetry += someHandler;
  await adrClient.ObserveAssetUpdatesAsync("someAepName", "someAssetName");
}
```

That way, the user can add/change/remove handlers over time without needing to invoke any of our methods like:

 
```csharp
public static void main()
{
  var adrClient = new();
  adrClient.OnReceiveAssetUpdateEventTelemetry += someHandler;
  await adrClient.ObserveAssetUpdatesAsync("someAepName", "someAssetName");
  await Task.Delay(...);
  adrClient.OnReceiveAssetUpdateEventTelemetry += anotherHandler;
  adrClient.OnReceiveAssetUpdateEventTelemetry -= someHandler;
}
```

timtay-microsoft: We use the above pattern in places like the MQTT client layer (setting handler + calling subscribe)

dotnet/src/Azure.Iot.Operations.Services/AssetAndDeviceRegistry/IAdrServiceClient.cs
timtay-microsoft: From the discussion with the service folks, it sounds like this can stay as-is after all

Issue #488, opened 2025-02-18
[Feature Request] Surface OT device errors via mRPC protocol


koepalex:
Usecase:

As a developer writing connectors
I want to be able to communicate protocol specific errors (like `Bad_NodeIdUnknown` for OPC UA)
So that I can inform the initial caller, that the request was correct, the application (customer and connector) worked correctly but still the call wasn't successful. 

Todo:

* Extend `CommandStatusCode` with BadGateway (502)
* Add new `AkriGatewayException` allowing the connector to define a custom status message
* Extend `AkriSystenProperties` with `IsGatewayError` (proposal ReservedPrefix + "gwErr")
* Update `CommandExecutor.cmdFunc`  to catch the AkriGatewayException and pass IsGatewayError along with the status message thru GenerateAndPublishResponse method

PR #676, opened 2025-04-03
[.NET + ADR] Error modeling in headers

timtay-microsoft:


jrdouceur:
Is there ever a need for an application error code in the header of a message that contains normal application payload?  Or can we be confident that the inflexible payload schemas alluded to in [this comment](https://github.com/Azure/iot-operations-sdks/issues/488#issuecomment-2707496996) apply only when there are no errors?  If an error (either protocol-level or application-level) implies that there is no normal payload, then the MQTT payload is available for conveying detailed error information.  This would allow us to forego the second header field and the awkward packing of encoded data into a header string value.  It would also enable a minor (but critical) change to ADR 19 that would avoid modifying the payload in success cases, because the distinction between success and error would come from a header instead of which property in the payload's top-level object contains a value.

doc/dev/adr/0020-error-modeling-headers.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Samples
+
+We want to establish this pattern as the standard for communicating RPC application errors. To do so, we must either update or add a sample per-language that demonstrates handling at least one application error.
+
+For instance, we would make the counter service return an application error if the invocation specifies a negative value to increment by. This change won't include any DTDL level changes yet.
+
+## Open Questions
+
+- MQTT user properties are strings, not bytes, so is there an issue with using UTF8 encoding to convert a serialized payload to a string? Does this work for the serializers we already support (protobuf, avro, raw, etc.)
```
jrdouceur, line 96: The formats of AVRO, Protobuf, and raw are binary, so these would need to be Base64-encoded if we want to pack them into MQTT header values.

doc/dev/adr/0020-error-modeling-headers.md
```diff
@@ -0,0 +1,103 @@
...
+  - Maybe this suggests the serialization interface we use here converts object <-> string rather than object <-> bytes
+
+- MQTT user property value size limit considerations? Users may provide very large payload objects and MQTT as a protocol may not be built for that?
+
+- Do we remove support for modeling application errors in the MQTT message payload as described in (ADR 19)[./0019-codegen-user-errs.md]? No one has onboarded to this model yet
```
jrdouceur, line 101: I believe @abhipsaMisra has begun using the new DTDL user-error modeling mechanisms in the Mqtt v3 extension.

timtay-microsoft:
> Is there ever a need for an application error code in the header of a message that contains normal application payload? Or can we be confident that the inflexible payload schemas alluded to in https://github.com/Azure/iot-operations-sdks/issues/488#issuecomment-2707496996 apply only when there are no errors? If an error (either protocol-level or application-level) implies that there is no normal payload, then the MQTT payload is available for conveying detailed error information. This would allow us to forego the second header field and the awkward packing of encoded data into a header string value. It would also enable a minor (but critical) change to ADR 19 that would avoid modifying the payload in success cases, because the distinction between success and error would come from a header instead of which property in the payload's top-level object contains a value.

This is an important question. @koepalex in the scenario you highlighted in the above issue thread, will your connector be expected to send both a MQTT message payload (where you cannot add error modeling as discussed in the issue thread) + this proposed header payload?

I would personally prefer we keep payload-like objects in the MQTT message payload, but I gather that isn't feasible for your use case.

doc/dev/adr/0020-error-modeling-headers.md
timtay-microsoft: Good point. I do like that her use case allows for modeling of the error in the payload. The only question then becomes "do we care that there are 2 ways to model errors in DTDL?"

I know that we haven't really investigated modeling headers in DTDL land, but do you think it is feasible for us to support DTDL modeling of errors in either (and maybe both?) headers and MQTT payloads?

From the pure command invoker/executor side of things, it's certainly doable and not much overhead.

jrdouceur: It is certainly feasible from a technical perspective.  But having multiple ways to express error conditions runs counter to the goal (as I understand it) of ensuring commonality across applications.

doc/dev/adr/0020-error-modeling-headers.md
```diff
@@ -0,0 +1,103 @@
+# ADR 19: Modeling User Errors
+
+## Context
+
+We have received feedback from users that they would like to have RPC responses communicate application level failures in the headers of the response rather than the payload of the response. 
+
+This feedback is, in part, because some applications want to route RPC responses without reading/deserializing the entire payload. It is also partly because [some applications cannot extend or change the payload model to accommodate error handling.](https://github.com/Azure/iot-operations-sdks/issues/488#issuecomment-2707496996).
```
koepalex, line 7: `applications want to route RPC`
-->
`applications have to route RPC`

doc/dev/adr/0020-error-modeling-headers.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Decision
+
+Our SDKs will add APIs on the command executor side that mark an RPC response as an "application error" by attaching two new MQTT user properties ("AppErrCode" and "AppErrPayload") whose values are a user-defined string and a user-defined payload object (serialized to bytes and then encoded as a UTF8 string) respectively. When a user wants to indicate that an RPC call failed with an application error they must provide the "AppErrCode" value, but "AppErrPayload" is optional.
```
koepalex, line 11: Why enforcing a user-defined payload object, serialized to bytes and then encoded as UTF8, to support code gen?

doc/dev/adr/0020-error-modeling-headers.md
```diff
@@ -0,0 +1,103 @@
...
+
+In addition, users must be able to include arbitrary user properties in their command response to support this error reporting in case our standard fields are insufficient. This feature is likely in place already for all languages.
+
+On the command invoker side, we will add APIs for checking if a response was an application error and returning the error code and error data fields parsed from the MQTT message "AppErrCode" and "AppErrPayload" user properties.
+
+Similar to how our SDKs handle serializing the actual MQTT message payload, our SDKs will require the user provide the serializer for serializing/deserializing the AppErrPayload object.
```
koepalex, line 17: This is only required, if the AppErrorPayload isn't string

doc/dev/adr/0020-error-modeling-headers.md
```diff
@@ -0,0 +1,103 @@
...
+
+Other than these two new user properties, the over-the-wire behavior of our protocol won't change as a result of this decision.
+
+In order to provide a strongly-typed experience, we will also add codegen support for modeling both the possible error codes (enum-like list of possible values?) and the type of the error payload in DTDL. This modeling will be detailed in a separate ADR, though.
```
koepalex, line 21: `both the possible error codes` you mean possible header fields? Because the AppErrCode can have various values...

doc/dev/adr/0020-error-modeling-headers.md
```diff
@@ -0,0 +1,103 @@
...
+
+By convention, the value of the AppErrPayload object should be a UTF-8 encoded JSON string since MQTT user property values must be UTF-8 encoded strings.
```
koepalex, line 23: Above is state that the Error object is serialized to bytes and encoded as UTF8 string, this is mismatching with the UTF-8 encoded JSON string

doc/dev/adr/0020-error-modeling-headers.md
```diff
@@ -0,0 +1,103 @@
...
+
+    IncrementRequestPayload payload = new IncrementRequestPayload
+    {
+        IncrementValue = -1
+    };
+
+    var response = counterClient.Increment(executorId, payload).WithMetadata();
+    
+    // Check the RPC response for an application error
+    if (response.TryGetApplicationError(SomeJsonSerializer, out string? errorCode, out CustomErrorPayload? errorPayload))
```
koepalex, line 71: What is if the client use different Json(de-)Serializer?

doc/dev/adr/0020-error-modeling-headers.md
```diff
@@ -0,0 +1,103 @@
...
+    {
+        // use the error code and strongly typed custom error payload type as wanted.
+    }
+
+    // Alternatively just fetch the error code if you don't want to deserialize the payload yet.
+    if (response.TryGetApplicationError(out string? errorCode))
```
koepalex, line 77: Can we think about using something like: https://dotnet.github.io/dotNext/features/core/result.html That no matter how the "user" models errors (payload or header), there is kind of more opinionated error handling. In the current proposal the customer can just ignore the error and access the payload object

doc/dev/adr/0020-error-modeling-headers.md
timtay-microsoft: Yes. We eventually want DTDL to model these headers so that we can provide strong typing to users via codegen

doc/dev/adr/0020-error-modeling-headers.md
timtay-microsoft: I'll reword this. I'm trying to say that DTDL will model possible values of ```AppErrCode``` and will model the value of the ```AppErrPayload``` object

doc/dev/adr/0020-error-modeling-headers.md
timtay-microsoft: Good catch. I'll rewrite this.

doc/dev/adr/0020-error-modeling-headers.md
timtay-microsoft: This proposal allows for user to have different serializers for handling error models in MQTT message payloads vs error models in MQTT user properties.

doc/dev/adr/0020-error-modeling-headers.md
timtay-microsoft: > no matter how the "user" models errors (payload or header), there is kind of more opinionated error handling

I'm not sure this is feasible. Partly because a user may not have any error modeling (in payload or headers), and partly because our library allows users to model errors in payload without feeding that information to code gen. We really don't have a way to force users to do all error modeling in a standard way that our library could then parse out.

I think it is reasonable to assert that the user of an RPC invoker will know where to expect where to find an error if one existed (in headers vs in payload), but do you feel differently?

doc/dev/adr/0020-error-modeling-headers.md
timtay-microsoft: I envision that we'd do the same thing here that we did for raw byte payloads in MQTT messages where we provide a PassthroughSerializer that takes and returns a string as-is. I'll call this out in the doc, though.

doc/dev/adr/0020-error-modeling-headers.md
jrdouceur: I suspect Alex means that if the SDK understands the meaning of the header, it can determine that the response indicates a user error, and it can throw an exception or return an error value instead of returning the normal result value from the payload.  ADR 19 describes how to do this using codegen support, but if the SDKs are directly aware of user-error headers, this could be done without any need for codegen.

However, on reflection, I'm not sure how this would be compatible with [Alex's comment below](https://github.com/Azure/iot-operations-sdks/pull/676#issuecomment-2787512103) that we might need to support both valid response payloads and reported user errors in the same response message.

timtay-microsoft:
TODO: TryGetApplicationErrorsFromHeaders

timtay-microsoft:
> > TODO: AppErrPayload does not need to be modeled by DTDL, and it can just be a utf-8 string (usually json, but not forcibly so)
> 
> I recommend (open to discussion, of course) that we prescribe JSON specifically, in the interest of cross-application consistency and the ability for an SDK to express the value via a DOM. I don't think it needs to be a JSON object; it could be a JSON string, JSON number, JSON array, etc., but unless there is a scenario for which JSON encoding would be problematic, it could be helpful to remove a degree of variability here.

The only (mild) issue here is that we will then need to take a dependency on a JSON library in our RPC code to make that prescription. This is fine for .NET since we have System.Text.Json as a pretty standard official JSON library. 

@clecompt-msft is there a conventional JSON library for Go?

@vaavva same question for Rust

jrdouceur:
> @clecompt-msft is there a conventional JSON library for Go that we can take a dependency on in our RPC layer?
> 
> @vaavva same question for Rust

I can answer both of these.  JSON is supported natively in Go.  Rust would require taking a dependency on an external crate, but it's a very common one.

timtay-microsoft:
> > @clecompt-msft is there a conventional JSON library for Go that we can take a dependency on in our RPC layer?
> > @vaavva same question for Rust
> 
> I can answer both of these. JSON is supported natively in Go. Rust would require taking a dependency on an external crate, but it's a very common one.

Then I have no objections on this approach

vaavva:
> > > @clecompt-msft is there a conventional JSON library for Go that we can take a dependency on in our RPC layer?
> > > @vaavva same question for Rust
> > 
> > 
> > I can answer both of these. JSON is supported natively in Go. Rust would require taking a dependency on an external crate, but it's a very common one.
> 
> Then I have no objections on this approach

I would strongly prefer to not take a JSON serializer dependency in the SDK unless we really needed to, and I'm not sure this change justifies it. In a generic solution, there's no reason that this needs to be specifically JSON as opposed to anything serialized into UTF-8. JSON should just be a recommendation from us as a convenient option. Cross-application consistency for a free-form field should be enforced by over the wire contracts (whether DTDL or not), not the SDK. Even for our payload, the invoker/executor using the same payload format is enforced by either the DTDL or contracts between the writers of both sides. If stronger enforcement for this solution is required/desired, creating wrapper client implementations that force using this error model would be the way to go.

With the discussion earlier that we wouldn't model the error and just provide a UTF-8 string, I'm not sure why enforcing JSON doesn't fall into the same category. If we want to enforce this at the codegen level, where we already have the JSON dependency, then that would seem more reasonable, although my concern about unnecessarily restricting to JSON I think still applies.

jrdouceur:
> Cross-application consistency for a free-form field should be enforced by over the wire contracts (whether DTDL or not), not the SDK.

I think we should get @lt72's input on this.  I know he has fairly strong opinions on having the SDKs enforce consistency directly for some core aspects of the protocol, including the means by which user errors are represented.

> With the discussion earlier that we wouldn't model the error and just provide a UTF-8 string, I'm not sure why enforcing JSON doesn't fall into the same category.

It seems to me it would be reasonable to say that:

* The value of `AppErrCode` is a string, representing an error condition, that can be used as an enumerated value in common languages, so it:
  * must contain only letters and digits
  * must start with a letter
  * (possibly allows underscores, but there are subtleties here we might wish to avoid)
* The value of `AppErrDetail` is a string, representing error detail information, that is a valid JSON element.

Requiring JSON for the detail header ensures that no one does something crazy like Base64-encoding a binary representation that might present inter-operational challenges.

lt72:
> > Cross-application consistency for a free-form field should be enforced by over the wire contracts (whether DTDL or not), not the SDK.
> 
> I think we should get @lt72's input on this. I know he has fairly strong opinions on having the SDKs enforce consistency directly for some core aspects of the protocol, including the means by which user errors are represented.
> 
> > With the discussion earlier that we wouldn't model the error and just provide a UTF-8 string, I'm not sure why enforcing JSON doesn't fall into the same category.
> 
> It seems to me it would be reasonable to say that:
> 
> * The value of `AppErrCode` is a string, representing an error condition, that can be used as an enumerated value in common languages, so it:
>   
>   * must contain only letters and digits
>   * must start with a letter
>   * (possibly allows underscores, but there are subtleties here we might wish to avoid)
> * The value of `AppErrDetail` is a string, representing error detail information, that is a valid JSON element.
> 
> Requiring JSON for the detail header ensures that no one does something crazy like Base64-encoding a binary representation that might present inter-operational challenges.

Absolutely. Criss app consistency should also be enforced by SDK as much as possible, because we will have situation, like OPC UA, where we will have non-codegen server talking to codgen clients.

clecompt-msft:
I want to add some context for purposes of this discussion that came up in the sync this morning: One of the scenarios OPCUA described is having a _mix_ of success and error in a single message (e.g. where you have a list of results in the payload, some of which succeeded and some of which failed). This in itself throws into question the concept of binary success/error results, and there's no way to idiomatically represent that across languages and across error states.

Given scenarios like that (and others that don't have the SDK on both sides of the equation and so may not follow our error model at all, like DSS), we can't assume that we can "enforce" consistency - the best we can do is encourage it, which is I believe the goal of this ADR. And given _that_, we probably want to make sure we can handle (and ideally model as best as possible) endpoints that don't align with out recommendations.

doc/dev/adr/0020-error-modeling-headers.md
clecompt-msft: Might be worth mentioning that we'd ideally want DTDL to model headers in general, with the ability to tag them with error metadata similar to what we support for the payload currently.

timtay-microsoft:
> The value of AppErrCode is a string, representing an error condition, that can be used as an enumerated value in common languages, so it:
must contain only letters and digits
must start with a letter
(possibly allows underscores, but there are subtleties here we might wish to avoid)

Do we even need to impose limits like this? Our SDK gives these values as-is to the user. The only real limitation comes from MQTT user properties. Longer term, we plan on modeling the possible values in DTDL anyways, right?

doc/dev/adr/0021-error-modeling-headers.md
```diff
@@ -0,0 +1,99 @@
...
+
+In addition, users must be able to include arbitrary user properties in their command response to support this error reporting in case our standard fields are insufficient. This feature is likely in place already for all languages.
+
+On the command invoker side, we will add APIs for checking if a response was an application error and returning the error code and error data fields parsed from the MQTT message "AppErrCode" and "AppErrPayload" user properties.
+
+Other than these two new user properties, the over-the-wire behavior of our protocol won't change as a result of this decision.
+
+In order to provide a strongly-typed experience, we will also add codegen support for modeling the error codes in DTDL. This modeling will be detailed in a separate ADR, though. We will not provide modeling of the application error payload types in DTDL, though.
```
jrdouceur, line 19: > We will not provide modeling of the application error payload types in DTDL, though.

Actually, we already [do](https://github.com/Azure/opendigitaltwins-dtdl/blob/master/DTDL/v4/DTDL.mqtt.v3.md#error-and-errormessage-adjunct-types).

jrdouceur, line 19: > We will not provide modeling of the application error payload types in DTDL, though.

Actually, we already [do](https://github.com/Azure/opendigitaltwins-dtdl/blob/master/DTDL/v4/DTDL.mqtt.v3.md#error-and-errormessage-adjunct-types).

jrdouceur (approved):

doc/dev/adr/0021-error-modeling-headers.md
timtay-microsoft: Neat! I'll fix this

clecompt-msft (approved):

PR #678, opened 2025-04-04
[.NET + Rust + Go] Minor logging changes, cleanup

timtay-microsoft:
Logging changes
- Fix minor capitalization issues in rust logs
- Remove/rework some .NET logs that touch/may touch secrets

### DIFF
--- a/dotnet/src/Azure.Iot.Operations.Connector/TelemetryConnectorWorker.cs
+++ b/dotnet/src/Azure.Iot.Operations.Connector/TelemetryConnectorWorker.cs
@@ -77,7 +77,7 @@ namespace Azure.Iot.Operations.Connector
 
             // Create MQTT client from credentials provided by the operator
             MqttConnectionSettings mqttConnectionSettings = MqttConnectionSettings.FromFileMount();
-            _logger.LogInformation($"Connecting to MQTT broker with {mqttConnectionSettings}");
+            _logger.LogInformation("Connecting to MQTT broker with hostname {hostname} and port {port}", mqttConnectionSettings.HostName, mqttConnectionSettings.TcpPort);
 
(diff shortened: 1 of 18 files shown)


go/services/statestore/client.go
```diff
@@ -252,7 +252,7 @@ func (c *Client[K, V]) logKV(
 	if c.log.Enabled(ctx, slog.LevelDebug) {
 		all := []slog.Attr{
 			slog.String("key", string(key)),
-			slog.String("value", string(value)),
+			slog.String("value", string(value)), // Should we avoid logging values? Customers may put secrets in here
```
timtay-microsoft, line 255: @clecompt-msft would you be in favor of only logging keys for state store level operations?

 I know that the state store isn't _supposed_ to hold secrets, but I bet some customers will use it that way.

dotnet/src/Azure.Iot.Operations.Connector/TelemetryConnectorWorker.cs
```diff
@@ -77,7 +77,7 @@ protected override async Task ExecuteAsync(CancellationToken cancellationToken)
 
             // Create MQTT client from credentials provided by the operator
             MqttConnectionSettings mqttConnectionSettings = MqttConnectionSettings.FromFileMount();
-            _logger.LogInformation($"Connecting to MQTT broker with {mqttConnectionSettings}");
```
timtay-microsoft, line 80: This would log any username the user may have in their connection string

go/services/statestore/client.go
clecompt-msft: I'm okay with that. Lets the two log commands collapse into one, too.

maximsemenov80 (approved):

PR #681, opened 2025-04-07
[.Net] ADR Client

maximsemenov80:
Adding ADR Client


dotnet/src/Azure.Iot.Operations.Services/AssetAndDeviceRegistry/Models/ProtocolConverter.cs
```diff
@@ -0,0 +1,196 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Azure.Iot.Operations.Services.AssetAndDeviceRegistry.AdrBaseService;
+using Azure.Iot.Operations.Services.AssetAndDeviceRegistry.AepTypeService;
+
+namespace Azure.Iot.Operations.Services.AssetAndDeviceRegistry.Models;
+
+public static class ProtocolConverter
```
timtay-microsoft, line 9: This is a class for converting from public APIs to code gen, right? If so, this class shouldn't be public since no user should need to access it

dotnet/src/Azure.Iot.Operations.Services/AssetAndDeviceRegistry/Models/ModelsConverter.cs
```diff
@@ -0,0 +1,253 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Azure.Iot.Operations.Services.AssetAndDeviceRegistry.AdrBaseService;
+using Azure.Iot.Operations.Services.AssetAndDeviceRegistry.AepTypeService;
+
+namespace Azure.Iot.Operations.Services.AssetAndDeviceRegistry.Models;
+
+public static class ModelsConverter
```
timtay-microsoft, line 9: Same note here. Doesn't need to be public

dotnet/src/Azure.Iot.Operations.Services/AssetAndDeviceRegistry/Models/EventsSchemaSchemaElement.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Azure.Iot.Operations.Services.AssetAndDeviceRegistry.Models;
+
+public record EventsSchemaSchemaElement
```
timtay-microsoft, line 6: I'm guessing this is the name codegen gives us, but is there any way to rework this to avoid "SchemaSchema" in the name?

dotnet/src/Azure.Iot.Operations.Services/AssetAndDeviceRegistry/Models/CreateDiscoveredAssetEndpointProfileRequest.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Azure.Iot.Operations.Services.AssetAndDeviceRegistry.Models;
+
+public record CreateDiscoveredAssetEndpointProfileRequest
+{
+    public string? AdditionalConfiguration { get; set; } = default;
+
+    public string? DaepName { get; set; } = default;
```
timtay-microsoft, line 10: This name is also a bit clunky. Maybe the DTDL could just call this field "name" since the "discovered asset endpoint profile" portion of the name is covered by the class name?

dotnet/src/Azure.Iot.Operations.Services/AssetAndDeviceRegistry/IAdrServiceClient.cs
```diff
@@ -0,0 +1,128 @@
...
+    /// <param name="commandTimeout">Optional timeout for the command execution.</param>
+    /// <param name="cancellationToken">Token to cancel the operation.</param>
+    /// <returns>The response containing details of the created discovered AEP.</returns>
+    Task<CreateDiscoveredAssetEndpointProfileResponse> CreateDiscoveredAssetEndpointProfileAsync(string aepName,
+        CreateDiscoveredAssetEndpointProfileRequest request,
+        TimeSpan? commandTimeout = null,
+        CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// Event triggered when telemetry updates for an Asset Endpoint Profile (AEP) are received.
```
timtay-microsoft, line 120: I'd call out in these headers that users must turn on telemetry updates by calling the Observe APIs here. Otherwise a user may think they just need to set this callback

dotnet/src/Azure.Iot.Operations.Services/AssetAndDeviceRegistry/AdrServiceClient.cs
```diff
@@ -0,0 +1,254 @@
...
+
+namespace Azure.Iot.Operations.Services.AssetAndDeviceRegistry;
+
+public class AdrServiceClient(ApplicationContext applicationContext, IMqttPubSubClient mqttClient) : IAdrServiceClient
+{
+    private const string _aepNameTokenKey = "aepName";
+    private static readonly TimeSpan _defaultTimeout = TimeSpan.FromSeconds(10);
+    private readonly AssetEndpointProfileServiceClientStub _assetEndpointProfileServiceClient = new(applicationContext, mqttClient);
+    private readonly AssetServiceClientStub _assetServiceClient = new(applicationContext, mqttClient);
+    private readonly SemaphoreSlim _syncLock = new(1, 1);
```
timtay-microsoft, line 16: nit: It looks like you could have separate semaphores for observing assets vs observing asset endpoint profiles

dotnet/src/Azure.Iot.Operations.Services/AssetAndDeviceRegistry/IAdrServiceClient.cs
```diff
@@ -0,0 +1,128 @@
...
+/// </summary>
+public interface IAdrServiceClient : IAsyncDisposable
+{
+    /// <summary>
+    /// Starts observing updates for a specified Asset Endpoint Profile (AEP).
+    /// </summary>
+    /// <param name="aepName">The name of the Asset Endpoint Profile to observe.</param>
+    /// <param name="cancellationToken">Token to cancel the operation.</param>
+    /// <returns>A notification response indicating the result of the operation.</returns>
+    Task<NotificationResponse> ObserveAssetEndpointProfileUpdatesAsync(string aepName, CancellationToken cancellationToken);
```
timtay-microsoft, line 16:
```suggestion
    Task<NotificationResponse> ObserveAssetEndpointProfileUpdatesAsync(string aepName, CancellationToken cancellationToken = default);
```

So that users don't have to manually specify the cancellation token if they don't want it

timtay-microsoft: Also, we should have an optional param in these methods for the RPC/Telemetry layer timeout like the schema registry client has. That way, a user isn't forced to use that 10 second default you hardcoded

timtay-microsoft (approved):

dotnet/src/Azure.Iot.Operations.Services/AssetAndDeviceRegistry/Models/EventsSchemaSchemaElement.cs
maximsemenov80: and this is after I got rid of one _Schema_ already original name was `EventsSchemaSchemaElementSchema` :)

dotnet/src/Azure.Iot.Operations.Services/AssetAndDeviceRegistry/AdrServiceClient.cs
maximsemenov80: I was thinking about doing that, but we have one method to stop both receivers for AEP and Asset telemetry, so we don't want to stop it before both flags `_observingAssetEndpointProfileUpdates` and `_observingAssetUpdates` hance one lock... but looking at it now I think even this logic is wrong cause it doesn't account for AEPNames and AssetName - we should stop only after all Observe(aepNmae, assetName) matched by Unobserve(aepNmae, assetName), right?

dotnet/src/Azure.Iot.Operations.Services/AssetAndDeviceRegistry/AdrServiceClient.cs
```diff
@@ -0,0 +1,222 @@
...
+
+    /// <inheritdoc/>
+    public async Task<NotificationResponse> UnobserveAssetEndpointProfileUpdatesAsync(string aepName, TimeSpan? commandTimeout = null,
+        CancellationToken cancellationToken = default)
+    {
+        cancellationToken.ThrowIfCancellationRequested();
+        ObjectDisposedException.ThrowIf(_disposed, this);
+
+        if (_observedAeps.TryRemove(aepName, out _) && _observedAeps.IsEmpty)
+            await _assetServiceClient.AssetEndpointProfileUpdateEventTelemetryReceiver.StopAsync(cancellationToken);
```
timtay-microsoft, line 64:
```suggestion
        if (_observedAeps.TryRemove(aepName, out _) && _observedAeps.IsEmpty)
        {
            await _assetServiceClient.AssetEndpointProfileUpdateEventTelemetryReceiver.StopAsync(cancellationToken);
        }
```

Try to avoid skipping braces for even one line if statements

timtay-microsoft (approved):

PR #690, opened 2025-04-10
[Rust][Tools] Stub Service + Schema Registry Stub

RLeclair:
# Context

The schema registry service deployed by AIO has external dependencies, making it challenging to use for inner loop testing. To address this, the stub service introduced in this PR provides a simple, local implementation of the schema registry service to accelerate inner loop development.

# Changes

- Adds a new crate in the `tools` folder for the stub service.
- Introduces a simple CI job to build the stub service.
- Stub Service:
  - Includes multiple stubs for different services in the `Services` crate that have external dependencies and are difficult to simulate.
  - Each service has its own session to approximate real-world behavior.
  - Each service logs its state and activity to an output directory specified by an environment variable.
  - The output format is `[Environment variable]/stub_service_[Timestamp]/[Service Name]/[State / Log name].json`.
- Schema Registry Stub:
  - Implements a basic schema registry service supporting `Get` and `Put` operations.
  - Maintains an internal hashmap where the key is a schema's hash, and the value is a set of schema versions.
  - Outputs logs to the `SchemaRegistry` folder within the current instance of the output directory.
  - Supports only happy path interactions (valid `Get` and `Put` operations).

# Future Improvements

- Add proper error handling for schema registry service errors. Plans are underway to define these errors in the schema registry's DTDL, so error handling is not implemented in this version.
- Integrate the stub service with the SDK testing pipeline.

# Comments

A version of the schema registry without external dependencies is under evaluation. If implemented, this stub service will be deprecated and removed.


tools/stub_service/src/schema_registry/service.rs
```diff
@@ -0,0 +1,318 @@
...
+            }
+        };
+
+        Ok(())
+    }
+
+    async fn get_schema_runner(
+        mut get_command_executor: service_gen::GetCommandExecutor<C>,
+        schemas: Arc<Mutex<HashMap<String, BTreeSet<Schema>>>>,
+    ) -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
```
vaavva, line 105: does this need to be the Box error - it looks like only an AIOProtocolError could be returned for either fn

tools/stub_service/src/schema_registry/service.rs
```diff
@@ -0,0 +1,318 @@
...
+                            let mut schemas =
+                                schemas.lock().expect("Mutex management should be safe");
+
+                            schemas
+                                .entry(schema_name.clone())
+                                .and_modify(|schema_set| {
+                                    // Case in which the schema already exists
+
+                                    // Replace the schema with the new one or add it to the set if it doesn't exist
+                                    let old_schema = schema_set.replace(schema.clone());
```
vaavva, line 243: in this scenario, does the one that got replaced have to be exactly the same or just have the same version?

vaavva (approved):

tools/stub_service/src/lib.rs
```diff
@@ -0,0 +1,156 @@
...
+
+use azure_iot_operations_mqtt::{
+    MqttConnectionSettingsBuilder,
+    session::{Session, SessionOptionsBuilder},
+};
+
+/// Module for the schema registry stub service.
+pub mod schema_registry;
+
+const HOSTNAME: &str = "localhost";
```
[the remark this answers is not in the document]
RLeclair: Moved localhost and port, I think the output directory name and the environment variable should still be here as they should never change

tools/azure_iot_operations_stub_services/src/schema_registry.rs
```diff
@@ -180,7 +182,7 @@ impl From<service_gen::PutRequestSchema> for Schema {
                 .version
                 .expect("Schema version is required")
                 .parse()
-                .unwrap(), // TODO: Implement error handling for incorrect version number
-        }
+                .map_err(|_| "Invalid schema version, skipping request".to_string())?, // TODO: Implement error handling for incorrect version number
```
vaavva, line 185: should this be updated to just log, similar to the change you made in service.rs?

RLeclair: The difference here is that I am doing a conversion from a PutRequest into a Schema, the only way to let the service.rs layer know that we have encountered an invalid version and skip that request is to surface the message as an error

When we receive a get we don't do a conversion so I can detect the invalid version in that layer and just skip the request

PR #694, opened 2025-04-14
[Go] ADR Client

lev-i:


go/services/adr/gen.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+ROOT=$(git rev-parse --show-toplevel)
+
+"$(find "$ROOT/codegen/src" -name Azure.Iot.Operations.ProtocolCompiler -type f)" \
+	--modelFile "$ROOT/eng/dtdl/aep-name-based-operations.json" \
+	--outDir "$ROOT/go/services/adr/adrnamegen" \
```
clecompt-msft, line 6: It looks like you're already re-exporting the types users will care about, so I'd probably put the generated files in `internal`.

go/services/adr/client.go
```diff
@@ -0,0 +1,539 @@
...
+	DetectedAsset                  = adrbaseservice.DetectedAsset
+	DiscoveredAssetEndpointProfile = aeptypeservice.DiscoveredAssetEndpointProfile
+)
+
+// Error represents an error returned by the ADR service.
+type Error struct {
+	Message       string
+	PropertyName  string
+	PropertyValue string
+	Condition     interface{}
```
clecompt-msft, line 37: nit: Modern Go tends towards `any` instead of `interface{}`. I could have sworn `mage ci` autofixes this, though...

go/services/adr/client.go
```diff
@@ -0,0 +1,539 @@
...
+	getAssetEndpointProfileCommandInvoker              *adrbaseservice.GetAssetEndpointProfileCommandInvoker
+	updateAssetEndpointProfileStatusCommandInvoker     *adrbaseservice.UpdateAssetEndpointProfileStatusCommandInvoker
+	createDiscoveredAssetEndpointProfileCommandInvoker *aeptypeservice.CreateDiscoveredAssetEndpointProfileCommandInvoker
+	createDetectedAssetCommandInvoker                  *adrbaseservice.CreateDetectedAssetCommandInvoker
+	getAssetCommandInvoker                             *adrbaseservice.GetAssetCommandInvoker
+	updateAssetStatusCommandInvoker                    *adrbaseservice.UpdateAssetStatusCommandInvoker
+}
+
+// ClientOption allows setting optional parameters on the Client.
+type ClientOption func(*Client)
```
clecompt-msft, line 67: nit: This is a bit different than how the rest of the SDK does options. It's the traditional functional options pattern, so I won't block on it, but figured I'd mention in case you care about the discrepancy.

go/services/adr/client.go
```diff
@@ -0,0 +1,539 @@
...
+	}
+}
+
+// New creates a new ADR client.
+func New(
+	app *protocol.Application,
+	client protocol.MqttClient,
+	opts ...ClientOption,
+) (*Client, error) {
+	notifyInvoker, err := adrbaseservice.NewNotifyOnAssetEndpointProfileUpdateCommandInvoker(
```
clecompt-msft, line 100: Invokers are also listeners, so you probably want to include this in `c.listeners`. You also need to `Close()` everything if there's an error (which the `listeners` collection is also useful for). That being said, why aren't you using the generated wrapper(s) here, which should already be doing all that?

go/services/adr/client.go
```diff
@@ -0,0 +1,539 @@
...
+	}
+	c.listeners = append(c.listeners, aepUpdateReceiver)
+
+	return c, nil
+}
+
+// Close releases resources associated with the client.
+func (c *Client) Close() error {
+	var err error
+	c.closeOnce.Do(func() {
```
clecompt-msft, line 153: nit: `c.listeners.Close()` should already be idempotent, so the `Once` shouldn't be necessary.

go/services/adr/client.go
```diff
@@ -0,0 +1,539 @@
...
+	}
+
+	tokens := map[string]string{
+		aepNameTokenKey: aepName,
+	}
+
+	resp, err := c.updateAssetEndpointProfileStatusCommandInvoker.UpdateAssetEndpointProfileStatus(
+		ctx,
+		req,
+		protocol.WithTopicTokens(tokens),
```
clecompt-msft, line 358: nit: If you're not reusing the `tokens` variable anywhere else, you can actually do `protocol.WithTopicTokens{ aepNameTokenKey: aepName }`. That was one of the benefits of using a type over the traditional functional option.

go/services/adr/client.go
```diff
@@ -0,0 +1,539 @@
...
+		aepName,
+	)
+
+	err := c.onAepUpdate(aepName, &msg.Payload)
+	msg.Ack()
+	return err
+}
+
+// translateError converts protocol errors to client errors.
+func translateError(err error) error {
```
clecompt-msft, line 531: I assume there will be more logic here?

go/services/adr/client.go
```diff
@@ -0,0 +1,539 @@
...
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	c := &Client{
+		logger:         slog.Default(),
+		protocol:       app,
+		mqtt:           client,
+		observedAeps:   make(map[string]struct{}),
```
clecompt-msft, line 113: These are written to but never read. What are they doing?

go/services/adr/client.go
lev-i: I ran `mage ci` and it made formatting fixes, not sure why it wasn't amended. Changed to any

go/services/adr/client.go
lev-i: Properly implemented token

lev-i: Got them all this time

go/services/adr/client.go
```diff
@@ -0,0 +1,600 @@
...
+) error {
+	aepName := msg.TopicTokens[aepNameTokenKey]
+
+	// Skip if we are not observing the asset.
+	assetKey := aepName + "~" + msg.Payload.Name
+	c.mu.RLock()
+	_, observed := c.observedAssets[assetKey]
+	c.mu.RUnlock()
+	if !observed {
+		msg.Ack()
```
clecompt-msft, line 464: nit: since you're always `Ack`ing anyways, it might be cleaner to `defer msg.Ack()` near the top of the method. Ditto for `handleAepUpdateTelemetry`.

clecompt-msft (approved):

PR #695, opened 2025-04-14
[.NET][Akri] Update CreateFromFileMount flow

timtay-microsoft:
- Several env vars have changed
- Moved this flow to connector package since only connectors should be doing this


dotnet/test/Azure.Iot.Operations.Connector.UnitTests/TestMountFiles/ConnectorConfig/AIO_METADATA
```diff
@@ -0,0 +1,4 @@
+﻿{
```
vaavva, line 1: can you put these files in the /eng folder - I bet these will be helpful for testing in the other languages too

timtay-microsoft: Good idea. Done.

maximsemenov80 (approved):

PR #700, opened 2025-04-15
[.NET] ADR client updated

maximsemenov80:
This pull request includes significant changes to the `AdrBaseService` class in the Azure IoT Operations Services. The changes primarily focus on renaming and restructuring command executors and telemetry senders to better align with device-centric terminology.

Key changes include:

### Command Executors and Telemetry Senders Renaming:
* Renamed `GetAssetEndpointProfileCommandExecutor` to `GetDeviceCommandExecutor`.
* Renamed `UpdateAssetEndpointProfileStatusCommandExecutor` to `UpdateDeviceStatusCommandExecutor`.
* Renamed `NotifyOnAssetEndpointProfileUpdateCommandExecutor` to `SetNotificationPreferenceForDeviceUpdatesCommandExecutor`.
* Renamed `AssetEndpointProfileUpdateEventTelemetrySender` to `DeviceUpdateEventTelemetrySender`.

### Constructor Adjustments:
* Updated the constructor to initialize the new command executors and telemetry senders.

### Method Renaming:
* Updated method names to reflect the new device-centric terminology, such as `GetDeviceAsync` and `UpdateDeviceStatusAsync`.

### Telemetry Method Update:
* Updated the `SendTelemetryAsync` method to use `DeviceUpdateEventTelemetry` instead of `AssetEndpointProfileUpdateEventTelemetry`.


dotnet/src/Azure.Iot.Operations.Services/AssetAndDeviceRegistry/AdrServiceClient.cs
```diff
@@ -31,197 +34,292 @@ public async ValueTask DisposeAsync()
...
+    /// <inheritdoc />
+    public async Task<NotificationResponse> ObserveDeviceEndpointUpdatesAsync(string deviceName, string endpointName, TimeSpan? commandTimeout = null,
         CancellationToken cancellationToken = default)
     {
         cancellationToken.ThrowIfCancellationRequested();
         ObjectDisposedException.ThrowIf(_disposed, this);
 
-        _observedAeps[aepName] = _dummyByte;
-        await _assetServiceClient.AssetEndpointProfileUpdateEventTelemetryReceiver.StartAsync(cancellationToken);
+        _observedEndpoints[$"{deviceName}_{endpointName}"] = _dummyByte;
```
timtay-microsoft, line 49: The service-side docs refer to this field as "InboundEndpointName" rather than just "EndpointName" so we should probably follow suit

dotnet/src/Azure.Iot.Operations.Services/AssetAndDeviceRegistry/IAdrServiceClient.cs
```diff
@@ -8,129 +8,168 @@ namespace Azure.Iot.Operations.Services.AssetAndDeviceRegistry;
...
-    /// <param name="cancellationToken">Token to cancel the operation.</param>
-    /// <returns>The response containing details of the created discovered AEP.</returns>
-    Task<CreateDiscoveredAssetEndpointProfileResponse> CreateDiscoveredAssetEndpointProfileAsync(string aepName,
+    /// <param name="deviceName">The name of the device.</param>
+    /// <param name="endpointName">The name of the endpoint.</param>
+    /// <param name="request">The request containing discovered asset endpoint profile creation parameters.</param>
+    /// <param name="commandTimeout">Optional timeout for the command.</param>
+    /// <param name="cancellationToken">Optional cancellation token.</param>
+    /// <returns>A task that represents the asynchronous operation, containing the response for the created discovered asset endpoint profile.</returns>
+    Task<CreateDiscoveredAssetEndpointProfileResponse> CreateDiscoveredAssetEndpointProfileAsync(
```
timtay-microsoft, line 157: "AssetEndpointProfile" still? Is this coming from the DTDL, or do we just need to rename the wrapper model class?

timtay-microsoft: Looks like the DTDL still calls it this in a few spots. Can you follow up with Abhipsa to see if that was intentional? I'm guessing it isn't. I wouldn't block this PR waiting on a response, though

timtay-microsoft: https://github.com/Azure/iot-operations-sdks/blob/a533017069425a22754fb21a742abb9c9919681d/eng/dtdl/aep-name-based-operations.json#L1433

This is another field that I think she will probably want to rename now that "Asset Endpoint Profile" isn't the term we use

maximsemenov80: this part in DTDL didn't change

maximsemenov80: actully good catch, I missed that this method is coming from different client and need different parameters:
image

PR #710, opened 2025-04-18
[Dev Branch] [Rust][ADR] Device structs

vaavva:
Optimized structs for Device operations.

Into/From traits not included in this PR
Discovery related structs not included in this PR

### DIFF
--- a/rust/azure_iot_operations_services/src/azure_device_registry.rs
+++ b/rust/azure_iot_operations_services/src/azure_device_registry.rs
@@ -5,6 +5,11 @@
 
 //! Types for Azure Device Registry operations.
 
+use core::fmt::Debug;
+use std::collections::HashMap;
+
+use crate::azure_device_registry::device_name_gen::adr_base_service::client as adr_name_gen;
+
 /// Azure Device Registry generated code
 mod device_name_gen;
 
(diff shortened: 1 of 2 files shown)


rust/azure_iot_operations_services/src/azure_device_registry.rs
```diff
@@ -5,20 +5,179 @@
...
+                info: details.info,
+                message: details.message,
+            }),
+            inner_error: value.inner_error,
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+/// Details describing an error?
```
RLeclair, line 72:
```suggestion
/// Details describing an error
```

Not sure if this is a typo

rust/azure_iot_operations_services/src/azure_device_registry.rs
```diff
@@ -5,20 +5,179 @@
...
+/// Represents the status of a Device in the ADR Service.
+pub struct DeviceStatus {
+    /// The 'config' Field.
+    pub config: Option<DeviceStatusConfig>,
+    /// The 'endpoints' Field.
+    pub endpoints: HashMap<String, Option<ConfigError>>,
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct DeviceStatusConfig {
```
RLeclair, line 175: I think this can eventually be shared with the asset structs (from asset.yaml), they have the same fields, so maybe it can go in common

vaavva: ooh yep, I'll change that now so Oliva can use the same one - good catch!

rust/azure_iot_operations_services/src/azure_device_registry.rs
vaavva: we're not quite sure what this struct is for yet. I'll remove the doc comment as a whole for now

PR #728, opened 2025-04-22
[.Net] ADR integration tests

maximsemenov80:


dotnet/test/Azure.Iot.Operations.Services.IntegrationTests/resources/coap-asset.yaml
```diff
@@ -0,0 +1,77 @@
+apiVersion: v1
```
vaavva, line 1: can you move these to the /eng folder so we can use them for other language tests too?

PR #736, opened 2025-04-24
[Rust][Connector] ADR File Mount Implementation

RLeclair:
This PR:
- Simplified the API for observing an ADR file mount
  - Now a single observation for the file mount is created from which device endpoint observations are derived. Each device observation can in turn derive asset observations.
- Removed unnecessary methods based on the above.
- Added error handling for the API with a few exceptions (marked TODO) and errors surfacing at the watcher level.
- Adds happy path testing for the API


rust/azure_iot_operations_connector/src/file_mount_azure_device_registry.rs
```diff
@@ -1,4 +1,1110 @@
...
+        // Tracks devices and assets in the file mount.
+        let mut file_mount_map = FileMountMap::new(create_device_tx.clone());
+
+        // Add the already existing device endpoints and their assets to the file mount map
+        get_device_endpoint_names(&mount_path)?
+            .iter()
+            .try_for_each(|device| {
+                file_mount_map.insert_device_endpoint(device);
+
+                let assets = get_asset_names(&mount_path, device)?;
```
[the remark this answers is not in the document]
vaavva: It's an interesting question if this fn should be able to return both some success, but also report some errors - If there are any errors, the operator might need to take some action, but some of the devices might be able to run in the meantime? Maybe this depends on whether a failure to parse one of the files (because of the contents, not an access error) can only be caused by an error in the implementation of the operator service, or whether bad input from a customer could cause that as well

rust/azure_iot_operations_connector/src/file_mount_azure_device_registry.rs
```diff
@@ -1,4 +1,1110 @@
...
+    ///
+    /// # Arguments
+    /// * `debounce_duration` - The duration to debounce incoming I/O events.
+    ///
+    /// # Errors
+    /// - [`Error`] of kind [`ErrorKind::EnvironmentVariableError`] if the environment variable was not able to be read.
+    /// - [`Error`] of kind [`ErrorKind::WatcherError`] if the watcher could not be created.
+    /// - [`Error`] of kind [`ErrorKind::IoError`] if there are issues accessing the file mount.
+    /// - [`Error`] of kind [`ErrorKind::ParseError`] if there are issues parsing the file names and content.
+    pub fn new(debounce_duration: Duration) -> Result<DeviceEndpointCreateObservation, Error> {
```
vaavva, line 347: maybe weird question, but would we want to use an options builder struct so this can have a default? If not, a recommended default might be good to have in the docs

rust/azure_iot_operations_connector/src/file_mount_azure_device_registry.rs
```diff
@@ -1,4 +1,1110 @@
...
+                match res {
+                    Ok(events) => {
+                        // TODO: There might be a case where we receive events out of order (i.e a modify before a create). The file mount map accounts for this but
+                        // it might be better to just match on an Any.
+                        for debounced_event in &events {
+                            match debounced_event.event.kind {
+                                EventKind::Create(event::CreateKind::File) => { // Event signals the creation of a device endpoint
+                                    debounced_event.paths.iter().for_each(|path| {
+                                        // TODO: We could use an expect here since we are sure the path is valid
+                                        let Some(mount_path) = path.parent() else {
```
vaavva, line 384: is this just a safeguard check? Because we know what this value should be already, right?

rust/azure_iot_operations_connector/src/file_mount_azure_device_registry.rs
```diff
@@ -1,4 +1,1110 @@
...
+                                            log::warn!("Failed to get parent path from device endpoint");
+                                            return;
+                                        };
+
+                                        let device = match DeviceEndpointRef::try_from(path) {
+                                            Ok(device) => device,
+                                            Err(err) => {
+                                                log::warn!("Failed to parse device endpoint from path: {err:?}");
+                                                return;
```
vaavva, line 393: log and continue for the next event instead of return? Since this would be from a malformed file name I think

rust/azure_iot_operations_connector/src/file_mount_azure_device_registry.rs
```diff
@@ -1,4 +1,1110 @@
...
+                                                    log::warn!("Failed to get asset names: {err:?}");
+                                                    return;
+                                                }
+                                            };
+
+                                        // Update the file mount map with the new assets
+                                        file_mount_map.update_assets(&device, &assets);
+                                    });
+                                }
+                                EventKind::Modify(_) => { // Event signals the creation or deletion of an asset
```
vaavva, line 414: do we care about the contents of this event?

rust/azure_iot_operations_connector/src/file_mount_azure_device_registry.rs
```diff
@@ -1,4 +1,1110 @@
...
+    ///
+    /// Returns Some(([`DeviceEndpointRef`], [`AssetCreateObservation`])) if a notification is received or `None`
+    /// if there will be no more notifications (i.e. the channel is closed). This should not happen unless the
+    /// [`DeviceEndpointCreateObservation`] is dropped.
+    ///
+    /// The [`AssetCreateObservation`] should be used to receive notifications for asset creation events
+    /// associated with the device endpoint.
+    pub async fn recv_notification(
+        &mut self,
+    ) -> Option<(DeviceEndpointRef, AssetCreateObservation)> {
```
vaavva, line 495: should a device deletion token be returned here too, or is that information supposed to be inferred from the assetCreateObservation sending None? (if so, some more explicit doc around this would be good or maybe there's a better name for this Observation)

rust/azure_iot_operations_connector/src/file_mount_azure_device_registry.rs
```diff
@@ -1,4 +1,1110 @@
...
+            device_with_assets!("device2", "endpoint3");
+
+        file_mount_manager.add_device_endpoint(&device1_endpoint1, &device1_endpoint1_assets);
+        file_mount_manager.add_device_endpoint(&device1_endpoint2, &device1_endpoint2_assets);
+        file_mount_manager.add_device_endpoint(&device2_endpoint3, &device2_endpoint3_assets);
+
+        let mut device_endpoints =
+            HashSet::from([device1_endpoint1, device1_endpoint2, device2_endpoint3]);
+
+        temp_env::async_with_vars(
```
vaavva, line 719: does this need an assert around it? Checking to make sure that an error in the async portion will get properly propagated to fail the test

rust/azure_iot_operations_connector/src/file_mount_azure_device_registry.rs
```diff
@@ -1,4 +1,1110 @@
...
+        FileMountMap {
+            file_mount_map: HashMap::new(),
+            create_device_tx,
+        }
+    }
+
+    /// Inserts a new [`DeviceEndpointRef`] into the file mount map.
+    ///
+    /// This function creates a new channel for asset creation notifications and adds the device endpoint
+    /// to the file mount map. If the device endpoint already exists, it does nothing. It will also
```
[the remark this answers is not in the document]
RLeclair: It is to protect against debouncing but more specifically it is to prevent us from creating the unbounded channel below for the same device endpoint

rust/azure_iot_operations_connector/src/file_mount_azure_device_registry.rs
```diff
@@ -1,4 +1,1110 @@
...
+            .send((
+                device.clone(),
+                AssetCreateObservation::new(asset_creation_rx),
+            ))
+            .is_err()
+        {
+            // TODO: The cases in which this fails are:
+            // 1. The receiver is closed which means the `FileMountMap` is dropped, this should not happen.
+            // 2. The receiver is full which means we are not receiving notifications fast enough or
+            //    are out of space, this should be handled by the caller via a retry.
```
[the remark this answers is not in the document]
RLeclair: I was not sure whether I should do that here or not, but I'll add that as part of the TODO

rust/azure_iot_operations_connector/src/file_mount_azure_device_registry.rs
```diff
@@ -1,4 +1,1110 @@
...
+        // Get the asset creation channel and the tracked assets for this device
+        let Some((create_asset_tx, tracked_assets)) = self.file_mount_map.get_mut(device) else {
+            // If the device is non-existent we can't update the assets. Most likely a create
+            // notification has not been parsed yet but this function will be called again once
+            // the device is created.
+            return;
+        };
+
+        // Clean up the assets that are no longer in the file mount
+        let asset_set: HashSet<_> = assets.iter().collect();
```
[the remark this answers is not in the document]
RLeclair: I changed the get_assets function to return a HashSet to make it possible

rust/azure_iot_operations_connector/src/file_mount_azure_device_registry.rs
```diff
@@ -1,4 +1,1110 @@
...
+        // When an asset is not retained it gets dropped and its deletion token is triggered when
+        // the channel associated with it is dropped.
+        tracked_assets.retain(|tracked_asset, _| asset_set.contains(tracked_asset));
```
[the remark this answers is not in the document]
RLeclair: That's a great idea, just changing to `remove` does it :)

rust/azure_iot_operations_connector/src/file_mount_azure_device_registry.rs
RLeclair: Good point, I added a default in the example so customers can choose to copy that one, I'll add a comment here too

rust/azure_iot_operations_connector/src/file_mount_azure_device_registry.rs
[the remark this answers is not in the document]
RLeclair: In this case we are not parsing out the rest of the devices and just failing the operation, it would be good to check with the ADR folks on this

rust/azure_iot_operations_connector/src/file_mount_azure_device_registry.rs
RLeclair: This is the same as a continue since we're in the for_each

rust/azure_iot_operations_connector/src/file_mount_azure_device_registry.rs
RLeclair: That's a good point, this should not be different from the one we started watching

rust/azure_iot_operations_connector/src/file_mount_azure_device_registry.rs
RLeclair: We only care about the paths it gives us

rust/azure_iot_operations_connector/src/file_mount_azure_device_registry.rs
RLeclair: It is in the case that the asset create observation sends none, I have those details in the documentation for that structure but will add here too

rust/azure_iot_operations_connector/src/file_mount_azure_device_registry.rs
RLeclair: This just returns `()`, any failure inside will fail the test

PR #739, opened 2025-04-24
[Rust][Connector] File Mount ADR Client + Sample

RLeclair:
From #736 :
- Simplified the API for observing an ADR file mount
  - Now a single observation for the file mount is created from which device endpoint observations are derived. Each device observation can in turn derive asset observations.
- Removed unnecessary methods based on the above.
- Added error handling for the API with a few exceptions (marked TODO) and errors surfacing at the watcher level.
- Adds happy path testing for the API

_Addressed comments in previous PR_

This PR:
- Adds filemount module which will eventually hold watcher functionality.
- Adds an example showing sample usage.


rust/azure_iot_operations_connector/src/filemount/azure_device_registry.rs
```diff
@@ -0,0 +1,1114 @@
...
+
+    /// An error occurred while parsing the files in the file mount.
+    #[error("{0}")]
+    ParseError(String),
+}
+
+// ~~~~~~~~~~~~~~~~~ Observations ~~~~~~~~~~~~~~~~~~~~~
+
+/// Represents an observation for device endpoint creation events.
+#[allow(dead_code)]
```
vaavva, line 53: might not need this anymore since you have a sample (check for other instances of this as well)

rust/azure_iot_operations_connector/src/filemount/azure_device_registry.rs
```diff
@@ -0,0 +1,1114 @@
...
+    // Get the content of the file
+    let file_content = String::from_utf8(std::fs::read(file_path).map_err(ErrorKind::from)?)
+        .map_err(|e| ErrorKind::ParseError(e.to_string()))?;
+
+    // If the file is empty, return an empty vector
+    if file_content.is_empty() {
+        return Ok(HashSet::new());
+    }
+
+    // Split the file content by ';' and create a vector of AssetRef
```
vaavva, line 342:
```suggestion
    // Split the file content by ';' and create a HashSet of AssetRef
```

rust/azure_iot_operations_connector/examples/filemount_azure_device_registry_observation.rs
```diff
@@ -0,0 +1,244 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! This example demonstrates how to use the Azure Device Registry (ADR) file mount
+//! observation feature of the Azure IoT Operations Connector.
+//!
+//! To use the example, set the `ADR_RESOURCES_NAME_MOUNT_PATH` environment variable to the
+//! directory where the file mount will be created. The example will create
```
vaavva, line 8: Ideally we wouldn't have to mock this for the sample, but this is very convenient for now

vaavva (approved):
Check that cargo doc generates well if you haven't yet, LGTM!

rust/azure_iot_operations_connector/src/filemount/azure_device_registry.rs
RLeclair: It was just here!

PR #749, opened 2025-04-28
[.NET][Akri] Add replica field in connector samples, check for 401 errors when making HTTP requests

timtay-microsoft:
And bump versions for a preview release of these Akri bits


dotnet/samples/Connectors/EventDrivenTcpThermostatConnector/KubernetesResources/connector-template.yaml
```diff
@@ -14,7 +14,8 @@ spec:
       managedConfigurationType: imageConfiguration
       imageConfigurationSettings:
         image: eventdriventcpthermostatconnector:latest
-        imagePullPolicy: Never
+        imagePullPolicy: Never # This sample image is built locally, so no need to pull
+        replicas: 1
```
timtay-microsoft, line 18: It took me a bit of digging to find where this was configurable, so I figure including it in our samples will spare our users from the same

maximsemenov80 (approved):

PR #751, opened 2025-04-28
[.NET] update unit tests for better coverage

maximsemenov80:
primarily focus on removing unused models, updating unit tests, and enhancing test coverage for asset and device registry functionalities. The changes include the removal of obsolete files, updates to unit tests to verify new properties and configurations, and adjustments to project files to reflect the removals.


dotnet/src/Azure.Iot.Operations.Services/Azure.Iot.Operations.Services.csproj
```diff
@@ -36,4 +36,9 @@
     </Content>
   </ItemGroup>
 
+  <ItemGroup>
```
timtay-microsoft, line 39: Is this necessary? If the file still exists, but we don't want to compile it, we should remove the file

timtay-microsoft (approved):

dotnet/src/Azure.Iot.Operations.Services/Azure.Iot.Operations.Services.csproj
maximsemenov80: weird, I've simply deleted these files... need to keep an eye on IDE (facepalm)

PR #759, opened 2025-04-29
[Rust][Services] Split leased_lock client (ADR 20) and add lease & lock auto-renewal

ewertons:
According to ADR 20 this leased lock client should be split into two layers, lease and lock clients. The lease client will have the bulk of logic for acquiring, renewing and releasing a key. The lock layer will present the expected semantics for locking operations.
Additionally auto-renewal of leases was not yet implemented in the leased lock client. included in this PR is the logic to perform the auto-renewal of leases/locks.

### DIFF
--- a/rust/azure_iot_operations_services/examples/leased_lock_client.rs
+++ b/rust/azure_iot_operations_services/examples/leased_lock_client.rs
@@ -4,7 +4,6 @@
 use std::{sync::Arc, time::Duration};
 
 use env_logger::Builder;
-use tokio::sync::Notify;
 
 use azure_iot_operations_mqtt::MqttConnectionSettingsBuilder;
 use azure_iot_operations_mqtt::session::{
@@ -20,7 +19,7 @@ use azure_iot_operations_services::state_store;
 async fn main() {
     let client_id1 = "someClientId1";
     let client_id2 = "someClientId2";
-    let lock_name = "someLock";
+    let key_name = "someLock";
 
     Builder::new()
         .filter_level(log::LevelFilter::Info)
@@ -29,32 +28,19 @@ async fn main() {
         .init();
 
     let (session1, exit_handle1, state_store_client_arc1, leased_lock_client1) =
-        create_clients(client_id1, lock_name);
+        create_clients(client_id1, key_name);
 
     let (session2, exit_handle2, state_store_client_arc2, leased_lock_client2) =
-        create_clients(client_id2, lock_name);
-
-    let client_1_notify = Arc::new(Notify::new());
-    let client_2_notify = client_1_notify.clone();
+        create_clients(client_id2, key_name);
 
     let client_1_task = tokio::task::spawn(async move {
-        leased_lock_client_1_operations(
-            state_store_client_arc1,
-            leased_lock_client1,
-            exit_handle1,
-            client_1_notify,
-        )
-        .await;
+        leased_lock_client_1_operations(state_store_client_arc1, leased_lock_client1, exit_handle1)
+            .await;
     });
 
     let client_2_task = tokio::task::spawn(async move {
-        leased_lock_client_2_operations(
-            state_store_client_arc2,
-            leased_lock_client2,
-            exit_handle2,
-            client_2_notify,
-        )
-        .await;
+        leased_lock_client_2_operations(state_store_client_arc2, leased_lock_client2, exit_handle2)
+            .await;
     });
 
     let _ = tokio::try_join!(
@@ -67,7 +53,7 @@ async fn main() {
 
 fn create_clients(
     client_id: &str,
-    lock_name: &str,
+    key_name: &str,
 ) -> (
     Session,
     SessionExitHandle,
@@ -107,7 +93,7 @@ fn create_clients(
 
     let leased_lock_client = leased_lock::Client::new(
         state_store_client_arc.clone(),
-        lock_name.as_bytes().to_vec(),
+        key_name.as_bytes().to_vec(),
         client_id.as_bytes().to_vec(),
     )
     .unwrap();
@@ -125,7 +111,7 @@ fn create_clients(
 /// make all these calls, but they do show all that can be done with this client.
 ///
 /// In `leased_lock_client_1_operations` you will find the following examples:
-/// 1. Acquire a lock using `acquire_lock()`
+/// 1. Acquire a lock using `acquire()`
 /// 2. Sets a key in the State Store using the `fencing_token` obtained from the lock.
 /// 3. Gets the current lock holder name.
 /// 4. Releases a lock.
@@ -133,7 +119,6 @@ async fn leased_lock_client_1_operations(
     state_store_client_arc: Arc<state_store::Client<SessionManagedClient>>,
     leased_lock_client: leased_lock::Client<SessionManagedClient>,
     exit_handle: SessionExitHandle,
-    notify: Arc<Notify>,
 ) {
     let lock_expiry = Duration::from_secs(10);
     let request_timeout = Duration::from_secs(120);
@@ -145,17 +130,11 @@ async fn leased_lock_client_1_operations(
         expires: Some(Duration::from_secs(15)),
     };
 
-    notify.notified().await; // Wait for Client 2 to observe lock.
-    log::info!("Client 1 notified that Client 2 is observing lock.");
-
     // 1.
-    let fencing_token = match leased_lock_client
-        .acquire_lock(lock_expiry, request_timeout)
-        .await
-    {
-        Ok(acquire_lock_result) => {
+    let fencing_token = match leased_lock_client.lock(lock_expiry, request_timeout).await {
+        Ok(acquire_result) => {
             log::info!("Lock acquired successfully");
-            acquire_lock_result // a.k.a., the fencing token.
+            acquire_result // a.k.a., the fencing token.
         }
         Err(e) => {
             log::error!("Failed acquiring lock {e}");
@@ -190,10 +169,9 @@ async fn leased_lock_client_1_operations(
     };
 
     // 3.
-    get_lock_holder(&leased_lock_client, request_timeout).await;
 
     // 4.
-    match leased_lock_client.release_lock(request_timeout).await {
+    match leased_lock_client.unlock(request_timeout).await {
         Ok(()) => {
             log::info!("Lock released successfully");
         }
@@ -211,13 +189,12 @@ async fn leased_lock_client_1_operations(
 /// In `leased_lock_client_2_operations` you will find the following examples:
 /// 5. Start observing a lock, waiting for it to be released.
 /// 6. Stop observing a lock.
-/// 7. Make a call to `acquire_lock_and_update_value()` to set a key.
-/// 8. Make a call to `acquire_lock_and_update_value()` to delete a key.
+/// 7. Make a call to `lock_and_update_value()` to set a key.
+/// 8. Make a call to `lock_and_update_value()` to delete a key.
 async fn leased_lock_client_2_operations(
     state_store_client_arc: Arc<state_store::Client<SessionManagedClient>>,
     leased_lock_client: leased_lock::Client<SessionManagedClient>,
     exit_handle: SessionExitHandle,
-    notify: Arc<Notify>,
 ) {
     let lock_expiry = Duration::from_secs(10);
     let request_timeout = Duration::from_secs(10);
@@ -230,46 +207,9 @@ async fn leased_lock_client_2_operations(
         expires: Some(Duration::from_secs(15)),
     };
 
-    // 5.
-    let mut received_lock_free_notification = false;
-    while !received_lock_free_notification {
-        let mut observe_response = leased_lock_client
-            .observe_lock(request_timeout)
-            .await
-            .unwrap();
-
-        notify.notify_one();
-
-        loop {
-            let Some((notification, _)) = observe_response.response.recv_notification().await
-            else {
-                log::warn!(
-                    "Received None for lock notification. Client probably disconnected. observe_lock() must be called again."
-                );
-                break;
-            };
-
-            log::info!(
-                "Client2 received lock notification: {:?}",
-                notification.operation
-            );
-
-            if notification.operation == state_store::Operation::Del {
-                received_lock_free_notification = true;
-                break;
-            }
-        }
-    }
-
-    // 6.
-    leased_lock_client
-        .unobserve_lock(request_timeout)
-        .await
-        .unwrap();
-
     // 7.
     match leased_lock_client
-        .acquire_lock_and_update_value(
+        .lock_and_update_value(
             lock_expiry,
             request_timeout,
             shared_resource_key_name.to_vec(),
@@ -288,11 +228,11 @@ async fn leased_lock_client_2_operations(
         )
         .await
     {
-        Ok(acquire_lock_and_update_value_result) => {
-            if acquire_lock_and_update_value_result.response {
+        Ok(lock_and_update_value_result) => {
+            if lock_and_update_value_result.response {
                 log::info!("Key successfully set");
             } else {
-                log::error!("Could not set key {acquire_lock_and_update_value_result:?}");
+                log::error!("Could not set key {lock_and_update_value_result:?}");
                 return;
             }
         }
@@ -306,7 +246,7 @@ async fn leased_lock_client_2_operations(
 
     // 8.
     match leased_lock_client
-        .acquire_lock_and_update_value(
+        .lock_and_update_value(
             lock_expiry,
             request_timeout,
             shared_resource_key_name.to_vec(),
@@ -317,11 +257,11 @@ async fn leased_lock_client_2_operations(
         )
         .await
     {
-        Ok(acquire_lock_and_update_value_result) => {
-            if acquire_lock_and_update_value_result.response {
+        Ok(lock_and_update_value_result) => {
+            if lock_and_update_value_result.response {
                 log::info!("Key successfully deleted");
             } else {
-                log::error!("Could not delete key {acquire_lock_and_update_value_result:?}",);
+                log::error!("Could not delete key {lock_and_update_value_result:?}",);
                 return;
             }
         }
@@ -334,26 +274,4 @@ async fn leased_lock_client_2_operations(
     state_store_client_arc.shutdown().await.unwrap();
 
     exit_handle.try_exit().await.unwrap();
-}
-
-async fn get_lock_holder(
-    leased_lock_client: &azure_iot_operations_services::leased_lock::Client<SessionManagedClient>,
-    request_timeout: Duration,
-) {
-    match leased_lock_client.get_lock_holder(request_timeout).await {
-        Ok(lock_holder_response) => match lock_holder_response.response {
-            Some(holder_name) => {
-                log::info!(
-                    "Lock being held by {}",
-                    String::from_utf8(holder_name).unwrap()
-                );
-            }
-            None => {
-                log::info!("Lock is currently free");
-            }
-        },
-        Err(e) => {
-            log::error!("Failed getting lock holder: {e}");
-        }
-    };
 }
--- a/rust/azure_iot_operations_services/examples/leased_lock_client.rs
+++ b/rust/azure_iot_operations_services/examples/leased_lock_client.rs
@@ -10,9 +10,7 @@ use azure_iot_operations_mqtt::session::{
     Session, SessionExitHandle, SessionManagedClient, SessionOptionsBuilder,
 };
 use azure_iot_operations_protocol::application::ApplicationContextBuilder;
-use azure_iot_operations_services::leased_lock::{
-    self, AcquireAndUpdateKeyOption, SetCondition, SetOptions,
-};
+use azure_iot_operations_services::leased_lock::{SetCondition, SetOptions, lock};
 use azure_iot_operations_services::state_store;
 
 #[tokio::main(flavor = "current_thread")]
@@ -27,20 +25,18 @@ async fn main() {
         .filter_module("rumqttc", log::LevelFilter::Warn)
         .init();
 
-    let (session1, exit_handle1, state_store_client_arc1, leased_lock_client1) =
+    let (session1, exit_handle1, state_store_client_arc1, lock_client1) =
         create_clients(client_id1, key_name);
 
-    let (session2, exit_handle2, state_store_client_arc2, leased_lock_client2) =
+    let (session2, exit_handle2, state_store_client_arc2, lock_client2) =
         create_clients(client_id2, key_name);
 
     let client_1_task = tokio::task::spawn(async move {
-        leased_lock_client_1_operations(state_store_client_arc1, leased_lock_client1, exit_handle1)
-            .await;
+        lock_client_1_operations(state_store_client_arc1, lock_client1, exit_handle1).await;
     });
 
     let client_2_task = tokio::task::spawn(async move {
-        leased_lock_client_2_operations(state_store_client_arc2, leased_lock_client2, exit_handle2)
-            .await;
+        lock_client_2_operations(state_store_client_arc2, lock_client2, exit_handle2).await;
     });
 
     let _ = tokio::try_join!(
@@ -58,7 +54,7 @@ fn create_clients(
     Session,
     SessionExitHandle,
     Arc<state_store::Client<SessionManagedClient>>,
-    leased_lock::Client<SessionManagedClient>,
+    lock::Client<SessionManagedClient>,
 ) {
     let application_context = ApplicationContextBuilder::default().build().unwrap();
 
@@ -91,33 +87,27 @@ fn create_clients(
 
     let state_store_client_arc = Arc::new(state_store_client);
 
-    let leased_lock_client = leased_lock::Client::new(
+    let lock_client = lock::Client::new(
         state_store_client_arc.clone(),
         key_name.as_bytes().to_vec(),
         client_id.as_bytes().to_vec(),
     )
     .unwrap();
 
-    (
-        session,
-        exit_handle,
-        state_store_client_arc,
-        leased_lock_client,
-    )
+    (session, exit_handle, state_store_client_arc, lock_client)
 }
 
 /// In the functions below we show different calls that an application could make
 /// into the `leased_lock::Client`. Not necessarily an application would need to
 /// make all these calls, but they do show all that can be done with this client.
 ///
-/// In `leased_lock_client_1_operations` you will find the following examples:
-/// 1. Acquire a lock using `acquire()`
+/// In `lock_client_1_operations` you will find the following examples:
+/// 1. Acquire a lock using `lock()`
 /// 2. Sets a key in the State Store using the `fencing_token` obtained from the lock.
-/// 3. Gets the current lock holder name.
-/// 4. Releases a lock.
-async fn leased_lock_client_1_operations(
+/// 3. Releases a lock.
+async fn lock_client_1_operations(
     state_store_client_arc: Arc<state_store::Client<SessionManagedClient>>,
-    mut leased_lock_client: leased_lock::Client<SessionManagedClient>,
+    mut lock_client: lock::Client<SessionManagedClient>,
     exit_handle: SessionExitHandle,
 ) {
     let lock_expiry = Duration::from_secs(10);
@@ -131,10 +121,7 @@ async fn leased_lock_client_1_operations(
     };
 
     // 1.
-    let fencing_token = match leased_lock_client
-        .lock(lock_expiry, request_timeout, None)
-        .await
-    {
+    let fencing_token = match lock_client.lock(lock_expiry, request_timeout, None).await {
         Ok(acquire_result) => {
             log::info!("Lock acquired successfully");
             acquire_result // a.k.a., the fencing token.
@@ -174,7 +161,7 @@ async fn leased_lock_client_1_operations(
     // 3.
 
     // 4.
-    match leased_lock_client.unlock(request_timeout).await {
+    match lock_client.unlock(request_timeout).await {
         Ok(()) => {
             log::info!("Lock released successfully");
         }
@@ -189,87 +176,70 @@ async fn leased_lock_client_1_operations(
     exit_handle.try_exit().await.unwrap();
 }
 
-/// In `leased_lock_client_2_operations` you will find the following examples:
-/// 5. Start observing a lock, waiting for it to be released.
-/// 6. Stop observing a lock.
-/// 7. Make a call to `lock_and_update_value()` to set a key.
-/// 8. Make a call to `lock_and_update_value()` to delete a key.
-async fn leased_lock_client_2_operations(
+/// In `lock_client_2_operations` you will find the following examples:
+/// 4. Waits until lock the lock is acquired.
+/// 5. Sets a key in the State Store using the `fencing_token` obtained from the lock.
+/// 6. Releases the lock.
+async fn lock_client_2_operations(
     state_store_client_arc: Arc<state_store::Client<SessionManagedClient>>,
-    mut leased_lock_client: leased_lock::Client<SessionManagedClient>,
+    mut lock_client: lock::Client<SessionManagedClient>,
     exit_handle: SessionExitHandle,
 ) {
     let lock_expiry = Duration::from_secs(10);
-    let request_timeout = Duration::from_secs(10);
+    let request_timeout = Duration::from_secs(120);
 
     let shared_resource_key_name = b"someKey";
-    let shared_resource_key_value1 = b"someValue1";
     let shared_resource_key_value2 = b"someValue2";
     let shared_resource_key_set_options = SetOptions {
         set_condition: SetCondition::Unconditional,
         expires: Some(Duration::from_secs(15)),
     };
 
-    // 7.
-    match leased_lock_client
-        .lock_and_update_value(
-            lock_expiry,
-            request_timeout,
-            shared_resource_key_name.to_vec(),
-            &|key_current_value: Option<Vec<u8>>| {
-                // Perform some check on `key_current_value`...
-                if key_current_value.unwrap() == shared_resource_key_value1.to_vec() {
-                    AcquireAndUpdateKeyOption::Update(
-                        shared_resource_key_value2.to_vec(),
-                        shared_resource_key_set_options.clone(),
-                    )
-                } else {
-                    AcquireAndUpdateKeyOption::DoNotUpdate
-                    // Handle unexpected value...
-                }
-            },
-        )
-        .await
-    {
-        Ok(lock_and_update_value_result) => {
-            if lock_and_update_value_result.response {
-                log::info!("Key successfully set");
-            } else {
-                log::error!("Could not set key {lock_and_update_value_result:?}");
-                return;
-            }
+    // 4.
+    let fencing_token = match lock_client.lock(lock_expiry, request_timeout, None).await {
+        Ok(acquire_result) => {
+            log::info!("Lock acquired successfully");
+            acquire_result // a.k.a., the fencing token.
         }
         Err(e) => {
-            log::error!("Failed setting key: {e}");
+            log::error!("Failed acquiring lock {e}");
             return;
         }
     };
 
-    // Perform any application-logic and when done, delete key...
-
-    // 8.
-    match leased_lock_client
-        .lock_and_update_value(
-            lock_expiry,
-            request_timeout,
+    // The purpose of the lock is to protect setting a shared key in the state store.
+    // 5.
+    match state_store_client_arc
+        .set(
             shared_resource_key_name.to_vec(),
-            &|_key_current_value| {
-                // Perform some check on `key_current_value`...
-                AcquireAndUpdateKeyOption::Delete
-            },
+            shared_resource_key_value2.to_vec(),
+            request_timeout,
+            Some(fencing_token),
+            shared_resource_key_set_options.clone(),
         )
         .await
     {
-        Ok(lock_and_update_value_result) => {
-            if lock_and_update_value_result.response {
-                log::info!("Key successfully deleted");
+        Ok(set_response) => {
+            if set_response.response {
+                log::info!("Key set successfully");
             } else {
-                log::error!("Could not delete key {lock_and_update_value_result:?}",);
+                log::error!("Could not set key {set_response:?}");
                 return;
             }
         }
         Err(e) => {
-            log::error!("Failed deleting key: {e}");
+            log::error!("Failed setting key {e}");
+            return;
+        }
+    };
+
+    // 6.
+    match lock_client.unlock(request_timeout).await {
+        Ok(()) => {
+            log::info!("Lock released successfully");
+        }
+        Err(e) => {
+            log::error!("Failed releasing lock {e}");
             return;
         }
     };
--- a/rust/azure_iot_operations_services/examples/leased_lock_client.rs
+++ b/rust/azure_iot_operations_services/examples/leased_lock_client.rs
@@ -117,7 +117,7 @@ fn create_clients(
 /// 4. Releases a lock.
 async fn leased_lock_client_1_operations(
     state_store_client_arc: Arc<state_store::Client<SessionManagedClient>>,
-    leased_lock_client: leased_lock::Client<SessionManagedClient>,
+    mut leased_lock_client: leased_lock::Client<SessionManagedClient>,
     exit_handle: SessionExitHandle,
 ) {
     let lock_expiry = Duration::from_secs(10);
@@ -131,7 +131,10 @@ async fn leased_lock_client_1_operations(
     };
 
     // 1.
-    let fencing_token = match leased_lock_client.lock(lock_expiry, request_timeout).await {
+    let fencing_token = match leased_lock_client
+        .lock(lock_expiry, request_timeout, None)
+        .await
+    {
         Ok(acquire_result) => {
             log::info!("Lock acquired successfully");
             acquire_result // a.k.a., the fencing token.
@@ -193,7 +196,7 @@ async fn leased_lock_client_1_operations(
 /// 8. Make a call to `lock_and_update_value()` to delete a key.
 async fn leased_lock_client_2_operations(
     state_store_client_arc: Arc<state_store::Client<SessionManagedClient>>,
-    leased_lock_client: leased_lock::Client<SessionManagedClient>,
+    mut leased_lock_client: leased_lock::Client<SessionManagedClient>,
     exit_handle: SessionExitHandle,
 ) {
     let lock_expiry = Duration::from_secs(10);
--- a/rust/azure_iot_operations_services/src/leased_lock.rs
+++ b/rust/azure_iot_operations_services/src/leased_lock.rs
@@ -17,12 +17,10 @@ pub type LeaseObservation = KeyObservation;
 /// Represents the errors that occur in the Azure IoT Operations State Store Service.
 pub type ServiceError = StateStoreServiceError;
 
-/// Leased Lock Client implementation
-mod client;
-/// Key Lease Client implementation
+/// Lease Client implementation
 pub mod lease;
-
-pub use client::Client;
+/// Lock Client implementation
+pub mod lock;
 
 /// Represents an error that occurred in the Azure IoT Operations Key Lease implementation.
 #[derive(Debug, Error)]
@@ -102,15 +100,4 @@ impl From<state_store::ErrorKind> for ErrorKind {
             state_store::ErrorKind::DuplicateObserve => ErrorKind::DuplicateObserve,
         }
     }
-}
-
-/// Enumeration used as a response for `leased_lock::Client::lock_and_update_value`.
-pub enum AcquireAndUpdateKeyOption {
-    /// Indicates a State Store key shall be updated.
-    /// The first argument is new value for the State Store key.
-    Update(Vec<u8>, SetOptions),
-    /// Indicates the State Store key shall not be updated nor deleted.
-    DoNotUpdate,
-    /// Indicates the State Store key shall be deleted.
-    Delete,
 }
--- a/rust/azure_iot_operations_services/src/leased_lock.rs
+++ b/rust/azure_iot_operations_services/src/leased_lock.rs
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-//! Types for Key Lease operations.
+//! Types for Lease and Lock operations.
 
 use core::fmt::Debug;
 
@@ -11,7 +11,7 @@ use thiserror::Error;
 use crate::state_store::{self, KeyObservation, ServiceError as StateStoreServiceError};
 pub use crate::state_store::{Response, SetCondition, SetOptions};
 
-/// A struct to manage receiving notifications for a lock
+/// A struct to manage receiving notifications for a lease
 pub type LeaseObservation = KeyObservation;
 
 /// Represents the errors that occur in the Azure IoT Operations State Store Service.
@@ -22,7 +22,7 @@ pub mod lease;
 /// Lock Client implementation
 pub mod lock;
 
-/// Represents an error that occurred in the Azure IoT Operations Key Lease implementation.
+/// Represents an error that occurred in the Azure IoT Operations Lease and Lock implementation.
 #[derive(Debug, Error)]
 #[error(transparent)]
 pub struct Error(#[from] ErrorKind);
@@ -42,13 +42,13 @@ impl From<state_store::Error> for Error {
     }
 }
 
-/// Represents the kinds of errors that occur in the Azure IoT Operations Key Lease implementation.
+/// Represents the kinds of errors that occur in the Azure IoT Operations Lease and Lock implementation.
 #[derive(Error, Debug)]
 #[allow(clippy::large_enum_variant)]
 pub enum ErrorKind {
-    /// The lock is already in use by another holder.
-    #[error("lock is already held by another holder")]
-    KeyAlreadyLeased,
+    /// The lease is already in use by another holder.
+    #[error("lease is already held by another holder")]
+    LeaseAlreadyHeld,
     /// An error occurred in the AIO Protocol. See [`AIOProtocolError`] for more information.
     #[error(transparent)]
     AIOProtocolError(#[from] AIOProtocolError),
@@ -58,18 +58,6 @@ pub enum ErrorKind {
     /// The key length must not be zero.
     #[error("key length must not be zero")]
     KeyLengthZero,
-    /// The lease name length must not be zero.
-    #[error("lease name length must not be zero")]
-    LeaseNameLengthZero,
-    /// The lock name length must not be zero.
-    #[error("lock name length must not be zero")]
-    LockNameLengthZero,
-    /// The lease holder name length must not be zero.
-    #[error("lease holder name length must not be zero")]
-    LeaseHolderNameLengthZero,
-    /// The lock holder name length must not be zero.
-    #[error("lock holder name length must not be zero")]
-    LockHolderNameLengthZero,
     /// Fencing token not received from service.
     #[error("Fencing token not received from service")]
     MissingFencingToken,
--- a/rust/azure_iot_operations_services/src/leased_lock.rs
+++ b/rust/azure_iot_operations_services/src/leased_lock.rs
@@ -58,12 +58,18 @@ pub enum ErrorKind {
     /// The key length must not be zero.
     #[error("key length must not be zero")]
     KeyLengthZero,
+    /// The lease name length must not be zero.
+    #[error("lease name length must not be zero")]
+    LeaseNameLengthZero,
     /// The lock name length must not be zero.
     #[error("lock name length must not be zero")]
     LockNameLengthZero,
+    /// The lease holder name length must not be zero.
+    #[error("lease holder name length must not be zero")]
+    LeaseHolderNameLengthZero,
     /// The lock holder name length must not be zero.
     #[error("lock holder name length must not be zero")]
-    LeaseHolderNameLengthZero,
+    LockHolderNameLengthZero,
     /// An error occurred during serialization of a request.
     #[error("{0}")]
     SerializationError(String),
--- a/rust/azure_iot_operations_services/src/leased_lock.rs
+++ b/rust/azure_iot_operations_services/src/leased_lock.rs
@@ -67,8 +67,8 @@ pub enum ErrorKind {
     /// The payload of the response does not match the expected type for the request.
     #[error("Unexpected response payload for the request type: {0}")]
     UnexpectedPayload(String),
-    /// A lock may only have one [`LeaseObservation`] at a time.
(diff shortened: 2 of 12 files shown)


rust/azure_iot_operations_services/src/leased_lock/client.rs
```diff
@@ -54,129 +52,64 @@ where
...
         Ok(Self {
             state_store,
-            lock_name,
-            lock_holder_name,
+            lease_client,
         })
     }
 
-    /// Attempts to acquire a lock, returning if it cannot be acquired after one attempt.
+    /// Waits until a lock is available (if not already) and attempts to acquire it.
```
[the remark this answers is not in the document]
ewertons: Done already.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,416 @@
...
+                                lease_expiration,
+                                request_timeout,
+                                current_fencing_token.clone(),
+                            )
+                            .await
+                            {
+                                Ok(_) => {
+                                    // new fencing token is already saved by `internal_try_acquire()`.
+                                }
+                                Err(_) => {
```
cartertinney, line 195: Since you only want to do something on `Err` perhaps you should use `is_err()` instead of `match`

RLeclair:
I see that code that `leased_lock::Client` has been split into `leased_lock::Client` and `leased_lock::lease::Client`

I believe it would be better to have a module for lease and one for lock so that they can be imported as `lease::Client` and `lock::Client`. So we'd have an extra two folders inside of the lease_lock top level one. 

It'd be a bit odd to use a lock using: `lease_lock::lock::Client` so I would suggest re-exporting the lock and lease modules from lib.rs so it is easier to use

rust/azure_iot_operations_services/src/leased_lock.rs
```diff
@@ -12,17 +12,19 @@ use crate::state_store::{self, KeyObservation, ServiceError as StateStoreService
...
 
 /// Represents the errors that occur in the Azure IoT Operations State Store Service.
 pub type ServiceError = StateStoreServiceError;
 
 /// Leased Lock Client implementation
 mod client;
+/// Key Lease Client implementation
+pub mod lease;
 
 pub use client::Client;
```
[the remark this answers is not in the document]
cartertinney: That's interesting - I like it! Feels like a decision that requires approval from PMs or equivalent decision making though as it affects crate structure fairly significantly.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,416 @@
...
+    /// - `lease_holder_name` is expected to be the client ID used in the underlying MQTT connection settings.
+    /// - There must be one instance of `leased_lock::Client` per lease.
+    ///
+    /// # Errors
+    /// [`struct@Error`] of kind [`KeyLengthZero`](ErrorKind::KeyLengthZero) if the `key_name` is empty
+    ///
+    /// [`struct@Error`] of kind [`LeaseHolderNameLengthZero`](ErrorKind::LeaseHolderNameLengthZero) if the `lease_holder_name` is empty
+    pub fn new(
+        state_store: Arc<state_store::Client<C>>,
+        key_name: Vec<u8>,
```
vaavva, line 50: if a lease is over a key, should this be the key name or the lease name?

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,416 @@
...
+    /// Gets the latest fencing token related to the most recent lease.
+    ///
+    /// Returns either None or an actual Fencing Token (`HybridLogicalClock`).
+    /// None means that either a lease has not been acquired previously with this client, or
+    /// if a lease renewal has failed (if lease auto-renewal is used). The presence of a `HybridLogicalClock`
+    /// does not mean that it is the most recent (and thus valid) Fencing Token - in case
+    /// auto-renewal has not been used and the lease has already expired.
+    #[must_use]
+    pub async fn get_current_lease_fencing_token(&self) -> Option<HybridLogicalClock> {
+        self.current_fencing_token.clone().lock().await.clone()
```
vaavva, line 79: I think this first clone shouldn't be needed, it feels a little weird to be cloning the mutex

cartertinney: It's cloning the Arc I think. But, yes, I agree, this seems very unnecessary

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,416 @@
...
+    ///
+    /// [`struct@Error`] of kind [`KeyAlreadyLeased`](ErrorKind::KeyAlreadyLeased) if the `lease` is already in use by another holder
+    /// # Panics
+    /// Possible panic if, for some error, the fencing token (a.k.a. `version`) of the acquired lease is None.
+    pub async fn try_acquire(
+        &mut self,
+        lease_expiration: Duration,
+        request_timeout: Duration,
+    ) -> Result<HybridLogicalClock, Error> {
+        let acquire_result = Self::internal_try_acquire(
```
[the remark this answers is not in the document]
cartertinney: Also depending on some other changes you have pointed out, perhaps this helper becomes entirely unnecessary

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,416 @@
...
+        match self.unobserve(request_timeout).await {
+            Ok(_) => {
+                if acquire_result.is_ok() {
+                    self.auto_renewal_task = Self::start_lease_auto_renewal(
+                        self.state_store.clone(),
+                        self.key_name.clone(),
+                        self.lease_holder_name.clone(),
+                        lease_expiration,
+                        request_timeout,
+                        renewal_period,
```
vaavva, line 295: I think instead of passing in renewal_period as an option here and potentially returning a dummy task, you could instead only call start_lease_auto_renewal if renewal_period is Some

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,416 @@
...
+pub struct Client<C>
+where
+    C: ManagedClient + Clone + Send + Sync + 'static,
+    C::PubReceiver: Send + Sync,
+{
+    state_store: Arc<state_store::Client<C>>,
+    key_name: Vec<u8>,
+    lease_holder_name: Vec<u8>,
+    current_fencing_token: Arc<Mutex<Option<HybridLogicalClock>>>,
+    auto_renewal_task: JoinHandle<()>,
```
vaavva, line 25: instead of saving a handle to this task, it might be cleaner to have a channel/Notify/cancellation token into the task that can cancel it

rust/azure_iot_operations_services/src/leased_lock.rs
```diff
@@ -12,17 +12,19 @@ use crate::state_store::{self, KeyObservation, ServiceError as StateStoreService
...
-pub type LockObservation = KeyObservation;
+pub type LeaseObservation = KeyObservation;
 
 /// Represents the errors that occur in the Azure IoT Operations State Store Service.
 pub type ServiceError = StateStoreServiceError;
 
 /// Leased Lock Client implementation
 mod client;
+/// Key Lease Client implementation
+pub mod lease;
```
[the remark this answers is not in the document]
ewertons: I do agree.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,416 @@
...
+}
+
+/// Key Lease client implementation
+///
+/// Notes:
+/// Do not call any of the methods of this client after the `state_store` parameter is shutdown.
+/// Calling any of the methods in this implementation after the `state_store` is shutdown results in undefined behavior.
```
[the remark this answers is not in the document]
ewertons: To me it seems line an unresolved aspiration of the language, since it leads us to this kind of construct anyway. I didn't like it either. At lease it was an opportunity to learn more about multi-thread data sharing.

rust/azure_iot_operations_services/src/leased_lock/client.rs
```diff
@@ -6,25 +6,23 @@
...
-    AcquireAndUpdateKeyOption, Error, ErrorKind, LockObservation, Response, SetCondition,
-    SetOptions,
+    AcquireAndUpdateKeyOption, Error, ErrorKind, Response, lease::Client as LeaseClient,
 };
 use crate::state_store::{self};
 use azure_iot_operations_mqtt::interface::ManagedClient;
 use azure_iot_operations_protocol::common::hybrid_logical_clock::HybridLogicalClock;
 
-/// Leased Lock client struct.
+/// Key Lease client struct.
```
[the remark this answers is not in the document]
ewertons: Yes! MIA

rust/azure_iot_operations_services/src/leased_lock.rs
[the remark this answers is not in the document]
ewertons: Hmmm. I hear you, it makes sense from the actual implementation it is based on (state store), but the major point will be how the design is supposed to be made public. Check with the PMs I must.

rust/azure_iot_operations_services/src/leased_lock/client.rs
```diff
@@ -201,14 +134,17 @@ where
     /// [`struct@Error`] of kind [`UnexpectedPayload`](ErrorKind::UnexpectedPayload) if the State Store returns a response that isn't valid for the request
     ///
     /// [`struct@Error`] of kind [`AIOProtocolError`](ErrorKind::AIOProtocolError) if there are any underlying errors from the command invoker
-    pub async fn acquire_lock_and_update_value(
-        &self,
+    pub async fn lock_and_update_value(
```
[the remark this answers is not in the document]
ewertons: It's not on ADR, but it is in the go client implementation. The ADR should be updated to reflect that.

rust/azure_iot_operations_services/src/leased_lock.rs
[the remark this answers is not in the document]
cartertinney: I say for now at least lets at least make a flat structure under `leased_lock` and then figure out the rest later.

rust/azure_iot_operations_services/src/leased_lock.rs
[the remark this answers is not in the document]
ewertons: Done.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,416 @@
...
+/// Key Lease client struct.
+pub struct Client<C>
+where
+    C: ManagedClient + Clone + Send + Sync + 'static,
+    C::PubReceiver: Send + Sync,
+{
+    state_store: Arc<state_store::Client<C>>,
+    key_name: Vec<u8>,
+    lease_holder_name: Vec<u8>,
+    current_fencing_token: Arc<Mutex<Option<HybridLogicalClock>>>,
```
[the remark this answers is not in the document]
ewertons: Yes. The calling app and the internal task to aito-renew the lease.

rust/azure_iot_operations_services/src/leased_lock.rs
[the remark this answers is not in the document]
ewertons: As discussed offline, we went with this for now:

azure_iot_operations_services::leased_lock::lease::Client
azure_iot_operations_services::leased_lock::lock::Client

rust/azure_iot_operations_services/src/leased_lock/lease.rs
ewertons: Changed to lease_name, as per offline discussion.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
ewertons: This all has been reworked. Please review the new code.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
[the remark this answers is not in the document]
ewertons: Code reworked. Please review the newest code.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,416 @@
...
+            Err(_) => {
+                *self.current_fencing_token.clone().lock().await = None;
+            }
+        }
+
+        acquire_result
+    }
+
+    fn start_lease_auto_renewal(
+        state_store: Arc<state_store::Client<C>>,
```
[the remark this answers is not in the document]
ewertons: Code reworked. Please review the newest code.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,416 @@
...
+            self.lease_holder_name.clone(),
+            lease_expiration,
+            request_timeout,
+            self.current_fencing_token.clone(),
+        )
+        .await;
+
+        match acquire_result {
+            Ok(ref version) => {
+                *self.current_fencing_token.clone().lock().await = Some(version.clone());
```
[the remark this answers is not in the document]
ewertons: Code reworked. Please review the newest code.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
ewertons: Yes, that was great advice. Done as we discussed. Thanks.

rust/azure_iot_operations_services/src/leased_lock/client.rs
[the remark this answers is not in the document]
ewertons: This function lock_and_update_value has been removed, as per our discussion.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
ewertons: Code has been reworked, could you please review again? Thanks.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,326 @@
...
+
+            Err(Error(ErrorKind::KeyAlreadyLeased))
+        }
+    }
+
+    /// Attempts to acquire a lease, returning if it cannot be acquired after one attempt.
+    ///
+    /// `lease_expiration` is how long the lease will remain held in the State Store after acquired, if not released before then.
+    /// `request_timeout` is the maximum time the function will wait for receiving a response from the State Store service, it is rounded up to the nearest second.
+    ///
```
vaavva, line 146: add doc for renewal_period as well

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,326 @@
...
+    /// Returns Ok with a fencing token (`HybridLogicalClock`) if completed successfully, or `Error(KeyAlreadyLeased)` if lease is not acquired.
```
vaavva, line 147:
```suggestion
    /// Returns Ok with a fencing token (`HybridLogicalClock`) if completed successfully, or `Error` if lease is not acquired.
```
since there are a few other errors that can be returned

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,326 @@
...
+    #[must_use]
+    pub fn get_current_lease_fencing_token(&self) -> Option<HybridLogicalClock> {
+        self.current_fencing_token
+            .lock()
+            .expect("Could not lock mutex")
+            .clone()
+    }
+
+    async fn internal_acquire(
+        &mut self,
```
vaavva, line 103: mut needed? (If not, same comment for the calling functions)

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,326 @@
...
+        if !self.auto_renewal_cancellation_token.is_cancelled() {
+            self.auto_renewal_cancellation_token.cancel();
+        }
+
+        let acquire_result = self
+            .internal_acquire(lease_expiration, request_timeout)
+            .await;
+
+        if renewal_period.is_some() {
+            self.auto_renewal_cancellation_token = CancellationToken::new();
```
vaavva, line 176: `auto_renewal_cancellation_token` might need to be under a Mutex on the Client struct in this case, otherwise a clone of the Client won't necessarily get the new CancellationToken I think? I'm not sure this is an issue within your implementation since you always cancel the current one and create a new one before cloning the Client (so there's never a cloned client that lasts longer than the lifetime of one token), but if the customer clones the Lease, I think things might be able to get a little funky?

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,326 @@
...
+        if !self.auto_renewal_cancellation_token.is_cancelled() {
+            self.auto_renewal_cancellation_token.cancel();
+        }
+
+        *self
+            .current_fencing_token
+            .lock()
+            .expect("Could not lock mutex") = None;
+
+        match self
```
cartertinney, line 232: You can do away with this whole `match` statement.

```
self.state_store
    .vdel(
        self.lease_name.clone(),
        self.lease_holder_name.clone(),
        None,
        request_timeout,
    )
    .await?;
```

Try to always chain your Options and Results if you can, it makes this much easier with more complex functions. Chain your assumptions into variable assignment so there are no enum wrappers, and use the `?` operator to propagate any errors along the way

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,326 @@
...
+        if renewal_period.is_some() {
+            self.auto_renewal_cancellation_token = CancellationToken::new();
+            let mut self_clone = self.clone();
+
+            tokio::task::spawn({
+                async move {
+                    loop {
+                        select! {
+                            () = self_clone.auto_renewal_cancellation_token.cancelled() => {
+                                break; // Auto-renewal is cancelled.
```
vaavva, line 184: if auto-renewal is cancelled, does the lease need to be released? Or is it desired for the lease to be released at the next lease_expiration interval?

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,326 @@
...
+                            }
+                            () = tokio::time::sleep(renewal_period.unwrap()) => {}
```
vaavva, line 186: won't this panic if renewal_period is None? might be better to wrap the task spawn like this:
```
if let Some(renewal_period) = renewal_period {
    tokio::task::spawn({
        ... existing functionality, without having to unwrap the renewal_period
    })
}
```

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,326 @@
...
+                        }
+
+                        if self_clone
```
vaavva, line 189: to be a bit more explicit about this only happening if cancellation hasn't, put this logic inside of the
```
() = tokio::time::sleep(renewal_period.unwrap()) => {
    <here>
}
```

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,326 @@
...
+    ///
+    /// # Panics
+    /// Possible panic if `std::Lock::lock()` fails.
+    pub async fn release(&mut self, request_timeout: Duration) -> Result<(), Error> {
+        // Stop auto-renewal.
+        if !self.auto_renewal_cancellation_token.is_cancelled() {
+            self.auto_renewal_cancellation_token.cancel();
+        }
+
+        *self
```
vaavva, line 227: should this only happen if the state store delete was successful?

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,326 @@
...
+                self.lease_name.clone(),
+                self.lease_holder_name.clone(),
+                None,
+                request_timeout,
+            )
+            .await
+        {
+            Ok(_) => Ok(()),
+            Err(state_store_error) => Err(state_store_error.into()),
+        }
```
vaavva, line 244: If the above comment about doing stuff on Ok doesn't make sense, then this can be simplified: 
```suggestion
         self
            .state_store
            .vdel(
                self.lease_name.clone(),
                self.lease_holder_name.clone(),
                None,
                request_timeout,
            )
            .await?
        Ok(())
```

rust/azure_iot_operations_services/src/leased_lock/lock.rs
```diff
@@ -0,0 +1,122 @@
...
+// TODO: move client.rs to lock.rs; pub mod for both and no import ... (see comment in leased_lock.rs)
+
+use std::{sync::Arc, time::Duration};
+
+use crate::leased_lock::{Error, ErrorKind, lease::Client as LeaseClient};
+use crate::state_store::{self};
+use azure_iot_operations_mqtt::interface::ManagedClient;
+use azure_iot_operations_protocol::common::hybrid_logical_clock::HybridLogicalClock;
+
+/// Key Lease client struct.
```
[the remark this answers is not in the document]
ewertons: Fixed.

rust/azure_iot_operations_services/src/leased_lock/lock.rs
```diff
@@ -0,0 +1,122 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Client for Lease Lock operations.
+
+// TODO: move client.rs to lock.rs; pub mod for both and no import ... (see comment in leased_lock.rs)
```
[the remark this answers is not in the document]
ewertons: Sorry, the branch was not fully updated. This comment had already been removed.

rust/azure_iot_operations_services/tests/lease_client_tests.rs
```diff
@@ -14,32 +14,27 @@ use azure_iot_operations_mqtt::session::{
     Session, SessionExitHandle, SessionManagedClient, SessionOptionsBuilder,
 };
 use azure_iot_operations_protocol::application::ApplicationContextBuilder;
-use azure_iot_operations_services::leased_lock::{
-    self, AcquireAndUpdateKeyOption, SetCondition, SetOptions,
-};
+use azure_iot_operations_services::leased_lock::lease;
 use azure_iot_operations_services::state_store::{self};
 
 // API:
```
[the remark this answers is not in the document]
ewertons: I agree. I followed Val's guidelines from state_store.

rust/azure_iot_operations_services/src/leased_lock.rs
```diff
@@ -63,7 +63,7 @@ pub enum ErrorKind {
     LockNameLengthZero,
     /// The lock holder name length must not be zero.
     #[error("lock holder name length must not be zero")]
-    LockHolderNameLengthZero,
+    LeaseHolderNameLengthZero,
```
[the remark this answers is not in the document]
cartertinney: Yeah, if we must have invalid argument errors, it should be one enum variant with a string explanation inside it.

rust/azure_iot_operations_services/src/lib.rs
```diff
@@ -11,7 +11,7 @@
 //! - `all`: Enables all features.
 //! - `schema_registry`: Enabled the Schema Registry Client.
 //! - `state_store`: Enabled the State Store Client.
-//! - `leased_lock`: Enables the Leased Lock Client.
+//! - `leased_lock`: Enables the Key Lease Client.
```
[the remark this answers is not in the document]
ewertons: Changed to 

- `leased_lock`: Enables the Lease and Lock Clients.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,326 @@
...
+    /// Returns either None or an actual Fencing Token (`HybridLogicalClock`).
+    /// None means that either a lease has not been acquired previously with this client, or
+    /// if a lease renewal has failed (if lease auto-renewal is used). The presence of a `HybridLogicalClock`
+    /// does not mean that it is the most recent (and thus valid) Fencing Token - in case
+    /// auto-renewal has not been used and the lease has already expired.
+    ///
+    /// # Panics
+    /// Possible panic if `std::Lock::lock()` fails.
+    #[must_use]
+    pub fn get_current_lease_fencing_token(&self) -> Option<HybridLogicalClock> {
```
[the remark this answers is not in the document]
ewertons: Renamed

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,326 @@
...
+        self.current_fencing_token
+            .lock()
+            .expect("Could not lock mutex")
```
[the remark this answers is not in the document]
ewertons: Done.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,326 @@
...
+
+        let state_store_response = a?;
+
+        if state_store_response.response {
+            self.current_fencing_token
+                .lock()
+                .expect("Could not lock mutex")
+                .clone_from(&state_store_response.version);
+
+            Ok(state_store_response.version.expect(
```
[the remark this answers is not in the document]
ewertons: After revisiting it, I do agree.  Changed to return an error.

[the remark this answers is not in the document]
ewertons: Done.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,326 @@
...
+                request_timeout,
+                None,
+                SetOptions {
+                    set_condition: SetCondition::OnlyIfEqualOrDoesNotExist,
+                    expires: Some(lease_expiration),
+                },
+            )
+            .await;
+
+        let state_store_response = a?;
```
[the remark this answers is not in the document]
ewertons: Oops, that was for debugging something. Fixed.

rust/azure_iot_operations_services/src/lib.rs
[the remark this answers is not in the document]
ewertons: Done.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,326 @@
...
+
+//! Client for Key Lease operations.
+
+use std::{sync::Arc, sync::Mutex, time::Duration};
+
+use tokio::select;
+use tokio_util::sync::CancellationToken;
+
+use crate::leased_lock::{Error, ErrorKind, LeaseObservation, Response, SetCondition, SetOptions};
+use crate::state_store::{self};
```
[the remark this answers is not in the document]
ewertons: Indeed. I thought the idiomatic way is to import the most specific component being used.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,326 @@
...
+
+impl<C> Drop for Client<C>
+where
+    C: ManagedClient + Clone + Send + Sync,
+    C::PubReceiver: Send + Sync,
+{
+    fn drop(&mut self) {
+        if !self.auto_renewal_cancellation_token.is_cancelled() {
+            self.auto_renewal_cancellation_token.cancel();
+        }
```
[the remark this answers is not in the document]
ewertons: Yeah, confirmed that it's not needed: https://docs.rs/tokio-util/latest/src/tokio_util/sync/cancellation_token/tree_node.rs.html#300

All the checks removed as suggested.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,326 @@
...
+impl<C> Client<C>
+where
+    C: ManagedClient + Clone + Send + Sync,
+    C::PubReceiver: Send + Sync,
+{
+    /// Create a new Key Lease Client.
+    ///
+    /// Notes:
+    /// - `lease_holder_name` is expected to be the client ID used in the underlying MQTT connection settings.
+    /// - There must be one instance of `leased_lock::Client` per lease.
```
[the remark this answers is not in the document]
ewertons: That is so true! Changed.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,326 @@
...
+    /// Gets the latest fencing token related to the most recent lease.
+    ///
+    /// Returns either None or an actual Fencing Token (`HybridLogicalClock`).
+    /// None means that either a lease has not been acquired previously with this client, or
+    /// if a lease renewal has failed (if lease auto-renewal is used). The presence of a `HybridLogicalClock`
+    /// does not mean that it is the most recent (and thus valid) Fencing Token - in case
+    /// auto-renewal has not been used and the lease has already expired.
+    ///
+    /// # Panics
+    /// Possible panic if `std::Lock::lock()` fails.
```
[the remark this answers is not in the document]
ewertons: Taken.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
ewertons: Interesting. I though because a field is being set, `mut` would be needed. I need to reconcile that.

Changed.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
ewertons: The auto-renewal task is cancelled in two scenarios:
1. release() is called by the user app
2. the lease client is dropped.

On the first scenario it matches what you suggest, but on the second the lease remains until it expires. My logic was that if the customer has not called to intentionally release the lock, it should remain. Maybe the app crashed and they want to get a chance to reacquire it. Does that sound like a fair use case for you?

rust/azure_iot_operations_services/src/leased_lock/lease.rs
ewertons: We are already checking if `renewal_period` `is_some()` on line 170.
But your construct is more elegant, so I'll take it.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
ewertons: Done.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
ewertons: Done.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
ewertons: Done.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
ewertons: Done.

rust/azure_iot_operations_services/src/leased_lock/lock.rs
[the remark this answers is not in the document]
ewertons: This was there because the branch online was not updated for review. Apologies for that. The branch is now up-to-date.

rust/azure_iot_operations_services/src/leased_lock/lock.rs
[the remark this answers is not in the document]
ewertons: This was there because the branch online was not updated for review. Apologies for that. The branch is now up-to-date.

rust/azure_iot_operations_services/src/leased_lock/lock.rs
```diff
@@ -0,0 +1,122 @@
...
+    /// [`struct@Error`] of kind [`AIOProtocolError`](ErrorKind::AIOProtocolError) if there are any underlying errors from the command invoker
+    /// # Panics
+    /// Possible panic if, for some error, the fencing token (a.k.a. `version`) of the acquired lease is None.
+    pub async fn lock(
+        &mut self,
+        lock_expiration: Duration,
+        request_timeout: Duration,
+        renewal_period: Option<Duration>,
+    ) -> Result<HybridLogicalClock, Error> {
+        self.lease_client
```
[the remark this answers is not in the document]
ewertons: This was there because the branch online was not updated for review. Apologies for that. The branch is now up-to-date.

rust/azure_iot_operations_services/src/leased_lock/lock.rs
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        let lease_client = LeaseClient::new(state_store, lock_name, lock_holder_name)?;
+
+        Ok(Self { lease_client })
+    }
+
+    /// Waits until a lock is available (if not already) and attempts to acquire it.
+    ///
+    /// If a non-zero `Duration` is provided as `renewal_period`, the lock is automatically renewed
```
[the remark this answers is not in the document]
ewertons: The lock is not renewed. I think that's clear per the documentation, don't you think?

I do agree that renewing every second will not do anyone any good. A few more considerations:

1. There should be a minimum expected renew period? If yes, how much?
2. The auto-renewal period should not be longer than the actual lease expiration, I believe. We should check for that in the function.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,326 @@
...
+    ///
+    /// [`struct@Error`] of kind [`ServiceError`](ErrorKind::ServiceError) if the State Store returns an Error response
+    ///
+    /// [`struct@Error`] of kind [`UnexpectedPayload`](ErrorKind::UnexpectedPayload) if the State Store returns a response that isn't valid for a `Get` request
+    ///
+    /// [`struct@Error`] of kind [`AIOProtocolError`](ErrorKind::AIOProtocolError) if there are any underlying errors from the command invoker
+    pub async fn get_holder(
+        &self,
+        request_timeout: Duration,
+    ) -> Result<Response<Option<Vec<u8>>>, Error> {
```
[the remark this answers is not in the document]
ewertons: I see. Yeah, we could unwrap it, since the error will be propagated by ?. I makes sense. I removed as suggested from get_holder and observe and unobserve

rust/azure_iot_operations_services/examples/leased_lock_client.rs
```diff
@@ -4,57 +4,39 @@
...
-};
+use azure_iot_operations_services::leased_lock::{SetCondition, SetOptions, lock};
 use azure_iot_operations_services::state_store;
 
 #[tokio::main(flavor = "current_thread")]
 async fn main() {
     let client_id1 = "someClientId1";
     let client_id2 = "someClientId2";
-    let lock_name = "someLock";
+    let key_name = "someLock";
```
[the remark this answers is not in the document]
ewertons: Done.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,307 @@
...
+                .unwrap()
+                .clone_from(&state_store_response.version);
+
+            state_store_response
+                .version
+                .ok_or(Error(ErrorKind::MissingFencingToken))
+        } else {
+            *self.current_fencing_token.lock().unwrap() = None;
+
+            Err(Error(ErrorKind::KeyAlreadyLeased))
```
vaavva, line 128:
```suggestion
            Err(Error(ErrorKind::LeaseAlreadyHeld))
```
to keep away from the Key semantics?

rust/azure_iot_operations_services/src/leased_lock.rs
```diff
@@ -42,13 +42,13 @@ impl From<state_store::Error> for Error {
     }
 }
 
-/// Represents the kinds of errors that occur in the Azure IoT Operations Leased Lock implementation.
+/// Represents the kinds of errors that occur in the Azure IoT Operations Key Lease implementation.
 #[derive(Error, Debug)]
 #[allow(clippy::large_enum_variant)]
 pub enum ErrorKind {
     /// The lock is already in use by another holder.
     #[error("lock is already held by another holder")]
```
vaavva, line 50: I think this description will need to be updated

rust/azure_iot_operations_services/src/leased_lock/lease.rs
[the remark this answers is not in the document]
vaavva: same comment for the lock file

rust/azure_iot_operations_services/src/leased_lock/lease.rs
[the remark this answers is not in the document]
vaavva: update in release() doc as well

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,307 @@
...
+        }
+    }
+
+    /// Attempts to acquire a lease, returning if it cannot be acquired after one attempt.
+    ///
+    /// `lease_expiration` is how long the lease will remain held in the State Store after acquired, if not released before then.
+    /// `request_timeout` is the maximum time the function will wait for receiving a response from the State Store service, it is rounded up to the nearest second.
+    /// `renewal_period` is the frequency with which the lease will be auto-renewed by the lease client if acquired succesfully. `None` indicates the lease should not be auto-renewed.
```
vaavva, line 136:
```suggestion
    /// `renewal_period` is the frequency with which the lease will be auto-renewed by the lease client if acquired successfully. `None` indicates the lease should not be auto-renewed.
```

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,307 @@
...
+    ///
+    /// [`struct@Error`] of kind [`ServiceError`](ErrorKind::ServiceError) if the State Store returns an Error response
+    ///
+    /// [`struct@Error`] of kind [`UnexpectedPayload`](ErrorKind::UnexpectedPayload) if the State Store returns a response that isn't valid for a `V Delete` request
+    ///
+    /// [`struct@Error`] of kind [`AIOProtocolError`](ErrorKind::AIOProtocolError) if there are any underlying errors from the command invoker
+    ///
+    /// # Panics
+    /// Possible panic if `std::Lock::lock()` fails.
+    pub async fn release(&mut self, request_timeout: Duration) -> Result<(), Error> {
```
vaavva, line 213: don't think you'll need &mut self here either! (and then that will extend to unlock() as well)

rust/azure_iot_operations_services/src/leased_lock/lease.rs
vaavva: it's because the field is under an Arc<Mutex> - technically you aren't modifying the Arc, you're just changing what it references to

rust/azure_iot_operations_services/src/leased_lock/lease.rs
vaavva: If this changes, then I think that will also remove the need for &mut self in this fn signature

rust/azure_iot_operations_services/src/leased_lock/lease.rs
vaavva: yep, I think this makes sense!

For the scenario where the app crashes, there won't be any errors with them calling acquire(same lease_duration) again, right? It would just behave as an early lease renewal (which you're right, this is the ideal behavior!)

rust/azure_iot_operations_services/src/leased_lock/lease.rs
vaavva: I think this makes sense - even if the operation wasn't successful, you still don't want to be holding the lease anymore. So you'd probably re-try this operation, but you wouldn't want any of your other code to try and use the fencing token. I think my concern was holding different state than the state store, but you're right that we don't have any super strong guarantee that the lease is still held in state store, so it's better to be conservative and assume we don't. Calling acquire when not technically needed won't be a harmful operation.
(keeping this item open just to track adding the doc, please resolve this comment once you've done that!)

rust/azure_iot_operations_services/src/leased_lock/lease.rs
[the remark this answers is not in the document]
vaavva: update docs for these that still refer to Response<>

rust/azure_iot_operations_services/src/leased_lock/lock.rs
```diff
@@ -0,0 +1,162 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Client for Lock operations.
+
+use std::{sync::Arc, time::Duration};
+
+use crate::leased_lock::{Error, ErrorKind, lease::Client as LeaseClient};
```
vaavva, line 8:
```suggestion
use crate::leased_lock::{Error, ErrorKind, lease};
```
you're not using it that many places, but I think for clarity and symmetry with how we generally use the clients that look like this, just refer to it as lease::Client in the code

rust/azure_iot_operations_services/src/leased_lock/lease.rs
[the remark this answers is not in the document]
vaavva: update in lock.rs as well

rust/azure_iot_operations_services/src/leased_lock/lock.rs
```diff
@@ -0,0 +1,162 @@
...
+    ) -> Result<Self, Error> {
+        if lock_name.is_empty() {
+            return Err(Error(ErrorKind::LockNameLengthZero));
+        }
+
+        if lock_holder_name.is_empty() {
+            return Err(Error(ErrorKind::LockHolderNameLengthZero));
+        }
+
+        let lease_client = LeaseClient::new(state_store, lock_name, lock_holder_name)?;
```
vaavva, line 55: any errors that can be returned from this `?` that should be in the doc?

rust/azure_iot_operations_services/src/leased_lock/lock.rs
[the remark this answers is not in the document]
vaavva: 1. I think 1 second would be a reasonable minimum - this is what we have for message expiry. It's maybe a little sillier here, but it's still a semi-reasonable floor of "can't happen faster than the network operations"
2. Great catch, I agree that we should validate this! Otherwise that behavior is suuuper weird. I don't know that we should validate more strictly that auto-renewal period < (lease expiration - something), but maybe it would be good to add doc saying we don't recommend setting them to the same or super close values, as it could result in the lease expiring before auto-renewal can complete

I'm wondering too if a duration of 0 means the same as None, then do we need the Option, or can we just say use 0 for no renewal?

rust/azure_iot_operations_services/src/leased_lock/lease.rs
```diff
@@ -0,0 +1,307 @@
...
+    /// [`struct@Error`] of kind [`UnexpectedPayload`](ErrorKind::UnexpectedPayload) if the State Store returns a response that isn't valid for a `Set` request
+    ///
+    /// [`struct@Error`] of kind [`AIOProtocolError`](ErrorKind::AIOProtocolError) if there are any underlying errors from the command invoker
+    ///
+    /// [`struct@Error`] of kind [`KeyAlreadyLeased`](ErrorKind::KeyAlreadyLeased) if the `lease` is already in use by another holder
+    ///
+    /// [`struct@Error`] of kind [`MissingFencingToken`](ErrorKind::MissingFencingToken) if the fencing token in the service response is empty.
+    ///
+    /// # Panics
+    /// Possible panic if, for some error, the fencing token (a.k.a. `version`) of the acquired lease is None.
```
vaavva, line 153: This is changed to an error now, so update doc to remove the Panics section and add a doc for the new error (same on lock() doc)

rust/azure_iot_operations_services/src/leased_lock/lock.rs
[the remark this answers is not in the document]
clecompt-msft: I feel like customers have asked for sub-second failover, which means we'd need a lease duration (and thus a lock renewal) of less than 1s. Agree on validating the second point, though. Probably something we should add to the other languages too, though I don't know if it's particularly pressing.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
cartertinney: Ah, you'd have to use `and` as well, my mistake.

```
self.state_store
    .vdel(
        self.lease_name.clone(),
        self.lease_holder_name.clone(),
        None,
        request_timeout,
    )
    .await
    .and(Ok())?;
```

cartertinney (commented):
LGTM

rust/azure_iot_operations_services/src/leased_lock/lease.rs
ewertons: This has not changed as we changed the return to remove the Response. So in summary it now does not have a match anymore.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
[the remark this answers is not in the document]
ewertons: Done.

rust/azure_iot_operations_services/src/leased_lock.rs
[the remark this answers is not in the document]
ewertons: Alright, I removed the many LengthZero enum values (for lease and lock only) and used InvalidArgument.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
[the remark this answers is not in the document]
ewertons: Done.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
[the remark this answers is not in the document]
ewertons: Done.

rust/azure_iot_operations_services/src/leased_lock/lock.rs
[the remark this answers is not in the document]
ewertons: Ok, so this is what I implemented based on comments above:
1. Do not set a min valid for renew period. I kept the option there so users can pass None instead of Duraction::ZERO. Zero means None as well (added to documentation)
2. Validate if renewal_period < lease_expiration, if renewal_period is provided.

rust/azure_iot_operations_services/src/leased_lock.rs
```diff
@@ -12,17 +12,17 @@ use crate::state_store::{self, KeyObservation, ServiceError as StateStoreService
 pub use crate::state_store::{Response, SetCondition, SetOptions};
 
 /// A struct to manage receiving notifications for a lock
```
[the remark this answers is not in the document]
ewertons: Changed the doc

rust/azure_iot_operations_services/src/leased_lock.rs
```diff
@@ -42,13 +42,13 @@ impl From<state_store::Error> for Error {
     }
 }
 
-/// Represents the kinds of errors that occur in the Azure IoT Operations Leased Lock implementation.
+/// Represents the kinds of errors that occur in the Azure IoT Operations Key Lease implementation.
```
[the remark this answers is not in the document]
ewertons: Taken

rust/azure_iot_operations_services/src/leased_lock.rs
```diff
@@ -58,12 +58,21 @@ pub enum ErrorKind {
     /// The key length must not be zero.
     #[error("key length must not be zero")]
     KeyLengthZero,
```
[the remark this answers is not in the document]
ewertons: It is still used in state_store.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
ewertons: Done.

rust/azure_iot_operations_services/src/leased_lock.rs
ewertons: Done.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
ewertons: Gramarly

rust/azure_iot_operations_services/src/leased_lock/lock.rs
ewertons: Done.

rust/azure_iot_operations_services/src/leased_lock/lock.rs
ewertons: No, the only errors are from key or holder being empty, and they are covered already.

rust/azure_iot_operations_services/src/leased_lock/lease.rs
ewertons: Gahh. ADHD is tough.

rust/azure_iot_operations_services/src/leased_lock/lock.rs
```diff
@@ -0,0 +1,122 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Client for Lease Lock operations.
+
+// TODO: move client.rs to lock.rs; pub mod for both and no import ... (see comment in leased_lock.rs)
+
+use std::{sync::Arc, time::Duration};
+
+use crate::leased_lock::{Error, ErrorKind, lease::Client as LeaseClient};
```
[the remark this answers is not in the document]
ewertons: I agree. But... for simplicity, we will keep it. This code will probably not result in many future changes anyway.

rust/azure_iot_operations_services/src/leased_lock.rs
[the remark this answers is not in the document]
ewertons: Ok, changed in state store to use InvalidArgument

rust/azure_iot_operations_services/src/leased_lock/lease.rs
ewertons: Change CancellationToken for Notify in lease client. Also removed &mut in argument

rust/azure_iot_operations_services/src/leased_lock/lock.rs
```diff
@@ -67,7 +68,16 @@ where
...
     /// fencing token (see function documentation).
     ///
-    /// Note: `request_timeout` is rounded up to the nearest second.
+    /// Notes:
+    /// `request_timeout` is rounded up to the nearest second.
+    ///
+    /// If lock auto-renewal is used, an auto-renewal task is spawned.
+    /// To terminate this task and stop the lock auto-renewal, `lock::Client::unlock()` must be called.
+    /// Simply dropping the `lock::Client` instance will not terminate the auto-renewal task.
+    /// This logic is intended for a scenario where the `lock::Client` is cloned and a lease is acquired with auto-renewal by the original instance.
```
vaavva, line 77:
```suggestion
    /// This logic is intended for a scenario where the `lock::Client` is cloned and a lock is acquired with auto-renewal by the original instance.
```

rust/azure_iot_operations_services/src/state_store/client.rs
```diff
@@ -192,7 +192,7 @@ where
     ///
     /// Returns `true` if the `Set` completed successfully, or `false` if the `Set` did not occur because of values specified in `SetOptions`
     /// # Errors
-    /// [`struct@Error`] of kind [`KeyLengthZero`](ErrorKind::KeyLengthZero) if the `key` is empty
+    /// [`struct@Error`] of kind [`InvalidArgument`](ErrorKind::InvalidArgument) if the `key` is empty
     ///
     /// [`struct@Error`] of kind [`InvalidArgument`](ErrorKind::InvalidArgument) if the `timeout` is zero or > `u32::max`
```
vaavva, line 197:
```suggestion
    /// [`struct@Error`] of kind [`InvalidArgument`](ErrorKind::InvalidArgument) if:
    /// - the `key` is empty
    /// - the `timeout` is zero or > `u32::max`
```

vaavva: (same comment for other instances in this file)

rust/azure_iot_operations_services/src/leased_lock.rs
```diff
@@ -73,7 +67,7 @@ pub enum ErrorKind {
     /// The payload of the response does not match the expected type for the request.
     #[error("Unexpected response payload for the request type: {0}")]
     UnexpectedPayload(String),
-    /// A lock may only have one [`LockObservation`] at a time.
+    /// A lock may only have one [`LeaseObservation`] at a time.
```
vaavva, line 70: is this for lease or lock? You've got some mismatched wording

rust/azure_iot_operations_services/src/leased_lock/lock.rs
```diff
@@ -0,0 +1,177 @@
...
+    /// # Errors
+    /// [`struct@Error`] of kind [`InvalidArgument`](ErrorKind::InvalidArgument) if the `request_timeout` is zero or > `u32::max`
+    ///
+    /// [`struct@Error`] of kind [`ServiceError`](ErrorKind::ServiceError) if the State Store returns an Error response
+    ///
+    /// [`struct@Error`] of kind [`UnexpectedPayload`](ErrorKind::UnexpectedPayload) if the State Store returns a response that isn't valid for the request
+    ///
+    /// [`struct@Error`] of kind [`AIOProtocolError`](ErrorKind::AIOProtocolError) if there are any underlying errors from the command invoker
+    pub async fn lock(
+        &mut self,
```
cartertinney, line 92: This doesn't need to be mut anymore either as far as I can tell.

Should also then change in sample

vaavva (approved):

vaavva:
Forgot as part of my review, but if you haven't yet, run `cargo doc` and make sure there aren't any new errors from this and then when it generates everything looks nice!

rust/azure_iot_operations_services/src/state_store/client.rs
ewertons: Done.

rust/azure_iot_operations_services/src/leased_lock.rs
ewertons: For lease. Fixed.

rust/azure_iot_operations_services/src/leased_lock/lock.rs
ewertons: Fixed.

RLeclair (approved):
LGTM Ewerton, thanks for addressing the comments!

cartertinney (approved):

ewertons:
Manual run:
https://github.com/Azure/iot-operations-sdks/actions/runs/14938258631

PR #761, opened 2025-04-30
[Rust][Services] : ADR Client APIs

olivakar:


rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1730 @@
...
+    ///
+    /// # Errors
+    /// [`struct@Error`] of kind [`InvalidRequestArgument`](ErrorKind::InvalidRequestArgument)
+    /// if there is error building the request.
+    ///
+    /// [`struct@Error`] of kind [`AIOProtocolError`](ErrorKind::AIOProtocolError)
+    /// if topic tokens are not found.
+    ///
+    /// [`struct@Error`] of kind [`AIOProtocolError`](ErrorKind::AIOProtocolError)
+    /// if topic tokens are not found.
```
vaavva, line 759:
```suggestion
    /// [`struct@Error`] of kind [`AIOProtocolError`](ErrorKind::AIOProtocolError)
    /// if device, inbound_endpoint, or asset names are invalid.
```
duplicate. Also, topic tokens don't mean anything to the caller of this fn, so be more descriptive about what can cause the error. From these descriptions, I'm not sure what causes this error vs InvalidRequestArgument either

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1730 @@
...
+        )
+        .await;
+        assert!(matches!(
+            result.unwrap_err().kind(),
+            ErrorKind::InvalidRequestArgument(_)
+        ));
+    }
+
+    #[tokio::test]
+    async fn test_update_asset_status_empty_asset_name() {
```
vaavva, line 1286: any way to use test-case to combine some of these tests? Seems like a lot of them are pretty similar

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1730 @@
...
+        assert_eq!(hashed, "device1~endpoint1");
+
+        let unhashed = Client::<SessionManagedClient>::un_hash_device_endpoint(&hashed);
+        assert!(unhashed.is_some());
+
+        let (unhashed_device, unhashed_endpoint) = unhashed.unwrap();
+        assert_eq!(unhashed_device, device_name);
+        assert_eq!(unhashed_endpoint, endpoint_name);
+
+        // Test unhashing with invalid input
```
vaavva, line 1722: split this to a separate test

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1730 @@
...
+                },
+                asset_update_message = asset_update_telemetry_receiver.recv() => {
+                    match asset_update_message {
+                        Some(Ok((asset_update_telemetry, ack_token))) => {
+                            let Some(device_name) = asset_update_telemetry.topic_tokens.get(DEVICE_NAME_RECEIVED_TOPIC_TOKEN) else {
+                                log::error!("Asset Update Notification missing {DEVICE_NAME_RECEIVED_TOPIC_TOKEN} topic token.");
+                                continue;
+                            };
+                            let Some(inbound_endpoint_name) = asset_update_telemetry.topic_tokens.get(INBOUND_ENDPOINT_NAME_RECEIVED_TOPIC_TOKEN) else {
+                                log::error!("AssetUpdateEventTelemetry missing {INBOUND_ENDPOINT_NAME_RECEIVED_TOPIC_TOKEN} topic token.");
```
vaavva, line 423:
```suggestion
                                log::error!("Asset Update Notification missing {INBOUND_ENDPOINT_NAME_RECEIVED_TOPIC_TOKEN} topic token.");
```

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1730 @@
...
+                            let receiver_id = Self::hash_device_endpoint_asset(device_name, inbound_endpoint_name, &asset_update_telemetry.payload.asset_update_event.asset_name);
+                            match asset_update_notification_dispatcher.dispatch(&receiver_id, (asset_update_telemetry.payload.asset_update_event.asset.into(), ack_token)) {
+                                Ok(()) => {
+                                    log::debug!("Asset Update Notification dispatched for device {device_name:?}, inbound endpoint {inbound_endpoint_name:?}, and asset {:?}", asset_update_telemetry.payload.asset_update_event.asset_name);
+                                }
+                                Err(DispatchError::SendError(payload)) => {
+                                    log::warn!("Asset Update Observation has been dropped. Received Asset Update Notification: {payload:?}",);
+                                }
+                                Err(DispatchError::NotFound(payload)) => {
+                                    log::warn!("Asset is not being observed. Received Asset Update Notification: {payload:?}",);
```
vaavva, line 437:
```suggestion
                                Err(DispatchError::NotFound((receiver_id, (payload, _)))) => {
                                    log::warn!("Asset is not being observed. Received Asset Update Notification: {payload:#?} for {receiver_id:?}",);
```

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
vaavva: please apply this comment to the docs for all of the other functions - the error info needs to be actionable for the caller

vaavva (commented):
partial review, will get to the rest later this afternoon

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1730 @@
...
+        let asset_shutdown_notifier = Arc::new(Notify::new());
+
+        let mut device_receiver_closed = false;
+        let mut asset_receiver_closed = false;
+
+        loop {
+            tokio::select! {
+                () = shutdown_notifier.notified() => {
+                    if device_shutdown_attempt_count < 3 {
+                        device_shutdown_attempt_count += 1;
```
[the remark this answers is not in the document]
vaavva: Didn't get removed

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1652 @@
...
+    /// # Arguments
+    /// * `device_name` - The name of the Device.
+    /// * `inbound_endpoint_name` - The name of the inbound endpoint.
+    /// * `timeout` - The duration until the Client stops waiting for a response to the request, it is rounded up to the nearest second.
+    ///
+    /// Returns a [`Device`] if the device was found.
+    ///
+    /// # Errors
+    /// [`struct@Error`] of kind [`InvalidRequestArgument`](ErrorKind::InvalidRequestArgument)
+    /// if timeout is 0.
```
vaavva, line 465:
```suggestion
    /// if timeout is 0 or > `u32::max`.
```

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1652 @@
...
+    ///
+    /// [`struct@Error`] of kind [`AIOProtocolError`](ErrorKind::AIOProtocolError)
+    /// if device, inbound endpoint are invalid.
+    ///
+    /// [`struct@Error`] of kind [`AIOProtocolError`](ErrorKind::AIOProtocolError)
+    /// if there are any underlying errors from the AIO RPC protocol.
```
vaavva, line 471:
```suggestion
    /// [`struct@Error`] of kind [`AIOProtocolError`](ErrorKind::AIOProtocolError) if:
    /// - device or inbound endpoint names are invalid.
    /// - there are any underlying errors from the AIO RPC protocol.
```
just a style preference thing to be a bit more concise when the same error can be caused by different scenarios

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1652 @@
...
+    /// # Arguments
+    /// * `device_name` - The name of the Device.
+    /// * `inbound_endpoint_name` - The name of the inbound endpoint.
+    /// * `timeout` - The duration until the Client stops waiting for a response to the request, it is rounded up to the nearest second.
+    ///
+    /// Returns the [`DeviceUpdateObservation`] if the observation was started successfully or [`Error`].
+    ///
+    /// # Errors
+    /// [`struct@Error`] of kind [`DuplicateObserve`](ErrorKind::DuplicateObserve)
+    /// if the receiver is already registered.
```
vaavva, line 550:
```suggestion
    /// if the device is already being observed.
```
receiver and register aren't relevant terms for the user of this API

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1652 @@
...
+    /// if timeout is 0.
+    ///
+    /// [`struct@Error`] of kind [`AIOProtocolError`](ErrorKind::AIOProtocolError)
+    /// if there are any underlying errors from the AIO RPC protocol.
+    ///
+    /// [`struct@Error`] of kind [`AIOProtocolError`](ErrorKind::AIOProtocolError)
+    /// if device, inbound endpoint are invalid.
+    ///
+    /// [`struct@Error`] of kind [`ObservationError`](ErrorKind::ObservationError)
+    /// if the observation was not successful.
```
vaavva, line 562:
```suggestion
    /// if the observation was not accepted by the service.
```
any of these errors signify that the observation wasn't successful

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1652 @@
...
+                .map_err(ErrorKind::from)?;
+        let response = self
+            .update_device_status_command_invoker
+            .invoke(update_device_status_request)
+            .await
+            .map_err(ErrorKind::from)?;
+        Ok(response.payload.updated_device.into())
+    }
+
+    /// Starts observation of any Device updates from the Azure Device Registry service.
```
vaavva, line 537:
```suggestion
    /// Starts observation of a Device's updates from the Azure Device Registry service.
```

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1652 @@
...
+                    log::debug!(
+                        "Device `{device_name:?}` with inbound endpoint `{inbound_endpoint_name:?}` not in observed list"
+                    );
+                }
+                Err(Error(ErrorKind::from(e)))
+            }
+        }
+    }
+
+    /// Stops observation of any Device updates from the Azure Device Registry service.
```
vaavva, line 639:
```suggestion
    /// Stops observation of a Device's updates from the Azure Device Registry service.
```

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1652 @@
...
+
+    /// Retrieves an asset from a Azure Device Registry service.
+    ///
+    /// # Arguments
+    /// * `device_name` - The name of the Device.
+    /// * `inbound_endpoint_name` - The name of the inbound endpoint.
+    /// * `asset_name` - The name of the asset.
+    /// * `timeout` - The duration until the Client stops waiting for a response to the request, it is rounded up to the nearest second.
+    ///
+    /// Returns a [`Asset`] if the the asset was found.
```
vaavva, line 736:
```suggestion
    /// Returns an [`Asset`] if the the asset was found.
```

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1652 @@
...
+                "test-device".to_string(),
+                "test-endpoint".to_string(),
+                "".to_string(),
+                Duration::from_secs(10),
+            )
+            .await;
+
+        assert!(matches!(
+            result.unwrap_err(),
+            Error(ErrorKind::AIOProtocolError(_))
```
vaavva, line 1207: add the AIOProtocolError kind matching here too

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1652 @@
...
+        let response = self
+            .update_asset_status_command_invoker
+            .invoke(command_request)
+            .await
+            .map_err(ErrorKind::from)?;
+
+        Ok(response.payload.updated_asset.into())
+    }
+
+    /// Starts observation of any Asset updates from the Azure Device Registry service.
```
vaavva, line 823:
```suggestion
    /// Starts observation of an Asset's updates from the Azure Device Registry service.
```

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1652 @@
...
+    ///
+    /// Note: On cleanup, unobserve should always be called so that the service knows to stop sending notifications.
+    ///
+    /// # Arguments
+    /// * `device_name` - The name of the Device.
+    /// * `inbound_endpoint_name` - The name of the inbound endpoint.
+    /// * `asset_name` - The name of the asset.
+    /// * `timeout` - The duration until the Client stops waiting for a response to the request, it is rounded up to the nearest second.
+    ///
+    /// Returns a [`AssetUpdateObservation`].
```
vaavva, line 833:
```suggestion
    /// Returns an [`AssetUpdateObservation`].
```

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1652 @@
...
+
+    /// Stops observation of any Asset updates from the Azure Device Registry service.
+    ///
+    /// # Arguments
+    /// * `device_name` - The name of the Device.
+    /// * `inbound_endpoint_name` - The name of the inbound endpoint.
+    /// * `asset_name` - The name of the asset.
+    /// * `timeout` - The duration until the Client stops waiting for a response to the request, it is rounded up to the nearest second.
+    ///
+    /// Returns `Ok(())` if the Device Updates are no longer being observed.
```
vaavva, line 938:
```suggestion
    /// Returns `Ok(())` if the Asset Updates are no longer being observed.
```

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1652 @@
...
+
+        super::Client::new(
+            ApplicationContextBuilder::default().build().unwrap(),
+            managed_client,
+            super::ClientOptionsBuilder::default().build().unwrap(),
+        )
+        .unwrap()
+    }
+
+    // async fn call_update_asset_status(
```
vaavva: remove these lines before you merge

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1652 @@
...
+                "test-device".to_string(),
+                "test-endpoint".to_string(),
+                "".to_string(),
+                Duration::from_secs(10),
+            )
+            .await;
+
+        assert!(matches!(
+            result.unwrap_err().0,
+            ErrorKind::AIOProtocolError(ref e) if matches!(e.kind, AIOProtocolErrorKind::ClientError)
```
vaavva, line 1094: ClientError should only be for network failures, is this the right behavior to validate?

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1652 @@
...
+                "test-endpoint".to_string(),
+                "".to_string(),
+                AssetStatus::default(),
+                Duration::from_secs(10),
+            )
+            .await;
+
+        assert!(matches!(
+            result.unwrap_err(),
+            Error(ErrorKind::AIOProtocolError(_))
```
vaavva, line 1150: add the AIOProtocolError kind matching here too

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
vaavva: (and everywhere else that still is like this)

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
vaavva: remove here as well, right now just duplicated below

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1562 @@
...
+    ///
+    /// # Errors
+    /// [`struct@Error`] of kind [`InvalidClientId`](ErrorKind::InvalidClientId)
+    /// if the Client Id of the [`ManagedClient`] isn't valid as a topic token.
+    #[allow(clippy::needless_pass_by_value)]
+    pub fn new(
+        application_context: ApplicationContext,
+        client: C,
+        options: ClientOptions,
+    ) -> Result<Self, Error> {
```
vaavva, line 80: see if you can add a test for new() that can force this error

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1562 @@
...
+                    &command_options,
+                ),
+            ),
+            asset_update_notification_dispatcher,
+        })
+    }
+
+    /// Convenience function to get all observed device & inbound endpoint names to quickly unobserve all of them before cleaning up
+    #[must_use]
+    pub fn get_all_observed_device_endpoints(&self) -> Vec<(String, String)> {
```
vaavva, line 187: are there any unit tests that can be added for these fns without network connectivity?

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1562 @@
...
+    ///
+    /// Note: If this method is called, the [`Client`] should not be used again.
+    /// If the method returns an error, it may be called again to re-attempt unsubscribing.
+    ///
+    /// Returns Ok(()) on success, otherwise returns [`struct@Error`].
+    /// # Errors
+    /// [`struct@Error`] of kind [`ShutdownError`](ErrorKind::ShutdownError)
+    /// if any of the invoker unsubscribes fail or if the unsuback reason code doesn't indicate success.
+    /// This will be a vector of any shutdown errors, all invokers will attempt to be shutdown.
+    pub async fn shutdown(&self) -> Result<(), Error> {
```
vaavva, line 231: add test for this (should return all errors without network connectivity, but validates that it attempts all of them)

vaavva (commented):
This is looking good! Apply the doc comments around everywhere, I just didn't want to comment the same thing a bunch of times and then I'll come back for final review

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1559 @@
...
+    }
+
+    /// Hashes the device name and inbound endpoint name to create a single string.
+    fn hash_device_endpoint(device_name: &str, inbound_endpoint_name: &str) -> String {
+        // `~`` can't be in a topic token, so this will never collide with another device + inbound endpoint name combo
+        format!("{device_name}~{inbound_endpoint_name}")
+    }
+
+    /// Unhashes the device name and inbound endpoint name from a single string.
+    fn un_hash_device_endpoint(hashed_device_endpoint: &str) -> Option<(String, String)> {
```
cartertinney, line 715: "unhash" should probably be one word. It's not technically a word, but "un" by itself isn't either

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1559 @@
...
+        &self,
+        device_name: String,
+        inbound_endpoint_name: String,
+        timeout: Duration,
+    ) -> Result<DeviceUpdateObservation, Error> {
+        let receiver_id = Self::hash_device_endpoint(&device_name, &inbound_endpoint_name);
+        let rx = self
+            .device_update_notification_dispatcher
+            .register_receiver(receiver_id.clone())
+            .map_err(ErrorKind::from)?;
```
cartertinney, line 570: Design note, and not one we need to use here by any means, but you can do the error kind conversion nicely if there's a wrapper function. So, an internal function that returns `Result<X, ErrorKind>` can automatically convert using the `?` operator, and then the outer function that returns `Result<X, Error>` can also automatically convert using the `?` operator.

But mapping the error everywhere works too. I don't think this alone is enough reason to have separate internal functions, but if you have other reasons as well, it becomes very convenient.

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
cartertinney: Not network failures, it's due to literal client infrastructure errors (i.e. session dropped so no event loop to send MQTT requests to)

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
cartertinney: This could be valuable for testing, but since the internal AIOProtocolError is not pub, the end user can't match on the underlying kind. Perhaps we would like them to, but in that case the error definition needs to make the inner value of the enum pub.

I think. Maybe I'm wrong and the default to pub because it's a pub enum? Someone check - at least, if we want them to be accessing the inner AIOProtocolError

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1730 @@
...
+use crate::azure_device_registry::device_name_gen::adr_base_service::client as adr_name_gen;
+use crate::azure_device_registry::{
+    Asset, AssetStatus, AssetUpdateObservation, Device, DeviceStatus, DeviceUpdateObservation,
+    Error, ErrorKind,
+    device_name_gen::{
+        common_types::options::CommandInvokerOptionsBuilder,
+        common_types::options::TelemetryReceiverOptionsBuilder,
+    },
+};
+use crate::common::dispatcher::{DispatchError, Dispatcher};
```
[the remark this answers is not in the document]
vaavva: interesting, this might explain some dead code weirdness - does this need to be under the state_store feature flag? I would have assumed that since it's in common, it wouldn't need a feature flag at all

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1559 @@
...
+            .get_all_receiver_ids()
+        {
+            // best effort, if the id can't be parsed, then skip it
+            if let Some((device_name, inbound_endpoint_name)) =
+                Self::un_hash_device_endpoint(&device_receiver_id)
+            {
+                device_endpoints.push((device_name, inbound_endpoint_name));
+            }
+        }
+        device_endpoints
```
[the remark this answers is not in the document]
vaavva: you'll need an allocated String to pass in as the arguments for unobserve...(), so doesn't make a huge difference whether we allocate here or later imo

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1559 @@
...
+    }
+
+    /// Convenience function to get all observed asset names to quickly unobserve all of them before cleaning up
+    #[must_use]
+    pub fn get_all_observed_assets(&self) -> Vec<(String, String, String)> {
```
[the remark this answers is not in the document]
vaavva: as an optimization soon, I think we should probably switch to the model that the file mount adr operations have, which is using a deviceRef instead of the distinct fields, since you never really need them not together
```
pub struct DeviceEndpointRef {
    /// The name of the device
    pub device_name: String,
    /// The name of the inbound endpoint
    pub inbound_endpoint_name: String,
}
```

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
vaavva: we need this at minimum for testing just to make sure the unit test fails if it returns timeout instead of the error we're trying to validate for - otherwise the unit tests will always pass, since no initial error returned will result in the timeout aio error if there's no network connectivity.

There might be a different conversation here about how we're exposing aio protocol error for service clients - if they can't index into the enums, then we probably would want to expose it differently, since a validation vs network error should be handled differently

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1559 @@
...
+impl<C> Client<C>
+where
+    C: ManagedClient + Clone + Send + Sync + 'static,
+    C::PubReceiver: Send + Sync,
+{
+    // ~~~~~~~~~~~~~~~~~ General APIs ~~~~~~~~~~~~~~~~~~~~~
+    /// Create a new Azure Device Registry Client.
+    ///
+    /// # Errors
+    /// [`struct@Error`] of kind [`InvalidClientId`](ErrorKind::InvalidClientId)
```
vaavva, line 72: if you haven't run `cargo doc` yet and verified that the output looks good/everything links up properly, definitely do that before finalizing the PR!

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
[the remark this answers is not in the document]
cartertinney: submodules of common are featured out for optimization

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1529 @@
...
+                .build()
+                .map_err(ErrorKind::from)?;
+        let response = self
+            .notify_on_device_update_command_invoker
+            .invoke(unobserve_request)
+            .await
+            .map_err(ErrorKind::from)?;
+        match response.payload.notification_preference_response {
+            adr_name_gen::NotificationPreferenceResponse::Accepted => {
+                let receiver_id = Self::hash_device_endpoint(&device_name, &inbound_endpoint_name);
```
RLeclair, line 680: Maybe a note for the future but I think the dispatcher could have a generic as the key name as long as it implements the `Hash` trait. If you do that you won't need to implement the hash and unhash functions, you'd just use a struct in this case like so:

``` Rust
#[derive(Hash)]
struct device_endpoint {
  device_name: String, 
  inbound_endpoint_name: String,
}
```

RLeclair: Come to think of it, this also looks like `DeviceRef`, so we could potentially just have `DeviceRef` derive that trait

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1529 @@
...
+        Ok(response.payload.asset.into())
+    }
+
+    /// Updates the status of an asset in the Azure Device Registry service.
+    ///
+    /// # Arguments
+    /// * `device_name` - The name of the device.
+    /// * `inbound_endpoint_name` - The name of the inbound endpoint.
+    /// * `asset_name` - The name of the asset.
+    /// * [`AssetStatus`] - The status of an asset for the update.
```
RLeclair, line 767:
```suggestion
    /// * status - An [`AssetStatus`] containing the status of an asset for the update.
```

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
olivakar: changed into a separate error type as the request gets simply dropped by the service

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
olivakar: the connection settings cant be built without erroring out without client id, after that i need to create a mock client which can override the cleint id with emty string - havenot figured out any other way toc all new without client id other than this

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
olivakar: for checking that 6 shutdown were called i couldnt find a way to do it without mocking invokers , otherwise it is just returning Ok

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1559 @@
...
+        let mut asset_shutdown_attempt_count = 0;
+
+        let device_shutdown_notifier = Arc::new(Notify::new());
+        let asset_shutdown_notifier = Arc::new(Notify::new());
+
+        let mut device_receiver_closed = false;
+        let mut asset_receiver_closed = false;
+
+        loop {
+            tokio::select! {
```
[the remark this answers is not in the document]
olivakar: In my understanding adding a `biased` would mean:

Shutdown notifications would always be prioritized 
Device-related operations would be prioritized over asset-related operations

I am not sure if that's what we want. if you feel otherwise lets discuss.

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
vaavva: the invokers should be created on new(), so they should be there - are they returning Ok() because they never subscribed, so we don't call unsubscribe and the shutdown for them succeeds since it doesn't need a network op?

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1574 @@
...
+                    log::debug!(
+                        "Device, Endpoint and Asset combination not in observed list: {receiver_id:?}"
+                    );
+                }
+                Err(Error(ErrorKind::from(e)))
+            }
+        }
+    }
+
+    /// Stops observation of an asset's updates from the Azure Device Registry service.
```
vaavva, line 924:
```suggestion
    /// Stops observation of an [`Asset`]'s updates from the Azure Device Registry service.
```

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1574 @@
...
+    /// # Errors
+    /// [`struct@Error`] of kind [`InvalidRequestArgument`](ErrorKind::InvalidRequestArgument)
+    /// if timeout is 0 or > `u32::max`.
+    ///
+    /// [`struct@Error`] of kind [`AIOProtocolError`](ErrorKind::AIOProtocolError) if:
+    /// - device or inbound endpoint names are invalid.
+    /// - there are any underlying errors from the AIO RPC protocol.
+    ///
+    /// [`struct@Error`] of kind [`ObservationError`](ErrorKind::ObservationError)
+    /// if the observation was not successful.
```
vaavva, line 943:
```suggestion
    /// if the unobservation was not accepted by the service.
```

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
olivakar: yes most likely

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1608 @@
...
+    /// [`struct@Error`] of kind [`InvalidClientId`](ErrorKind::InvalidClientId)
+    /// if the Client Id of the [`ManagedClient`] isn't valid as a topic token.
+    #[allow(clippy::needless_pass_by_value)]
+    pub fn new(
+        application_context: ApplicationContext,
+        client: C,
+        options: ClientOptions,
+    ) -> Result<Self, Error> {
+        if !Self::is_valid_replacement(client.client_id()) {
+            return Err(ErrorKind::InvalidClientId(client.client_id().to_string()).into());
```
vaavva, line 81: This error looks like it's expecting an error message for the String, not just the field with the issue. Modify either here or in the error definition to include an error message

vaavva: Also not sure if the existing from impls for this errorKind still make sense now

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1608 @@
...
+    /// if the asset name is empty.
+    pub async fn get_asset(
+        &self,
+        device_name: String,
+        inbound_endpoint_name: String,
+        asset_name: String,
+        timeout: Duration,
+    ) -> Result<Asset, Error> {
+        if asset_name.trim().is_empty() {
+            return Err(ErrorKind::ValidationError("asset_name".to_string()).into());
```
vaavva, line 770:
```suggestion
            Err(Error(ErrorKind::ValidationError("asset_name".to_string()))
```
you don't need the return and ;, and then I think you should be able to do this to remove the into()

rust/azure_iot_operations_services/src/common.rs
```diff
@@ -6,7 +6,8 @@
 // NOTE: submodules should be behind the feature flags of the clients that use them to ensure they
 // are only compiled when necessary.
 
-#[cfg(feature = "state_store")]
+#[cfg(any(feature = "state_store", feature = "azure_device_registry"))]
+#[allow(dead_code)]
```
vaavva, line 10: needed again? Fixing the features I think should have fixed this

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1608 @@
...
+        device_name: String,
+        inbound_endpoint_name: String,
+        asset_name: String,
+        status: AssetStatus,
+        timeout: Duration,
+    ) -> Result<Asset, Error> {
+        if asset_name.trim().is_empty() {
+            return Err(
+                ErrorKind::ValidationError("Asset name must be present".to_string()).into(),
+            );
```
vaavva, line 822:
```suggestion
            Err(Error(ErrorKind::ValidationError("asset_name".to_string()))
```

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1608 @@
...
+    pub async fn observe_asset_update_notifications(
+        &self,
+        device_name: String,
+        inbound_endpoint_name: String,
+        asset_name: String,
+        timeout: Duration,
+    ) -> Result<AssetUpdateObservation, Error> {
+        if asset_name.trim().is_empty() {
+            return Err(ErrorKind::ValidationError("asset_name".to_string()).into());
+        }
```
vaavva, line 885:
```suggestion
        if asset_name.trim().is_empty() {
            Err(Error(ErrorKind::ValidationError("asset_name".to_string()))
        }
```

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1608 @@
...
+        &self,
+        device_name: String,
+        inbound_endpoint_name: String,
+        asset_name: String,
+        timeout: Duration,
+    ) -> Result<(), Error> {
+        if asset_name.trim().is_empty() {
+            return Err(
+                ErrorKind::ValidationError("Asset name must be present".to_string()).into(),
+            );
```
vaavva, line 993:
```suggestion
           Err(Error(ErrorKind::ValidationError("asset_name".to_string()))
```

rust/azure_iot_operations_services/src/azure_device_registry.rs
```diff
@@ -47,7 +49,7 @@ pub enum ErrorKind {
     #[error(transparent)]
     InvalidRequestArgument(#[from] rpc_command::invoker::RequestBuilderError),
     /// Client Id used for the ADR Client was invalid.
-    #[error("{0}")]
+    #[error("{0} is invalid as the identifier for the client")]
```
vaavva, line 52:
```suggestion
    #[error("{0} is invalid as client id")]
```
just to keep consistent terminology

rust/azure_iot_operations_services/src/azure_device_registry.rs
```diff
@@ -61,6 +63,9 @@ pub enum ErrorKind {
     /// An error occurred while shutting down the Azure Device Registry Client.
     #[error("Shutdown error occurred with the following protocol errors: {0:?}")]
     ShutdownError(Vec<AIOProtocolError>),
+    /// An error occured while validating the inputs.
+    #[error("{0} must not be empty")]
+    ValidationError(String),
```
cartertinney, line 68: Is this new error type really necessary? Couldn't we make just a single kind that is for all validation issues?

rust/azure_iot_operations_services/src/azure_device_registry.rs
```diff
@@ -6,7 +6,7 @@
 use core::fmt::Debug;
 use std::collections::HashMap;
 
-use azure_iot_operations_mqtt::control_packet::QoS as rumqttc_qos;
+use azure_iot_operations_mqtt::control_packet::QoS as mqtt_qos;
```
cartertinney, line 9: Why alias QoS?

rust/azure_iot_operations_services/src/azure_device_registry.rs
```diff
@@ -6,7 +6,7 @@
...
 use azure_iot_operations_mqtt::interface::AckToken;
 use azure_iot_operations_protocol::{common::aio_protocol_error::AIOProtocolError, rpc_command};
 use thiserror::Error;
```
cartertinney, line 12: Make sure to keep our dependencies separate between std library, third party and azure with a line break between the sections so cargo fmt doesn't undo it.

Here and everywhere.

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1604 @@
...
+    ///
+    /// [`struct@Error`] of kind [`AIOProtocolError`](ErrorKind::AIOProtocolError) if:
+    /// - device or inbound endpoint names are invalid.
+    /// - there are any underlying errors from the AIO RPC protocol.
+    ///
+    /// [`struct@Error`] of kind [`ObservationError`](ErrorKind::ObservationError)
+    /// if the observation was not accepted by the service.
+    pub async fn observe_device_update_notifications(
+        &self,
+        device_name: String,
```
cartertinney, line 581: This would actually be better as a &str because it would reduce allocations by 1, but honestly at this point we probably need to do a full pass specifically on optimizing strings given how many revisions there have been.

rust/azure_iot_operations_services/src/azure_device_registry.rs
olivakar: This is the only one for validation issue right now, if we have other validation issues in the future we can always combine

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
olivakar: The exisitng impls are needed as by default we are getting `Result<CommandInvokerOptions, CommandInvokerOptionsBuilderError>` on build. To avoid that we have to either add a panic or an else block to handle that error. I have kept the panic for now.

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -71,22 +71,28 @@ where
...
         }
+
         let command_options = CommandInvokerOptionsBuilder::default()
             .topic_token_map(HashMap::from([(
                 "connectorClientId".to_string(),
                 client.client_id().to_string(),
             )]))
             .build()
-            .map_err(ErrorKind::from)?;
+            .expect("Builder cannot fail as there is no validation function");
```
vaavva, line 95: the builder can (technically) fail, and there is a validation function at some layer. However, it can't fail _here_ because everything is statically generated (from the DTDL) or validated before this point (with the client id check we have)
```suggestion
            .expect("Builder cannot fail as input is statically generated or validated before this point");
```

vaavva: sync'd offline, this specific builder doesn't actually validate anything, so the original message is correct

rust/azure_iot_operations_services/src/azure_device_registry.rs
olivakar: We need to alias one as there are 2

rust/azure_iot_operations_services/src/azure_device_registry.rs
olivakar: It is already separated

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
olivakar: NA

rust/azure_iot_operations_services/src/azure_device_registry.rs
olivakar: merged the invalid client id and validation one

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -773,7 +777,9 @@ where
         timeout: Duration,
     ) -> Result<Asset, Error> {
         if asset_name.trim().is_empty() {
-            return Err(Error(ErrorKind::ValidationError("asset_name".to_string())));
+            return Err(Error(ErrorKind::ValidationError(
+                "{asset_name} is empty".to_string(),
```
vaavva, line 781:
```suggestion
                "{asset_name} must not be empty".to_string(),
```
change for all, just to be clearer on how to fix the error

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -83,7 +83,11 @@ where
         // THIS IS A TEMPORARY FIX. WORKAROUND FOR THE FACT THAT CODEGEN PANICS ON INVALID CLIENT ID
         // INSTEAD OF RETURNING AN ERROR
         if !Self::is_valid_replacement(client.client_id()) {
-            return Err(ErrorKind::InvalidClientId(client.client_id().to_string()).into());
+            return Err(ErrorKind::ValidationError(format!(
+                "Client id {} is invalid",
+                client.client_id()
+            ))
+            .into());
```
vaavva, line 90:
```suggestion
            return Err(Error(ErrorKind::ValidationError(format!(
                "Client id {} is invalid",
                client.client_id()
            ))));
```

rust/azure_iot_operations_services/src/azure_device_registry.rs
```diff
@@ -61,6 +57,9 @@ pub enum ErrorKind {
     /// An error occurred while shutting down the Azure Device Registry Client.
     #[error("Shutdown error occurred with the following protocol errors: {0:?}")]
     ShutdownError(Vec<AIOProtocolError>),
+    /// An error occured while validating the inputs.
```
vaavva, line 60:
```suggestion
    /// An error occurred while validating the inputs.
```

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1,2 +1,1622 @@
...
+impl<C> Client<C>
+where
+    C: ManagedClient + Clone + Send + Sync + 'static,
+    C::PubReceiver: Send + Sync,
+{
+    // ~~~~~~~~~~~~~~~~~ General APIs ~~~~~~~~~~~~~~~~~~~~~
+    /// Create a new Azure Device Registry Client.
+    ///
+    /// # Errors
+    /// [`struct@Error`] of kind [`InvalidClientId`](ErrorKind::InvalidClientId)
```
vaavva, line 72: update

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
vaavva: oh wait, actually you'll want
>"asset_name must not be empty".to_string(),

because otherwise it will print `"" must not be empty` and not actually tell the issue

vaavva (approved):

PR #762, opened 2025-04-30
[Rust][MQTT] Add the Debug and Getters traits to MqttConnectionSettings

clecompt-msft:


rust/azure_iot_operations_mqtt/src/connection_settings.rs
```diff
@@ -393,6 +393,13 @@ mod tests {
             .hostname("test_host".to_string())
             .build();
         assert!(connection_settings_builder_result.is_ok());
+
+        let connection_settings = connection_settings_builder_result.unwrap();
+        assert_eq!(connection_settings.client_id().as_str(), "test_client_id");
```
cartertinney, line 398: I think this is an unnecessary addition that distracts from the actual test. This test is about the builder validation, not the actual settings object, so the getters are immaterial to the unit under test.

If you want to have a getter test I would prefer it be separate (although I think it's rather unnecessary due to the derivation) or consistently applied to all tests (since we would want to show the getters work for all values, not just the minimum required values by the builder).

rust/azure_iot_operations_mqtt/src/lib.rs
```diff
@@ -28,6 +28,8 @@ pub mod interface_mocks;
 
 #[macro_use]
 extern crate derive_builder;
+#[macro_use]
+extern crate derive_getters;
```
cartertinney, line 32: Was this actually necessary to add? I didn't see anything in the docs that required this declaration, so I'm assuming you did this because we did it for `derive_builder` which may not apply here. Although if this is required, this is fine.

rust/azure_iot_operations_mqtt/src/connection_settings.rs
clecompt-msft: If you don't think the explicit test is necessary, I'm more inclined to just remove it and rely on the macro.

rust/azure_iot_operations_mqtt/src/lib.rs
clecompt-msft: Discussed offline; for posterity: added for consistency with `derive_builder`, and it adds the derive macro into the global namespace so it doesn't have to be `use`d repeatedly if we want to use it more in the future.

vaavva: This explains why I thought I was going crazy about sometimes needing to do `use` for derive_builder and sometimes not needing to 😅

vaavva (approved):

cartertinney (approved):

PR #763, opened 2025-05-01
Fix AVRO codegen when type is repeated

jrdouceur:
The ProtocolCompiler failed to deduplicate embedded type definitions when generating AVSC files, which could result in invalid schema definitions.  This change fixes the problem by ensuring that each definition is included in the AVSC file only once.


timtay-microsoft (approved):

jrdouceur:
Hi, @timtay-microsoft.  Thanks for the review/approval. Any idea what [this CI-Markdown failure](https://github.com/Azure/iot-operations-sdks/actions/runs/14784512117/job/41511896076?pr=763) could be? I can't make heads or tails of the error output, but it's failed 3 times in a row, so it does not seem to be transient. The checks were all passing before I merged from main.

timtay-microsoft:
I'm not too sure. It should be unrelated to your change (we're seeing this failure in other PRs as well), so feel free to merge

jrdouceur:
Merging despite CI-links failure, since it appears unrelated to this PR.

PR #765, opened 2025-05-01
[Rust] [Connector] Connector Stubs + BaseConnector new()

vaavva:
This PR adds stubs for the different Connector components needed by the protocol transformer.

Sections that are ready for full review:
- BaseConnector::new() (base_connector.rs) - with the caveat that more will be added here as the implementation is fleshed out
- connector_config changes
- lib.rs changes

Sections that will have implementation added in future PRs, and therefore probably have major changes and should be reviewed more closely then (but skimming for anything crazy isn't a bad idea at this point!):
- data_transformer.rs
- destination_endpoint.rs (missing needed functionality for base connector implementation as well)
- managed_azure_device_registry (implementation will likely change many of these structs, but this starts off the organization pattern)


rust/azure_iot_operations_connector/src/base_connector.rs
```diff
@@ -2,3 +2,118 @@
...
+            })
+        })
+    }
+
+    /// Runs the MQTT Session that allows all Connector Operations to be performed
+    /// Returns if the session ends. If this happens, the base connector will need to be recreated
+    ///
+    /// # Errors
+    /// Returns a [`SessionError`] if the session encounters a fatal error and ends.
+    pub async fn run(self) -> Result<(), SessionError> {
```
clecompt-msft, line 94: Was the eventual decision to keep the new/run split? Or was that elsewhere?

rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
```diff
@@ -0,0 +1,242 @@
...
+#[allow(dead_code)]
+impl<T> ManagedDataset<T>
+where
+    T: DataTransformer,
+{
+    pub(crate) fn new(dataset_definition: AssetDataset) -> Self {
+        // Create a new dataset
+        let forwarder = Forwarder::new(dataset_definition.clone());
+        let reporter = Arc::new(Reporter::new(dataset_definition.clone()));
+        let data_transformer = T::new(dataset_definition.clone(), forwarder, reporter.clone());
```
clecompt-msft, line 177: Worth noting that this now has a transformer per dataset. I think I prefer that design decision, for a few reasons, but it's different than what we discussed before.

rust/azure_iot_operations_connector/src/data_transformer.rs
```diff
@@ -0,0 +1,51 @@
...
+
+use std::sync::Arc;
+
+use azure_iot_operations_mqtt::interface::CompletionToken;
+use azure_iot_operations_services::azure_device_registry::{AssetDataset, AssetDatasetDataPoint};
+
+use crate::destination_endpoint::Forwarder;
+use crate::{Data, base_connector::managed_azure_device_registry::Reporter};
+
+pub trait DataTransformer {
```
clecompt-msft, line 16: Possibly outside the scope of this PR, but is there any way to share data between instances of a data-transformer with the current setup? If we have multiple WASM transformers, it'd probably benefit us to share the engine at least to reduce overhead.

rust/azure_iot_operations_connector/src/lib.rs
```diff
@@ -4,7 +4,24 @@
...
+pub mod data_transformer;
 pub mod destination_endpoint;
 pub mod filemount;
 pub mod source_endpoint;
+
+/// Message Schema to send to the Schema Registry Service
+pub type MessageSchema = PutRequest;
+
+/// Struct format for data sent to the [`DataTransformer`] and the destination
+pub struct Data {
```
clecompt-msft, line 20: Do we want to go ahead and add a HLC now, or wait until someone asks for it? Though I think there's a good chance that someone will be me. :)

rust/azure_iot_operations_connector/src/data_transformer.rs
cartertinney: If it's record-based data, Arc would allow is to share information. For something more complex, not sure, would need to know more about what you'd like to see shared.

rust/azure_iot_operations_connector/src/base_connector.rs
vaavva: because this is now going to be something that is used by instead of owning the protocol translator, then new and run have to be separate so you can call the get_managed_... (which is used to do the actual fn calls) in between the two calls (run will be consuming)

rust/azure_iot_operations_connector/src/data_transformer.rs
vaavva: kinda addressed in a response to your other comment, I'll add something for this PR to address this

rust/azure_iot_operations_connector/src/lib.rs
vaavva: good catch - I'll add it now, but with a comment that we might remove it. Easier to unplug things later than to try and fit it in later

rust/azure_iot_operations_connector/src/data_transformer.rs
clecompt-msft: In this case, it's the entire WASM engine. So a bit more complex than record-based data. It's designed to be thread-safe (or at least [the underlying engine is](https://docs.rs/wasmtime/32.0.0/wasmtime/struct.Engine.html); the TK wrapper is arc-mutex-ed, so there might be some contention there). It's effectively designed to be a per-process singleton.

RLeclair (approved):

rust/azure_iot_operations_connector/src/base_connector.rs
```diff
@@ -2,3 +2,118 @@
...
+        // Run the Session and Connector Operations
+        // TODO: make this a part of operation_with_retries to restart the connector if anything fails?
+        self.session.run().await
+    }
+
+    /// Creates a new `ManagedDeviceCreateObservation` to allow for all Azure Device Registry operations
```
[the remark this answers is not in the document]
vaavva: I left it more vague because this also allows for sending status updates, etc, not just receiving notifications

rust/azure_iot_operations_connector/src/base_connector.rs
```diff
@@ -2,3 +2,118 @@
...
+    pub fn create_managed_device_create_observation(&self) -> ManagedDeviceCreateObservation {
+        ManagedDeviceCreateObservation::new(self.connector_context.clone())
+    }
+}
+
+/// Helper function to perform any operation with retries.
+fn operation_with_retries<T, E: std::fmt::Debug>(operation: impl Fn() -> Result<T, E>) -> T {
```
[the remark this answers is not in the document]
vaavva: ooh I like it! I'll keep it this way for now in case we don't end up using this as much as I think we will, but if I do end up using it more, I'll take the macro

rust/azure_iot_operations_connector/src/lib.rs
vaavva: added - figured this would need to be Option, but it might not end up needing to be that way

rust/azure_iot_operations_connector/src/data_transformer.rs
vaavva: updated to having DataTransformer passed into connector::new() that is then able to create the DatasetDataTransformers. Added a lot of generics, but we were going to need those to be plumbed through anyways in the next PR

clecompt-msft (approved):

rust/azure_iot_operations_connector/src/base_connector.rs
```diff
@@ -2,3 +2,149 @@
...
+            // Create clients
+            // Create Azure Device Registry Client
+            let azure_device_registry_client = azure_device_registry::Client::new(
+                application_context.clone(),
+                session.create_managed_client(),
+                azure_device_registry::ClientOptions::default(),
+            )
+            .map_err(|e| e.to_string())?;
+
+            Ok((connector_config, azure_device_registry_client, session))
```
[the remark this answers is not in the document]
vaavva: retries exponentially forever - these operations should only fail if there's a problem with the config, so we're essentially retrying until there's new config values. Once we have update monitoring for the config, then I imagine this will get a bit smarter to only retry when there's something new

rust/azure_iot_operations_connector/src/base_connector.rs
```diff
@@ -2,3 +2,149 @@
...
+    pub async fn run(self) -> Result<(), SessionError> {
+        // Run the Session and Connector Operations
+        // TODO: make this a part of operation_with_retries to restart the connector if anything fails?
+        self.session.run().await
+    }
+
+    /// Creates a new `ManagedDeviceCreationObservation` to allow for all Azure Device Registry operations
+    /// TODO: restrict to only be able to call once for now
+    pub fn create_managed_device_create_observation(&self) -> ManagedDeviceCreationObservation<T> {
+        ManagedDeviceCreationObservation::new(self.connector_context.clone())
```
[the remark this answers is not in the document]
vaavva: connector_context is under an Arc<> already! Now, if you're talking about Arc'ing the ManagedDeviceCreationObservation instead, this is an idea I've been toying with

rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
```diff
@@ -0,0 +1,271 @@
...
+
+    /// Receives a notification for a newly created device endpoint. This notification includes
+    /// the [`ManagedDeviceEndpoint`], a [`ManagedDeviceEndpointUpdateObservation`] to observe for updates on
+    /// the new Device, and a [`ManagedAssetCreateObservation`] to observe for newly created
+    /// Assets related to this Device
+    #[allow(clippy::unused_async)]
+    pub async fn recv_notification(
+        &self,
+    ) -> Option<(
+        ManagedDeviceEndpoint<T>,
```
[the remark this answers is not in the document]
vaavva: there are many Assets that are linked to one DeviceEndpoint, so it doesn't really make sense to give you the first recv of the `ManagedAssetCreateObservation` up front.

Another clarification is that `ManagedDeviceEndpointUpdateObservation` doesn't allow you to receive additional `ManagedDeviceEndpoint`s, but it just gives you updates for this specific `ManagedDeviceEndpoint` that was returned here. (debating in the future maybe having this observation be a part of the `ManagedDeviceEndpoint` instead of separate, but that's another discussion)

rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
```diff
@@ -0,0 +1,271 @@
...
+/// Azure Device Registry Asset that includes additional functionality
+/// to report status, translate data, and send data to the destination
+#[allow(dead_code)]
+pub struct ManagedAsset<T: DataTransformer> {
+    // re-export of adr::Asset, but Dataset/Event/etc structs are of type ConnectorDataset/etc
+    /// Asset Definition
+    pub asset_definition: Asset,
+    data_transformer: Arc<T>,
+    /// datasets associated with the asset. Will be part of [`ManagedAsset`] struct in future, but for now this creates the right dependencies
+    pub datasets: Vec<ManagedDataset<T>>,
```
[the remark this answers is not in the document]
vaavva: yeah, probably going to want to do a bunch of the auto getter macro in this crate. Will add in the implementation PR

rust/azure_iot_operations_connector/src/base_connector.rs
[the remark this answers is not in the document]
vaavva: (actually, not sure if I checked this change into this branch, let me double check)

rust/azure_iot_operations_connector/src/base_connector.rs
```diff
@@ -2,3 +2,149 @@
...
+    /// Returns a [`SessionError`] if the session encounters a fatal error and ends.
+    pub async fn run(self) -> Result<(), SessionError> {
+        // Run the Session and Connector Operations
+        // TODO: make this a part of operation_with_retries to restart the connector if anything fails?
+        self.session.run().await
+    }
+
+    /// Creates a new `ManagedDeviceCreationObservation` to allow for all Azure Device Registry operations
+    /// TODO: restrict to only be able to call once for now
+    pub fn create_managed_device_create_observation(&self) -> ManagedDeviceCreationObservation<T> {
```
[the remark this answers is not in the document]
vaavva: For now, I've removed this TODO as I don't think this has as strong of a requirement to be a singleton as I was thinking before (although it might be a little weird to get two). But as we learn more from the implementation, I might take this change

rust/azure_iot_operations_connector/src/base_connector.rs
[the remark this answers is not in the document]
vaavva: pushed the change to reflect my earlier comment

rust/azure_iot_operations_connector/src/base_connector.rs
```diff
@@ -2,3 +2,149 @@
...
+        ManagedDeviceCreationObservation::new(self.connector_context.clone())
+    }
+}
+
+/// Helper function to perform any operation with retries.
+fn operation_with_retries<T, E: std::fmt::Debug>(operation: impl Fn() -> Result<T, E>) -> T {
```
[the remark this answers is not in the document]
vaavva: I think this might make a lot of sense. I'm going to hold off on refactoring this function until I know my most common use case for it/how often I'm using it, but once I have a better idea of that, I'll look into using the retry crate instead

PR #771, opened 2025-05-02
[.NET][Akri] Get device credentials using relative paths from GetDevice call

timtay-microsoft:
Also update connector template yaml files to get up-to-date with newly required fields

Also update polling rest thermostat connector sample to use 2 assets so that we can more easily test multi-asset scenarios

### DIFF
--- a/dotnet/samples/Connectors/EventDrivenTcpThermostatConnector/KubernetesResources/connector-template.yaml
+++ b/dotnet/samples/Connectors/EventDrivenTcpThermostatConnector/KubernetesResources/connector-template.yaml
@@ -13,7 +13,7 @@ spec:
     managedConfigurationSettings:
       managedConfigurationType: imageConfiguration
       imageConfigurationSettings:
-        imageName: eventdriventcpthermostatconnector:latest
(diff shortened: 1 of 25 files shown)


dotnet/samples/Connectors/EventDrivenTcpThermostatConnector/KubernetesResources/connector-template.yaml
```diff
@@ -13,9 +13,11 @@ spec:
     managedConfigurationSettings:
       managedConfigurationType: imageConfiguration
       imageConfigurationSettings:
-        image: eventdriventcpthermostatconnector:latest
+        imageName: "eventdriventcpthermostatconnector"
         imagePullPolicy: Never # This sample image is built locally, so no need to pull
-        replicas: 1
+        tagDigestSettings:
+          tagDigestType: tag
+          tag: "latest"
```
timtay-microsoft, line 20: The shape of these yaml files changed from the operator side recently, so I'm updating them in our samples accordingly

maximsemenov80 (approved):

PR #775, opened 2025-05-05
Update connection-settings.md. Default broker listener port

auyidi1:
updated default broker listener port


timtay-microsoft (approved):

vaavva:
This change will need to be made within the SDKs' code as well, otherwise the doc won't match the behavior (Tim, if you approved because that work is already being tracked, then this can be ignored)

timtay-microsoft:
It's a small enough change that I'll just make it in a branch and add @auyidi1 as a co-author

timtay-microsoft:
Closing this since #777 was merged

PR #778, opened 2025-05-06
[Go] Update ADR generated code and add ADR client wrapper

lev-i:


go/services/connector/main.go
```diff
@@ -0,0 +1,620 @@
...
+}
+
+// UpdateDeviceEndpoints updates the device endpoints in the file mount map.
+func (f *FileMountMap) UpdateDeviceEndpoints(
+	devices map[string]DeviceEndpointRef,
+) {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+
+	for deviceKey := range f.deviceMap {
```
clecompt-msft, line 248: nit: Consider using the `for deviceKey, data := range f.deviceMap {` syntax. Should save you the `if data, ok` below.

go/services/connector/main.go
```diff
@@ -0,0 +1,620 @@
...
+			delete(f.deviceMap, deviceKey)
+		}
+	}
+
+	for deviceKey, deviceRef := range devices {
+		if _, exists := f.deviceMap[deviceKey]; !exists {
+			assetCreationTx := make(chan AssetNotification, 100)
+			assetCreationRx := make(chan AssetNotification, 100)
+
+			go func(tx, rx chan AssetNotification) {
```
clecompt-msft, line 266: nit: Why not just capture the channels from the scope? I know this syntax is valid, but I've rarely if ever seen it used.

go/services/connector/main.go
```diff
@@ -0,0 +1,620 @@
...
+	assetsToRemove := make([]string, 0)
+	for assetName := range deviceData.trackedAssets {
+		if _, exists := assets[assetName]; !exists {
+			assetsToRemove = append(assetsToRemove, assetName)
+		}
+	}
+
+	// Remove assets that no longer exist
+	for _, assetName := range assetsToRemove {
+		if deletionChan, exists := deviceData.trackedAssets[assetName]; exists {
```
clecompt-msft, line 326: nit: It's slightly more idiomatic to use "ok" rather than "exists", even if the latter is more explicit.

go/services/connector/main.go
```diff
@@ -0,0 +1,620 @@
...
+		return ctx.Err()
+	}
+}
+
+// DeviceEndpointCreateObservation represents an observation for device endpoint creation events.
+type DeviceEndpointCreateObservation struct {
+	mountPath    string
+	deviceChan   chan DeviceNotification
+	fileMountMap *FileMountMap
+	ctx          context.Context
```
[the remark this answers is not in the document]
clecompt-msft: I think you can avoid this by just passing the context into the `watchLoop`?

go/services/connector/main.go
```diff
@@ -0,0 +1,620 @@
...
+		select {
+		case <-o.ctx.Done():
+			return
+		case event, ok := <-o.watcher.Events:
+			if !ok {
+				return
+			}
+
+			// Handle creation or removal of the mount path itself.
+			if event.Name == o.mountPath && (event.Op&fsnotify.Create) != 0 {
```
clecompt-msft, line 443: Huh. TIL this was a bitfield. We might want to review the SAT logic to make sure it's still valid in that light.

That being said, it's probably cleaner to use `event.Has` instead of mucking with the bits directly.

go/services/connector/main.go
```diff
@@ -0,0 +1,620 @@
...
+		o.lastDevices = devices
+	}
+}
+
+// RecvNotification receives the next device endpoint notification.
+func (o *DeviceEndpointCreateObservation) RecvNotification(
+	ctx context.Context,
+) (*DeviceNotification, error) {
+	select {
+	case notification, ok := <-o.deviceChan:
```
clecompt-msft, line 499: Worth noting that this value is not broadcast - only a single listener will receive any given notification. That may be fine, but it's different than the way your ConnectionMonitor behaves, so I figured it was worth calling out.

clecompt-msft:
Can you add the new package (and the ADR one) to the services readme?

PR #780, opened 2025-05-06
[ADR] Missing dataSetConfiguration dtdl

RLeclair:
There was a missing datasetConfiguration field in the ADR dtdl. for a dataset object Added relevant folks working on ADR. 

This would affect: #761 and #778 I believe


lev-i (approved):
Thanks

PR #782, opened 2025-05-06
[.NET][Akri] Add basic integration tests using the sample connectors

timtay-microsoft:
Just check that the connector is emitting the expected telemetry to the MQTT broker/setting the expected key in DSS


dotnet/src/Azure.Iot.Operations.Connector/ConnectorWorker.cs
```diff
@@ -297,15 +304,7 @@ public async Task ForwardSampledDatasetAsync(Asset asset, AssetDatasetSchemaElem
 
                     string stateStoreKey = destination.Configuration.Key ?? throw new AssetConfigurationException("Cannot publish sampled dataset to state store as it has no configured key");
 
-                    ulong? ttl = destination.Configuration.Ttl;
-                    StateStoreSetRequestOptions options = new StateStoreSetRequestOptions();
```
timtay-microsoft, line 301: After reading some operator-side documentation, I noticed that this field is for configuring MQTT message expiry intervals, not state store key expiry times

maximsemenov80 (approved):

PR #787, opened 2025-05-12
Fix latent codegen lint issues from PR 553

jrdouceur:


jrdouceur:

PR #789, opened 2025-05-12
[Rust] [Connector] Get assets implementation

vaavva:
This PR:
- Adds implementation for getting Assets and Datasets and reporting their statuses and message schemas using the base connector
- updates sample to test this flow **NOTE: I still don't think this sample will stick around forever, it's just a way to test this as we develop**
- adds derive-getters for base connector structs to ensure that consuming logic can't modify these fields
- adds schema registry integration to the base connector to support the message schema reporting flow
- removes the data transformer and related generics from the base connector
- switches to using refs for all names (device, endpoint, asset, and dataset) to avoid having a bunch of name fields on every struct and to keep them symmetrical
- updates some data sharing strategies for device and asset fields

This PR doesn't (will come in future PRs):
- [p0] block forwarding data if the MessageSchema has not been added yet.
- have correct data sharing strategies for specification fields since updates aren't supported yet. These will look more like status fields in the future
- share the device status down through the dataset level, it should in the future for completion
- have similar create observation flows for datasets as there are for devices and assets. This may come in the future as we determine whether this is helpful or adds bloac
- have a macro for creating new struct definitions that are similar to ADR struct definitions
- have support for events, streams, and management groups in addition to datasets. This change will likely group these implementations under an enum
- track the MessageSchema and MessageSchemaReference (fully) within the dataset. This should be added in the future so it's easier to determine whether a new MessageSchema is the same as the current one or not
- have the DatasetRef, AssetRef, and DeviceEndpointRefs in the correct places - currently they are defined in different locations and should be consolidated to the adr client

### DIFF
--- a/rust/azure_iot_operations_connector/Cargo.toml
+++ b/rust/azure_iot_operations_connector/Cargo.toml
@@ -25,6 +25,7 @@ azure_iot_operations_mqtt = { git = "https://github.com/Azure/iot-operations-sdk
 # tokio.workspace = true
 tokio = { version = "1.41", features = ["full"] } # TODO: If I use the workspace version of tokio I run into errors with tokio in the tests, will investigate later
 tokio-util.workspace = true
+tokio-retry2 = "0.5.7"
 notify.workspace = true
 notify-debouncer-full.workspace = true
 # futures = "0.3"
--- a/rust/azure_iot_operations_connector/Cargo.toml
+++ b/rust/azure_iot_operations_connector/Cargo.toml
@@ -15,6 +15,7 @@ publish = true
 azure_iot_operations_protocol = { version = "0.9", path = "../azure_iot_operations_protocol", registry = "aio-sdks"  }
 azure_iot_operations_services = { version = "0.8", path = "../azure_iot_operations_services", registry = "aio-sdks", features = ["state_store", "schema_registry", "azure_device_registry"]  }
 azure_iot_operations_mqtt = { version = "0.9", path = "../azure_iot_operations_mqtt", registry = "aio-sdks"  }
+derive-getters = { version = "0.5.0", features = ["auto_copy_getters"] }
 log.workspace = true
 notify.workspace = true
 notify-debouncer-full.workspace = true
--- a/rust/azure_iot_operations_connector/Cargo.toml
+++ b/rust/azure_iot_operations_connector/Cargo.toml
@@ -12,28 +12,19 @@ readme = "README.md"
 publish = true
 
 [dependencies]
-# azure_iot_operations_protocol = { version = "0.9", path = "../azure_iot_operations_protocol", registry = "aio-sdks"  }
-# temporary so that the adr client uses the same version of the protocol crate as the connector does
-azure_iot_operations_protocol = { git = "https://github.com/Azure/iot-operations-sdks", branch = "feature/rust-akri" }
-# azure_iot_operations_services = { version = "0.8", path = "../azure_iot_operations_services", registry = "aio-sdks", features = ["state_store", "schema_registry", "azure_device_registry"]  }
-# temporary import from branch while ADR client gets code reviewed
-azure_iot_operations_services = { git = "https://github.com/Azure/iot-operations-sdks", branch = "feature/rust-akri", features = ["state_store", "schema_registry", "azure_device_registry"]  }
-# azure_iot_operations_mqtt = { version = "0.9", path = "../azure_iot_operations_mqtt", registry = "aio-sdks"  }
-# temporary so that the adr client uses the same version of the mqtt crate as the connector does
-azure_iot_operations_mqtt = { git = "https://github.com/Azure/iot-operations-sdks", branch = "feature/rust-akri" }
-# derive_builder.workspace = true
-# tokio.workspace = true
-tokio = { version = "1.41", features = ["full"] } # TODO: If I use the workspace version of tokio I run into errors with tokio in the tests, will investigate later
-tokio-util.workspace = true
-tokio-retry2 = "0.5.7"
+azure_iot_operations_protocol = { version = "0.9", path = "../azure_iot_operations_protocol", registry = "aio-sdks"  }
+azure_iot_operations_services = { version = "0.8", path = "../azure_iot_operations_services", registry = "aio-sdks", features = ["state_store", "schema_registry", "azure_device_registry"]  }
+azure_iot_operations_mqtt = { version = "0.9", path = "../azure_iot_operations_mqtt", registry = "aio-sdks"  }
+log.workspace = true
 notify.workspace = true
 notify-debouncer-full.workspace = true
-# futures = "0.3"
-# tokio-stream = "0.1"
-thiserror.workspace = true
-log.workspace = true
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
+thiserror.workspace = true
+# tokio.workspace = true
+tokio = { version = "1.41", features = ["full"] } # TODO: If I use the workspace version of tokio I run into errors with tokio in the tests, will investigate later
+tokio-retry2 = "0.5.7"
+tokio-util.workspace = true
 
 [dev-dependencies]
 env_logger.workspace = true
--- a/rust/azure_iot_operations_connector/Cargo.toml
+++ b/rust/azure_iot_operations_connector/Cargo.toml
@@ -15,7 +15,7 @@ publish = true
 azure_iot_operations_protocol = { version = "0.9", path = "../azure_iot_operations_protocol", registry = "aio-sdks"  }
 azure_iot_operations_services = { version = "0.8", path = "../azure_iot_operations_services", registry = "aio-sdks", features = ["state_store", "schema_registry", "azure_device_registry"]  }
 azure_iot_operations_mqtt = { version = "0.9", path = "../azure_iot_operations_mqtt", registry = "aio-sdks"  }
-derive-getters = { version = "0.5.0", features = ["auto_copy_getters"] }
+derive-getters.workspace = true
 log.workspace = true
 notify.workspace = true
 notify-debouncer-full.workspace = true
--- a/rust/azure_iot_operations_connector/examples/base_connector_sample.rs
+++ b/rust/azure_iot_operations_connector/examples/base_connector_sample.rs
@@ -1,0 +1,108 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! This example demonstrates how to use the base connector SDK to initialize
+//! a deployed Connector and get [`DeviceEndpointClients`] and [`AssetClients`].
+//!
+//! This sample simply logs the device information received - a real
+//! connector would then use these to connect to the device/inbound endpoints
+//! and start operations defined in the assets.
+//!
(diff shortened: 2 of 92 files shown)


rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
```diff
@@ -408,45 +398,142 @@ where
...
+                asset_client_update_observation,
+                asset_deletion_token,
+            ));
+        }
     }
 }
 
 /// An Observation for asset update events that uses
 /// multiple underlying clients to get full information for a
 /// [`ProtocolTranslator`] to use.
```
RLeclair, line 530:
```suggestion
/// [`ProtocolTranslator`] to use.
```

Unrelated to this PR but since we might not have a trait, would this generate incorrectly in the docs?

rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
```diff
@@ -213,39 +218,35 @@ where
...
+pub struct DeviceEndpointClient {
+    /// The names of the Device and Inbound Endpoint
+    device_endpoint_ref: DeviceEndpointRef,
     /// The 'specification' Field.
-    pub specification: DeviceSpecification,
+    specification: Arc<DeviceSpecification>,
     /// The 'status' Field.
-    pub status: Option<DeviceEndpointStatus>,
-    connector_context: Arc<ConnectorContext<T>>,
+    status: Option<DeviceEndpointStatus>,
```
RLeclair, line 228: Why is this different than the asset status? For reference: `status: Arc<RwLock<Option<AssetStatus>>>,`

rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
```diff
@@ -455,110 +542,354 @@ where
...
+    ) -> Result<MessageSchemaReference, schema_registry::Error> {
+        // TODO: save message schema provided with message schema uri so it can be compared
+        // send message schema to schema registry service
+        let message_schema_reference = Retry::spawn(
+            RETRY_STRATEGY,
+            async || -> Result<schema_registry::Schema, RetryError<schema_registry::Error>> {
+                self.connector_context
+                    .schema_registry_client
+                    .put(
+                        message_schema.clone(),
```
RLeclair, line 793: Does this have to be cloned? I see that it doesn't get used anywhere else in the function and it is not a reference

RLeclair (approved):
LGTM!

rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
vaavva: It needs to be there because of the retry :( A bit annoying, but because of the retry, all values need to be able to be used again, similar to if they're in a loop

rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
vaavva: Point in time - since the status is only used within the DeviceEndpoint (which isn't clone), it doesn't need to be as locked down (compared to the asset status, which can be updated from the asset or the dataset). 
That being said, I'll go ahead and update this now for symmetry and to enable supporting Clone.

rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
cartertinney: Is this going to be an issue if the schema gets updated somehow?

rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
```diff
@@ -730,6 +1068,108 @@ impl DeviceEndpointStatus {
     }
 }
 
+/// Represents the specification of an Asset in the Azure Device Registry service.
+#[derive(Debug)]
+pub struct AssetSpecification {
```
cartertinney, line 1073: Obviously fine to keep this all `pub` for now, but can any of these be invalid?

cartertinney (approved):

rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
```diff
@@ -213,39 +217,36 @@ where
...
-    pub device_name: String,
-    /// The 'endpointName' Field.
-    pub inbound_endpoint_name: String, // needed for easy status reporting?
+#[derive(Clone, Debug, Getters)]
+pub struct DeviceEndpointClient {
+    /// The names of the Device and Inbound Endpoint
+    device_endpoint_ref: DeviceEndpointRef,
     /// The 'specification' Field.
-    pub specification: DeviceSpecification,
+    specification: Arc<DeviceSpecification>,
```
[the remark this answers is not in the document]
vaavva: because we don't want them to be publicly mutable

rust/azure_iot_operations_connector/src/base_connector.rs
```diff
@@ -52,58 +46,68 @@ where
...
+                (
+                    ConnectorConfiguration,
+                    azure_device_registry::Client<SessionManagedClient>,
+                    schema_registry::Client<SessionManagedClient>,
+                    Session,
+                ),
+                String,
+            >(|| {
+                // Get Connector Configuration
+                let connector_config =
```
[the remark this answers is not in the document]
vaavva: this is (largely) from a previous PR, but there aren't any network operations as part of this, just setting up all of the clients. The retry is for config changes that might make this succeed

rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
```diff
@@ -3,22 +3,29 @@
...
 
 use crate::{
-    Data, MessageSchema,
+    Data, DatasetRef, MessageSchema,
     base_connector::ConnectorContext,
-    data_transformer::{DataTransformer, DatasetDataTransformer},
     destination_endpoint::Forwarder,
     filemount::azure_device_registry::{
-        AssetCreateObservation, AssetDeletionToken, DeviceEndpointCreateObservation,
+        AssetCreateObservation, AssetDeletionToken, AssetRef, DeviceEndpointCreateObservation,
```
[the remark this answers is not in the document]
vaavva: yeah, I think this probably needs a new name, but it'll have to be changed across adr client, file mount adr, and here - it's tracked as a follow-up item in the PR description already. For context, AssetRef contains the name of the asset, device, and inbound endpoint

rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
vaavva: this is where the schema gets updated, so it would be pretty weird to call it twice concurrently. There is some future work around when retries should stop retrying though.

rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
```diff
@@ -455,110 +549,354 @@ where
...
+    ///
+    /// # Panics
+    /// if the asset status mutex has been poisoned, which should not be possible
+    #[must_use]
+    pub fn message_schema_reference(&self) -> Option<MessageSchemaReference> {
+        // unwrap can't fail unless lock is poisoned
+        self.asset_status
+            .read()
+            .unwrap()
+            .as_ref()?
```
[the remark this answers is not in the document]
vaavva: The `as_ref()` is to get beyond the readWriteGuard - this was the simplest I could get this, but also would love if there was a better way

rust/azure_iot_operations_connector/Cargo.toml
```diff
@@ -15,6 +15,7 @@ publish = true
 azure_iot_operations_protocol = { version = "0.9", path = "../azure_iot_operations_protocol", registry = "aio-sdks"  }
 azure_iot_operations_services = { version = "0.8", path = "../azure_iot_operations_services", registry = "aio-sdks", features = ["state_store", "schema_registry", "azure_device_registry"]  }
 azure_iot_operations_mqtt = { version = "0.9", path = "../azure_iot_operations_mqtt", registry = "aio-sdks"  }
+derive-getters = { version = "0.5.0", features = ["auto_copy_getters"] }
```
clecompt-msft, line 18: If we're starting to use this in a few places, should we move it to the workspace?

rust/azure_iot_operations_connector/examples/base_connector_sample.rs
```diff
@@ -86,9 +86,95 @@ async fn run_program(
             }
         };
 
+        // Start handling the assets for this device endpoint
+        // if we didn't accept the inbound endpoint, then there's no reason to manage the assets
+        if endpoint_status.is_ok() {
+            tokio::task::spawn({
+                async move {
+                    run_assets(asset_creation_observation).await;
```
clecompt-msft, line 94: Since run_assets already returns a `Future`, does this still need to be nested in an `async` block?

rust/azure_iot_operations_connector/examples/base_connector_sample.rs
```diff
@@ -86,9 +86,95 @@ async fn run_program(
...
+async fn run_dataset(dataset_client: DatasetClient) {
+    log::info!("new Dataset: {dataset_client:?}");
+
+    // now we should update the status of the dataset and report the message schema
+    dataset_client.report_status(Ok(())).await;
+    match dataset_client
+        .report_message_schema(
+            schema_registry::PutRequestBuilder::default()
+                .content(
+                    r#"
```
clecompt-msft, line 152: Since the connector already uses `serde_json`, would the `json!` macro be useful here?

clecompt-msft (approved):

rust/azure_iot_operations_connector/examples/base_connector_sample.rs
vaavva: probably, but since this is very brute force just to test for now, I'll leave it as is (if this were a better example, this json would actually get generated from the dataset). Ideally, this sample either won't exist very soon, or we'll use the actual implementation to generate this message schema

rust/azure_iot_operations_connector/examples/base_connector_sample.rs
vaavva: good call, updated

rust/azure_iot_operations_connector/Cargo.toml
vaavva: moved!

rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
cartertinney: i.e. can these be edited to have values that we wouldn't want it to have?

vaavva: ah, I see - so this struct is under a getter on the AssetClient, so if someone modifies something on it, it isn't tied to the actual Client anymore, it's just their copy that they'd be modifying

PR #790, opened 2025-05-13
Fix interaction between NormalResult and shared namespace

jrdouceur:
One-line fix to Rust code generation when using both CLI --shared option and Result co-type


jrdouceur:

abhipsaMisra (approved):
👍🏽

PR #791, opened 2025-05-13
[Rust][Services] Add public fields to structs, and wrap generated enums

cartertinney:
`PutRequest` now has some public fields
`SchemaType` and `Format` are no longer exposing generated code, and also implement `Eq` and `PartialEq`

### DIFF
--- a/rust/azure_iot_operations_services/src/schema_registry.rs
+++ b/rust/azure_iot_operations_services/src/schema_registry.rs
@@ -91,10 +91,10 @@ pub struct ServiceError {
 #[builder(setter(into), build_fn(validate = "Self::validate"))]
 pub struct GetRequest {
     /// The unique identifier of the schema to retrieve. Required to locate the schema in the registry.
-    id: String,
+    pub id: String,
     /// The version of the schema to fetch. If not specified, defaults to "1".
     #[builder(default = "DEFAULT_SCHEMA_VERSION.to_string()")]
-    version: String,
+    pub version: String,
 }
 
 impl GetRequestBuilder {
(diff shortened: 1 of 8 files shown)


rust/azure_iot_operations_services/src/schema_registry.rs
```diff
@@ -91,10 +91,10 @@ pub struct ServiceError {
 #[builder(setter(into), build_fn(validate = "Self::validate"))]
 pub struct GetRequest {
     /// The unique identifier of the schema to retrieve. Required to locate the schema in the registry.
-    id: String,
+    pub id: String,
```
vaavva, line 94: since this one has validate, do we need to add this validation later on as well in case it gets changed?

cartertinney: Oh, I'll take this out then.

cartertinney: If we think that this is going to be something that scales out in the future, maybe getters would be preferable.

vaavva (approved):

rust/azure_iot_operations_services/src/schema_registry/schemaregistry_gen/schema_registry/format.rs
```diff
@@ -3,7 +3,7 @@
 use serde::{Deserialize, Serialize};
 
 /// Supported schema formats
-#[derive(Serialize, Deserialize, Debug, Clone)]
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
```
vaavva, line 6: can't change this in the generated files otherwise they'll get overwritten anytime we regenerate

cartertinney: oh damn I just realized this was generated code....

clecompt-msft (approved):

rust/azure_iot_operations_services/src/schema_registry.rs
```diff
@@ -86,15 +120,16 @@ pub struct ServiceError {
     pub property_value: Option<String>,
 }
 
+// TODO: How should id be exposed, if at all?
```
vaavva, line 123: Does the GetRequest even need to be modifyable in the same way the PutRequest does? There isn't the same semantic meaning to it being something you'd save I think

vaavva (approved):

rust/azure_iot_operations_services/src/schema_registry.rs
vaavva: Advantage to validating before the service is avoiding the network call. But whether we validate in build vs get() seems fairly similar to me if we need the mutability. Agree that getters/setters would also solve the same issue since set could include the validation

rust/azure_iot_operations_services/src/schema_registry.rs
cartertinney: I mean, it doesn't need to be. If you'd rather I hold off on this change, I will

PR #793, opened 2025-05-13
Sender

olivakar:


rust/azure_iot_operations_protocol/src/telemetry/sender.rs
```diff
@@ -189,6 +189,9 @@ pub struct Message<T: PayloadSerialize> {
     /// Cloud event of the telemetry message.
     #[builder(default = "None")]
     cloud_event: Option<CloudEvent>,
+    /// Indicates whether the message should be retained or not.A retained message is a normal MQTT message with the retained flag set to true.
```
RLeclair, line 192:
```suggestion
    /// Indicates whether the message should be retained or not. A retained message is a normal MQTT message with the retained flag set to true.
```

RLeclair (commented):
Have you tested the retain functionality against MQ?

- yes , that test i added eventually to this PR

rust/azure_iot_operations_protocol/src/telemetry/sender.rs
cartertinney, line 192: Do we want to mention MQTT here?

rust/azure_iot_operations_protocol/src/telemetry/sender.rs
```diff
@@ -189,6 +189,9 @@ pub struct Message<T: PayloadSerialize> {
...
+    #[builder(default = "false")]
+    retain: bool,
```
cartertinney, line 194: Is this only relevant to the `Sender`? Might there be any reason the `Receiver` (or rather, the application that's using it) could also care about this?

rust/azure_iot_operations_protocol/src/telemetry/sender.rs
```diff
@@ -732,4 +735,27 @@ mod tests {
...
+            })
+            .times(1);
+
+        let message_builder_result = MessageBuilder::default()
+            .payload(mock_telemetry_payload)
+            .unwrap()
+            .retain(true)
+            .build();
+
+        assert!(message_builder_result.unwrap().retain);
```
cartertinney, line 759: This isn't really testing anything. It's testing that the builder sets the `retain` value as far as I can tell?
That's what a builder does, and it's derived code at that.

I'm not sure a test here is needed really, and if there is one required, there almost certainly shouldn't be one exclusively for testing `retain` - there should be a broad test of the Message struct

olivakar: now testing defaults

rust/azure_iot_operations_protocol/src/telemetry/sender.rs
olivakar: There are some docs like message expiry which mentions MQTT while others don't. I have taken out the entire second line as the first line of doc is sufficient.

rust/azure_iot_operations_protocol/src/telemetry/sender.rs
olivakar: Didn't see any place in receiver that it might be needed

clecompt-msft: Might be worth changing the comment to something like "retained by the broker" to clarify that this affects MQ rather than the receiver.

clecompt-msft (approved):

PR #796, opened 2025-05-14
Add connector metadata schema

timtay-microsoft:
This includes two example connector metadata files

Also fix one of the "additional configuration" fields in .NET being a string when it should be Json


doc/akri_connector/minimal-example-connector-metadata.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "RestConnector",
+  "description": "Connector for polling a REST server for information",
+  "version": "1.0.0",
+  "image": {
+    "repository": "contoso-acr.io/RestConnector",
+    "tag": "1.0.0"
+  },
+  "architecture": "amd64",
```
RLeclair, line 9: This field doesn't appear to be required

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,225 @@
...
+          },
+          "SDK": {
+            "type": "object",
+            "description": "Details about the versions of the AIO SDKs used to write this connector (if they were used)",
+            "properties": {
+              "MqttPackageVersion": {
+                "type": "string",
+                "description": "The version of the azure iot operations MQTT package"
+              },
+              "ServicesPackageVersion": {
```
RLeclair, line 127: Do we want a protocol package version as well?

doc/akri_connector/example-connector-metadata.json
```diff
@@ -0,0 +1,137 @@
...
+  "helm": {
+    "repository": "contoso-acr.io/helm/RestConnector",
+    "tag": "1.0.0",
+    "digest": "sha256:abcd1234..."
+  },
+  "architecture": "amd64",
+  "sourceCode": {
+    "language": ".NET",
+    "languageVersion": ".NET 8.0",
+    "SDK": {
```
varunpuranik, line 21: Even this - how is this expected to be used by the Akri Connector consumer?

doc/akri_connector/example-connector-metadata.json
```diff
@@ -0,0 +1,137 @@
...
+    "tag": "1.0.0",
+    "digest": "sha256:abcd1234..."
+  },
+  "helm": {
+    "repository": "contoso-acr.io/helm/RestConnector",
+    "tag": "1.0.0",
+    "digest": "sha256:abcd1234..."
+  },
+  "architecture": "amd64",
+  "sourceCode": {
```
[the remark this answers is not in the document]
timtay-microsoft: This isn't required since not every connector needs to use our SDKs

This is moreso for the connector writer to use. If we get an ICM complaining about a specific connector image, this makes it easier for us to debug it.

I figured it would be a low-cost thing to add to the schema, but I can remove this if you feel otherwise.

doc/akri_connector/example-connector-metadata.json
```diff
@@ -0,0 +1,137 @@
...
+    "minVersion": "1.0.0",
+    "maxVersion": "2.0.0"
+  },
+  "protocols": [
+    {
+      "name": "Rest",
+      "version": "1.0",
+      "configSchemas": {
+        "deviceEndpointAdditionalConfigSchema": {
+          "$schema": "http://json-schema.org/draft-07/schema#",
```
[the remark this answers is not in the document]
timtay-microsoft: I like offering the option to do either inline schema or load schema from file. I pushed changes to make that possible now

dotnet/test/Azure.Iot.Operations.Services.UnitTests/AssetAndDeviceRegistry/ModelsConverterTests.cs
```diff
@@ -782,7 +788,9 @@ public void Device_ToModel_ShouldConvertAllProperties()
             Assert.Equal("1.0", mqttEndpoint.Version);
             Assert.Equal("mqtt", mqttEndpoint.EndpointType);
             Assert.NotNull(mqttEndpoint.AdditionalConfiguration);
-            Assert.Equal("mqtt-config-value", mqttEndpoint.AdditionalConfiguration);
+            string actualJson = JsonSerializer.Serialize(mqttEndpoint.AdditionalConfiguration);
+            string expectedJson = JsonSerializer.Serialize(JsonDocument.Parse("{\"someKey\":\"someValue\"}"));
```
olivakar, line 792: Also manually typing the same strings in both the arrangement and assertion sections increases the risk of typos and also might require uptae if values change

dotnet/test/Azure.Iot.Operations.Services.UnitTests/AssetAndDeviceRegistry/ModelsConverterTests.cs
```diff
@@ -782,7 +788,9 @@ public void Device_ToModel_ShouldConvertAllProperties()
             Assert.Equal("1.0", mqttEndpoint.Version);
             Assert.Equal("mqtt", mqttEndpoint.EndpointType);
             Assert.NotNull(mqttEndpoint.AdditionalConfiguration);
-            Assert.Equal("mqtt-config-value", mqttEndpoint.AdditionalConfiguration);
+            string actualJson = JsonSerializer.Serialize(mqttEndpoint.AdditionalConfiguration);
```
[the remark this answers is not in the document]
timtay-microsoft: I'd rather that happen in a separate PR since that's not what this one is about

dotnet/test/Azure.Iot.Operations.Services.UnitTests/AssetAndDeviceRegistry/ModelsConverterTests.cs
timtay-microsoft: I can refactor this, sure

doc/akri_connector/example-connector-metadata.json
varunpuranik, line 1: 1. we need a version for the connector metadata schema itself..
2. we also need a way to identify that this is a connector metadata file.

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,251 @@
...
+      "required": [
+        "repository",
+        "tag"
+      ]
+    },
+    "kubernetes": {
+      "type": "object",
+      "description": "The kubernetes-level metadata",
+      "properties": {
+        "kubernetesPermissions": {
```
jollyaakash, line 97: As brought up in meeting, this needs to be removed as Operator + Connectors framework in Akri doesn't support any way to provide Kube API permissions to Connector.

doc/akri_connector/example-connector-metadata.json
timtay-microsoft: One field can cover this.

The schema itself will now have a version:

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "aioConnectorMetadataSchemaVersion": "1.0.0",
   ....
```

And the connector metadata files can be identified as connector metadata files because they specify which version of the AIO connector metadata schema they adhere to

```json
{
  "aioConnectorMetadataSchemaVersion": "1.0.0",
  "name": "RestConnector",
  "description": "Connector for polling a REST server for information",
  "maintainer": "dev@example.com",
  "vendor": "Contoso",
  ....
```

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,230 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "aioConnectorMetadataSchemaVersion": "1.0.0",
+  "definitions": {
+    "AdditionalConfiguration": {
+      "type": "object",
+      "description": "The schema for this 'additional configuration' field. Must provide inline schema xor the file that contains the schema",
```
abhipsaMisra, line 8: It might be worth calling out that this property is settable only for the `managedConfigurationSettings` variant of `runtimeConfiguration` in the `ConnectorTemplate`.

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,230 @@
...
+            "schemaFile"
+          ]
+        }
+      ]
+    }
+  },
+  "properties": {
+    "aioConnectorMetadataVersion": {
+      "type": "string",
+      "description": "The version of AIO connector metadata schema that this connector metadata file adheres to."
```
abhipsaMisra, line 34: I'm confused between `aioConnectorMetadataSchemaVersion` and `aioConnectorMetadataVersion`. Is the metadata schema also versioned?

Is `aioConnectorMetadataSchemaVersion` the version of the metadata schema while `aioConnectorMetadataVersion` is the version of the metadata file itself?

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,230 @@
...
+    },
+    "maintainer": {
+      "type": "string",
+      "description": "A name and/or email of the maintainer of this connector."
+    },
+    "vendor": {
+      "type": "string",
+      "description": "A name and/or email of the vendor that distributes this connector."
+    },
+    "image": {
```
abhipsaMisra, line 56: `ConnectorTemplate` calls this `imageConfigurationSettings`.

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,230 @@
...
+      "type": "object",
+      "properties": {
+        "repository": {
```
abhipsaMisra, line 59: `ConnectorTemplate` splits this into `registrySettings.containerRegistrySettings.registry`/`registrySettings.registryEndpointRef` and `imageName`.

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,230 @@
...
+          "description": "The tag of the docker image for this connector"
+        },
+        "digest": {
+          "type": "string",
+          "description": "The digest of the docker image for this connector"
+        }
+      },
+      "required": [
+        "repository",
+        "tag"
```
abhipsaMisra, line 74: Would it make sense defaulting this to `latest`? (That's how the Akri operator interprets this if not provided with an explicit value).

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,230 @@
...
+      ]
+    },
+    "helm": {
```
abhipsaMisra, line 77: Similar comment as the `image` settings

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,230 @@
...
+        {
+          "type": "object",
+          "properties": {
+            "name": {
+              "type": "string"
+            },
+            "version": {
+              "type": "string"
+            },
+            "configSchemas": {
```
abhipsaMisra, line 190:
```
properties:
  configurationSchemaRefs:
    nullable: true
    properties:
      additionalConfigSchemaRef:
        nullable: true
        type: string
      defaultDatasetConfigSchemaRef:
        nullable: true
        type: string
      defaultEventsConfigSchemaRef:
        nullable: true
        type: string
      defaultProcessControlConfigSchemaRef:
        nullable: true
        type: string
      defaultStreamsConfigSchemaRef:
        nullable: true
        type: string
    type: object
```

doc/akri_connector/connector-metadata-schema.json
varunpuranik: let's avoid latest.. it is usually difficult to deal with.

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,230 @@
...
+        },
+        {
+          "required": [
+            "schemaFile"
+          ]
+        }
+      ]
+    }
+  },
+  "properties": {
```
abhipsaMisra, line 31: @varunpuranik Do we need a section here for generic connector secret alias list?

varunpuranik: Yes, we do.. this should include teh list of secrets and additional config that the connector needs

abhipsaMisra: For ref `ConnectorTemplate` property: https://msazure.visualstudio.com/One/_git/akri?path=/deployment/helm/examples/akri.yaml&version=GBmain&line=561&lineEnd=562&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,230 @@
...
+            },
+            "version": {
+              "type": "string"
+            },
+            "configSchemas": {
+              "type": "object",
+              "description": "The JSON schemas for the various 'additional configuration' fields. A JSON schema must be provided for each 'additional configuration' only if that field is used by this connector.",
+              "properties": {
+                "deviceEndpointAdditionalConfigSchema": {
+                  "$ref": "#/definitions/AdditionalConfiguration"
```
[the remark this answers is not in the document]
timtay-microsoft: This $ref points to the data structure you are highlighting, so yes! Each of these config types can either be specified inline or by providing the file

doc/akri_connector/connector-metadata-schema.json
timtay-microsoft: "aioConnectorMetadataVersion" is just a name that I missed renaming here. I'll rename it to "aioConnectorMetadataSchemaVersion"

Broadly speaking, we want to version the connector metadata schema itself, and we want to version the specific connectors' metadata files.

"aioConnectorMetadataSchemaVersion" is the version of the connector metadata schema. This PR includes the first version of that schema (1.0.0)

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "aioConnectorMetadataSchemaVersion": "1.0.0",
```

Then, any connector that wants to use that schema should specify it:

```json
{
  "aioConnectorMetadataSchemaVersion": "1.0.0", // The version of the connector metadata schema
  "name": "RestConnector",
  "version": "2.0.0", // The version of the connector
```

doc/akri_connector/connector-metadata-schema.json
timtay-microsoft: I'd also rather they specify the specific tag (even if it is just latest)

doc/akri_connector/connector-metadata-schema.json
timtay-microsoft: I can split this into repo name + image name, sure

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,230 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "aioConnectorMetadataSchemaVersion": "1.0.0",
+  "definitions": {
+    "AdditionalConfiguration": {
+      "type": "object",
+      "description": "The schema for this 'additional configuration' field. Must provide inline schema xor the file that contains the schema",
+      "properties": {
+        "schema": {
```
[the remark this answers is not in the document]
timtay-microsoft: Within the ```properties``` section, each of those stringified json fields just points to this definition to say that they are all either inline json schemas, or pointers to a file with a json schema

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,230 @@
...
+          "type": "string",
+          "description": "The minimal (inclusive) version that can run this connector."
+        },
+        "maxVersion": {
+          "type": "string",
+          "description": "The maximal (inclusive) version that can run this connector."
+        }
+      }
+    },
+    "protocols": {
```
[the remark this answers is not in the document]
timtay-microsoft: Good idea

doc/akri_connector/minimal-example-connector-metadata.json
```diff
@@ -0,0 +1,33 @@
...
+          "schema": {
+            "$schema": "http://json-schema.org/draft-07/schema#",
+            "title": "Dataset Config Schema",
+            "type": "object",
+            "properties": {
+              "SamplingInterval": {
+                "description": "How frequently to sample each dataset by default (in milliseconds)",
+                "type": "integer"
+              }
+            }
```
hansgschossmann, line 26: you want to define a default value and maybe min, max. Integer: what happens if the user puts in a negative number, what is the connector doing when 0 is entered?

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,305 @@
...
+        },
+        "aioMaxVersion": {
+          "type": "string",
+          "description": "The maximal (inclusive) version that can run this connector."
+        }
+      }
+    },
+    "inboundEndpoints": {
+      "type": "array",
+      "description": "The array of inbound endpoints that this connector connects with.",
```
hansgschossmann, line 234: >The array of inbound endpoints that this connector connects with. [](http://example.com/codeflow?start=22&length=65)

It should be more like: For a all Device.endpoint.inbound's of this endpointType the connector will be deployed/notified.

@varunpuranik since we do now have a device CR i think we do no longer need to have more endpointTypes listed for a connector. if we support a list, this will cause ambiguity. if there are other connectors handling the same endpoint type, which additional configuration do we use?

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,305 @@
...
+      "items": [
+        {
+          "type": "object",
+          "properties": {
+            "name": {
+              "type": "string"
+            },
```
hansgschossmann, line 241: I thought you want to version the endpointTypes? It is called endpointType in Device.endpoints.inbound and we should be use similar naming. 
Should we also add a comment that shows the recommended (required?) format like <companyname>.<protocol>

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,305 @@
...
+        "secretKey",
+        "secretRef"
+      ]
+    },
+    "architecture": {
+      "type": "string",
+      "description": "The architecture that this connector should run on"
+    },
+    "sourceCode": {
+      "type": "object",
```
hansgschossmann, line 168: just curious: Whst is the purpose of this metadata?

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,305 @@
...
+                        "streamConfigSchema": {
+                          "$ref": "#/definitions/additionalConfigurationType"
+                        },
+                        "managementGroupConfigSchema": {
+                          "$ref": "#/definitions/additionalConfigurationType"
+                        },
+                        "processControlConfigSchema": {
+                          "$ref": "#/definitions/additionalConfigurationType"
+                        }
+                      }
```
hansgschossmann, line 283: is this real?

doc/akri_connector/connector-metadata-schema.json
hansgschossmann: @timtay-microsoft could you pls resolve/close the comment. i joined the review recently and it is confusing if those comments are still active.

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,305 @@
...
+      "description": "The version of AIO connector metadata schema that this connector metadata file adheres to."
+    },
+    "name": {
+      "type": "string",
+      "description": "The name of the connector."
+    },
+    "description": {
+      "type": "string",
+      "description": "A brief (human-readable) description of the connector."
+    },
```
hansgschossmann, line 43: Could you pls also mention where this is shown? I would also mention that preference is to have it in english.

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,305 @@
...
+    "version": {
+      "type": "string",
+      "description": "The version of this connector."
```
hansgschossmann, line 46: It should be required that this is in semantic version format.

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,305 @@
...
+          "type": "string",
+          "description": "The container registry that holds the image for the connector. For example: contoso-acr.io"
+        },
+        "imageName": {
+          "type": "string",
+          "description": "The docker image name for the connector. For example: MyConnector"
+        },
+        "tag": {
+          "type": "string",
+          "description": "The tag of the docker image for this connector"
```
hansgschossmann, line 69: Why is this required and we expect that the image version is equal to version?

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,305 @@
...
+          "type": "string",
+          "description": "The container registry name that holds the helm chart for the connector. For example: contoso-acr.io"
+        },
+        "helmChartName": {
+          "type": "string",
+          "description": "The helm chart name for the connector. For example: helm/MyConnector"
+        },
+        "tag": {
+          "type": "string",
+          "description": "The tag of the helm chart for this connector"
```
hansgschossmann, line 106: same as in image

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,305 @@
...
+          "type": "string",
+          "description": "The docker image name for the connector. For example: MyConnector"
+        },
+        "tag": {
+          "type": "string",
+          "description": "The tag of the docker image for this connector"
+        },
+        "digest": {
+          "type": "string",
+          "description": "The digest of the docker image for this connector"
```
hansgschossmann, line 73: what is the use case for specifying a digest? why do we not make a semantic version formatted tag a requirement?

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,305 @@
...
+        "helmChartName": {
+          "type": "string",
+          "description": "The helm chart name for the connector. For example: helm/MyConnector"
+        },
+        "tag": {
+          "type": "string",
+          "description": "The tag of the helm chart for this connector"
+        },
+        "digest": {
+          "type": "string",
```
hansgschossmann, line 109: same as for image

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,305 @@
...
+        "language": {
+          "type": "string",
+          "description": "The programming language used to write this connector"
+        },
+        "languageVersion": {
+          "type": "string",
+          "description": "The version of the programming language used to write this connector"
+        },
+        "sdk": {
+          "type": "object",
```
hansgschossmann, line 180: should this be plural?

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,305 @@
...
+          "description": "Details about the versions of the AIO SDKs used to write this connector (if they were used)",
+          "properties": {
```
hansgschossmann, line 182: For what are these versions used?

If there is no use case, then we should not add them.

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,305 @@
...
+            "mqttPackageVersion": {
+              "type": "string",
+              "description": "The version of the azure iot operations MQTT package"
+            },
```
hansgschossmann, line 186: Azure IoT Operations

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,305 @@
...
+            "servicesPackageVersion": {
+              "type": "string",
+              "description": "The version of the azure iot operations services package"
+            },
+            "connectorPackageVersion": {
+              "type": "string",
```
hansgschossmann, line 192: What is meant by this and what is the use case that this should be specified?

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,305 @@
...
+        "language",
+        "languageVersion"
+      ]
+    },
+    "resourceRequirements": {
+      "type": "object",
+      "description": "The minimal resource requirements to run this connector",
+      "properties": {
+        "cpu": {
+          "type": "string"
```
hansgschossmann, line 208: i think we need to specify limits and requests

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,305 @@
...
+        },
+        "memory": {
+          "type": "string"
```
hansgschossmann, line 211: same as cpu

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,305 @@
...
+          }
+        }
+      },
+      "required": [
+        "language",
+        "languageVersion"
+      ]
+    },
+    "resourceRequirements": {
+      "type": "object",
```
hansgschossmann, line 204: Could you explain how this is interpreted for helm and statefulset configurations?

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,305 @@
...
+                {
+                  "type": "object",
+                  "properties": {
+                    "name": {
+                      "type": "string"
+                    },
+                    "configSchemas": {
+                      "type": "object",
+                      "description": "The JSON schemas for the asset-level various 'additional configuration' fields. A JSON schema must be provided for each 'additional configuration' only if that field is used by this connector.",
+                      "properties": {
```
hansgschossmann, line 264: the terminology "additionalConfiguration" is almost gone, only used in Device CR

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,305 @@
...
+                        "datasetConfigSchema": {
+                          "$ref": "#/definitions/additionalConfigurationType"
```
hansgschossmann, line 266: we should use use as fieldname: "<fieldname-from-asset-CRD>Schema"

doc/akri_connector/connector-metadata-schema.json
hansgschossmann, line 264: we should point out that the schemas are used also for the corresponding default field values

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,305 @@
...
+              "properties": {
+                "deviceEndpointAdditionalConfigSchema": {
+                  "$ref": "#/definitions/additionalConfigurationType"
+                }
+              }
+            },
+            "assets": {
+              "type": "array",
+              "description": "The array of assets within an inbound endpoint",
+              "items": [
```
hansgschossmann, line 254: what do you mean with this? there are no assets within an inbound endpoint.

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,305 @@
...
+              "type": "object",
+              "description": "The inbound endpoint level JSON schemas for the various 'additional configuration' fields. A JSON schema must be provided for each 'additional configuration' only if that field is used by this connector.",
+              "properties": {
+                "deviceEndpointAdditionalConfigSchema": {
+                  "$ref": "#/definitions/additionalConfigurationType"
+                }
+              }
+            },
+            "assets": {
+              "type": "array",
```
hansgschossmann, line 252: @varunpuranik should we also add destinations defaults for the asset?

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,305 @@
...
+      "description": "The kubernetes-level metadata",
+      "properties": {
+        "minimumKubernetesVersion": {
+          "type": "string",
+          "description": "Minimum required version of kubernetes needed to run this connector"
+        }
+      }
+    },
+    "secrets": {
+      "type": "object",
```
hansgschossmann, line 141: @abhipsaMisra @varunpuranik @timtay-microsoft 
1) this has to be an array of secrets
2) i am not sure how this will work end to end:
- the secrets field is at top level, so portal need to handle it
-  there must be a description for the user what the purpose of the secret is
- we should add an optional flag as property as well to ensure the portal allows to not add it
- is the assumption that the user does create a secret with name secretRef ahead of deployment?
- does portal creates the secret during import?
- the connector is using secretAlias and secretKey, so secretRef could even added by the user in the portal. no?

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,305 @@
...
+          "description": "The name of the secret to be mounted"
+        }
+      },
+      "required": [
+        "secretAlias",
+        "secretKey",
+        "secretRef"
+      ]
+    },
+    "architecture": {
```
hansgschossmann, line 163: similar as with secrets we also need trustsettings at Connector and Asset level!

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,305 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
```
hansgschossmann, line 3: it would be great to add size limitations, defaults, min, max to fields where it make sense

doc/akri_connector/connector-metadata-schema.json
timtay-microsoft: I thought so, based on the naming of this field: https://github.com/Azure/iot-operations-sdks/blob/47a20a5662b5f1579c7c1d1cdddb9d218211dbd2/eng/dtdl/device-name-based-operations.json#L1024

If this isn't another "stringified json" field though, I'd be happy to remove it from this schema

doc/akri_connector/connector-metadata-schema.json
timtay-microsoft: I'll rename this to "endpointType", sure. I don't believe we have a recommended or required format, but maybe @abhipsaMisra can comment on that.

doc/akri_connector/connector-metadata-schema.json
timtay-microsoft: Sure

doc/akri_connector/connector-metadata-schema.json
timtay-microsoft: https://github.com/Azure/iot-operations-sdks/pull/796#discussion_r2109837305

doc/akri_connector/connector-metadata-schema.json
abhipsaMisra: Let me get back to you on this.

doc/akri_connector/connector-metadata-schema.json
timtay-microsoft: Resolving this since this feedback is being discussed in the image thread

doc/akri_connector/connector-metadata-schema.json
timtay-microsoft: I re-worded this a bit

doc/akri_connector/minimal-example-connector-metadata.json
timtay-microsoft: This doc describes a fictional connector. I'd rather not get into that level of detail.

doc/akri_connector/connector-metadata-schema.json
timtay-microsoft: I reworked this a bit from another comment you left, but now the schema asks users to specify the requests + limits which can be interpreted as "per pod" when dealing with statefulset configurations

doc/akri_connector/connector-metadata-schema.json
timtay-microsoft: I made the secrets an array now and added two new properties alongside "secretAlias": "isOptional" and "secretDescription".

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,343 @@
...
+  "aioConnectorMetadataSchemaVersion": "1.0",
+  "definitions": {
+    "schema": {
+      "type": "object",
+      "description": "The schema for a stringified JSON schema fields such as 'additionalConfigurationSchema'. Must provide inline schema xor the file that contains the schema. These schemas may be used to define the default values of unspecified fields.",
```
abhipsaMisra, line 8: For the `schemaFile` option, is this a reference to a schema that is already in schema registry, or a file from a filesystem, etc.?

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,343 @@
...
+    "vendor": {
+      "type": "string",
+      "description": "A name and/or email of the vendor that distributes this connector."
+    },
+    "imageConfigurationSettings": {
+      "type": "object",
+      "properties": {
+        "containerRegistryName": {
+          "type": "string",
+          "description": "The container registry that holds the image for the connector. For example: 'contoso-acr.io'"
```
abhipsaMisra, line 65: Question - 
- what if the registry is a private registry. Is that a valid scenario for the connector metadata consumer flow?
- related, we also support providing this registry information via a `registryEndpointRef`: https://msazure.visualstudio.com/One/_git/akri?path=/deployment/helm/examples/akri.yaml&version=GBmain&line=397&lineEnd=398&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents
Is that something you need to provide an option for?

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,343 @@
...
+      "properties": {
+        "minimumKubernetesVersion": {
+          "type": "string",
+          "description": "Minimum required version of kubernetes needed to run this connector"
+        }
+      }
+    },
+    "secrets": {
+      "type": "array",
+      "description": "The array of secrets that will be mounted onto all connector instances.",
```
abhipsaMisra, line 146: nit: 
```suggestion
      "description": "The array of secrets that are required for the connector application to work. The secret will be mounted onto all connector instances deployed for this connector application",
```

Something along these lines...?

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,343 @@
...
+          },
+          "assets": {
+            "type": "array",
+            "description": "The array of assets that reference this inbound endpoint",
+            "items": [
+              {
+                "type": "object",
+                "properties": {
+                  "name": {
+                    "type": "string"
```
abhipsaMisra, line 297: What is this name for?

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,343 @@
...
+          "version",
+          "configSchemas"
+        ]
+      }
+    ]
+  },
+  "required": [
+    "name",
+    "version",
+    "imageConfigurationSettings",
```
abhipsaMisra, line 339: This would need to be an either-or with the `helmConfigurationSettings`

timtay-microsoft: Good catch

doc/akri_connector/connector-metadata-schema.json
timtay-microsoft: asset name. I'll make this more explicit

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,360 @@
...
+          }
+        }
+      }
+    },
+    "required": [
+      "language",
+      "languageVersion"
+    ]
+  },
+  "resourceRequirements": {
```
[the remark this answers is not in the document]
timtay-microsoft: Yeah, we expect this will be needed in the future

doc/akri_connector/connector-metadata-schema.json
abhipsaMisra: you're required to specify the asset names upfront? So we have an inbound endpoint type -> and then all assets that could be associated with that inbound endpoint type?

doc/akri_connector/connector-metadata-schema.json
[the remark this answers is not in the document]
varunpuranik: hmm, if not modeled in Connector TEmplate, can we remove these for now?
I am worried if the shape does not fit what the TEmplate will need in the future.

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,360 @@
...
+            }
+          },
+          "assets": {
+            "type": "array",
+            "description": "The array of assets that reference this inbound endpoint",
+            "items": [
+              {
+                "type": "object",
+                "properties": {
+                  "assetName": {
```
varunpuranik, line 299: What is this assetName supposed to be? This is not a dev time artifact, it is a deployment time artifact

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,360 @@
...
+    }
+  },
+  "inboundEndpoints": {
+    "type": "array",
+    "description": "The array of inbound endpoints in a device that this connector will be deployed for.",
+    "items": [
+      {
+        "type": "object",
+        "properties": {
+          "endpointType": {
```
varunpuranik, line 279: What about the version?

doc/akri_connector/connector-metadata-schema.json
timtay-microsoft: I intended it to be how we distinguish one asset from another. I didn't know that it was only assigned at deployment time, though. Is there a unique id per asset I could use here instead?

varunpuranik: I would expect all Assets to be "similar" from the connector standpoint. So this may not need to be an array. 
@hansgschossmann / @marcschier - any thoughts?

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,328 @@
...
+          "servicesPackageVersion": {
+            "type": "string",
+            "description": "The version of the Azure IoT Operations services package"
+          },
+          "connectorPackageVersion": {
+            "type": "string",
+            "description": "The version of the Azure IoT Operations connector package"
+          }
+        }
+      }
```
hansgschossmann, line 217: what is the purpose of this field?

we use also a azure_iot_operations_protocol SDK.

to be honest, if there is no good reason we should not keep this. do we need the sourceCode field at all?

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,328 @@
...
+            "items": [
+              {
+                "type": "object",
+                "properties": {
+                  "assetName": {
+                    "type": "string",
+                    "description": "The name of the asset"
+                  },
+                  "configSchemas": {
+                    "type": "object",
```
hansgschossmann, line 270: The Asset CRD (see PR here: ([https://msazure.visualstudio.com/One/_git/Azure-IoT-Platform-DeviceRegistry/pullrequest/11879236](https://msazure.visualstudio.com/One/_git/Azure-IoT-Platform-DeviceRegistry/pullrequest/11879236)) contains the following schema fields:

- datasetConfiguration
- datasetDataPointConfiguration
- eventConfiguration
- eventDataPointConfiguration
- streamConfiguration
- managementGroupConfiguration
- actionConfiguration

Just add Schema at the end of each and then we are in sync with Asset CRD.

cc: @varunpuranik

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,328 @@
...
+    "items": [
+      {
+        "type": "object",
+        "properties": {
+          "endpointType": {
+            "type": "string",
+            "description": "The type of this inbound endpoint."
+          },
+          "configSchemas": {
+            "type": "object",
```
hansgschossmann, line 250: should be no plural. it is even redundant and additionalConfigurationSchema could be promoted up one level.

hansgschossmann (changes requested):
:clock1:

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,328 @@
...
+        "imageConfigurationSettings"
+      ]
+    },
+    {
+      "required": [
+        "helmConfigurationSettings"
+      ]
+    }
+  ]
+}
```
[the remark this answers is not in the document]
hansgschossmann: Other metadata we should add are: 
- Does this connector supports ACSA integration
- Does this connector supports TLS

Those could be used in the portal to show the corresponding UX.

          "supportAcsaVolumes": {
            "type": "bool",
            "description": "This connector does support ACSA volumes."
          }

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,328 @@
...
+                  }
+                }
+              }
+            ]
+          },
+          "version": {
+            "type": "string",
+            "description": "The version of this inbound endpoint."
+          }
+        },
```
hansgschossmann, line 301: Current discussion goes towards removing this. We need to stay in sync with ConnectorTemplate.

doc/akri_connector/connector-metadata-schema.json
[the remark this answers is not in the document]
hansgschossmann: We should also add metadata to indicate if the connector is in preview or stable or do we have it.

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,305 @@
...
+          ]
+        },
+        {
+          "required": [
+            "containerRegistryName",
+            "imageName",
+            "digest"
+          ]
+        }
+      ]
```
hansgschossmann, line 97: Is there a specific reason why we also support digest config and not only tags? I would start simple with tags and if we see customer need add digest later.

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,305 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "aioConnectorMetadataSchemaVersion": "1.0.0",
+  "definitions": {
+    "additionalConfigurationType": {
+      "type": "object",
+      "description": "The schema for this 'additional configuration' field. Must provide inline schema xor the file that contains the schema",
```
hansgschossmann: >xor [](http://example.com/codeflow?start=103&length=3)

typo

timtay-microsoft: This is deliberate. A user can specify one or the other, but not both

doc/akri_connector/connector-metadata-schema.json
timtay-microsoft: The connector template supports specifying digest or tag, so the connector metadata must as well

doc/akri_connector/connector-metadata-schema.json
timtay-microsoft: Added the protocol field as well. You brought up this same question above. It's moreso for us who write the connectors so that we can easily debug any issue logged against the connector image.

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,328 @@
...
+            "properties": {
+              "additionalConfigurationSchema": {
+                "$ref": "#/definitions/schema"
+              }
+            }
+          },
+          "assets": {
+            "type": "array",
+            "description": "The array of assets that reference this inbound endpoint",
+            "items": [
```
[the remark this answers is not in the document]
timtay-microsoft: My understanding is that each asset can have distinct configuration schemas. From [this thread](https://github.com/Azure/iot-operations-sdks/pull/796#discussion_r2124217220), it sounds like that may not be the case? Can you confirm one way or the other?

doc/akri_connector/connector-metadata-schema.json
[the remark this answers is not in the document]
timtay-microsoft: > Since connectors do not support all possible destinations, we should add metadata which allows to configure which destinations are supported. For example media connector does not support BSS, then DOE could suppress showing it to the user and prevent a runtime error.
> 
> cc: @varunpuranik

Something like?
```json
"supportedDestinations" : {
  "type": "array",
  "items": {
    "type": "string"
  }
}
```
I'm fine with adding that.

> Other metadata we should add are:
> 
> * Does this connector supports ACSA integration
> * Does this connector supports TLS
> 
> Those could be used in the portal to show the corresponding UX.
> 
> ```
>       "supportAcsaVolumes": {
>         "type": "bool",
>         "description": "This connector does support ACSA volumes."
>       }
> ```

I added the proposed "supportAcsaVolumes" property at the top level.

I believe all connectors that use our SDKs support TLS. We could add that flag too if you expect a non-trivial number of connectors to be written without using our SDKs.



> We should also add metadata to indicate if the connector is in preview or stable or do we have it.

I'll add an "isPreview" flag at the top level, sure.

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,337 @@
...
+    "type": "array",
+    "description": "The array of inbound endpoints in a device that this connector will be deployed for.",
+    "items": [
+      {
+        "type": "object",
+        "properties": {
+          "endpointType": {
+            "type": "string",
+            "description": "The type of this inbound endpoint."
+          },
```
hansgschossmann, line 263: The description should be:
The type of this inbound endpoint including the version like "Microsoft.Media:1.2.0"

doc/akri_connector/connector-metadata-schema.json
hansgschossmann: I have another comment in which i have the same comment and gave an explanation what this fields are about.

@timtay-microsoft pls remove the assetName field as well as making assets an array. What you try to define is what is stored in ADR. This metadata file has a very different purpose.

doc/akri_connector/connector-metadata-schema.json
hansgschossmann: this comment was referring to processControlSchema, which is not part of ADR.

You have now removed it. Thanks.

doc/akri_connector/connector-metadata-schema.json
hansgschossmann: i understand your point about technical possiblities. from a user perspective showing a digest and resolving it to a version is a though process. since this information is shown to the user in the portal i think forcing the connector providers to tag their images is reasonable.

Pls follow up with @varunpuranik @rangv

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,337 @@
...
+        "description": "If this connector supports ACSA volumes.",
+        "default": "false"
+    },
+    "supportedDestinations" : {
+      "type": "array",
+      "description": "The array of destinations that this connector supports.",
+      "items": {
+        "type": "string"
+      }
+    },
```
hansgschossmann, line 69: Pls check the Asset CRD: Destinations should go under Assets and could be different for datasets, events, and streams.

doc/akri_connector/connector-metadata-schema.json
hansgschossmann: hmm. i did not get it. the text says "xor" this is what i meant by typo. it should be "or"

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,337 @@
...
+                        "$ref": "#/definitions/schema"
+                      },
+                      "eventDataPointConfigSchema": {
+                        "$ref": "#/definitions/schema"
+                      },
+                      "streamConfigSchema": {
+                        "$ref": "#/definitions/schema"
+                      },
+                      "managementGroupConfigSchema": {
+                        "$ref": "#/definitions/schema"
```
hansgschossmann, line 295: Config -> Configuration to be in sync with naming in ADR CRD

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,337 @@
...
+        "type": "string",
+        "description": "The maximal (inclusive) version that can run this connector."
+      }
+    }
+  },
+  "additionalConfigurationSchema": {
+    "description": "The connector-level JSON schemas for the 'additional configuration' field. A JSON schema must be only if that field is used by this connector.",
+    "$ref": "#/definitions/schema"
+  },
+  "inboundEndpoints": {
```
hansgschossmann, line 253: tim, maybe my earlier comment was confusing.

the structure here should be:

inboundEndpoint
    endpointType
    version
    additionalConfiguration
    asset
        datasetConfigurationSchema
         .....
         actionConfigurationSchema
         supportedDatasetsDestinations
         defaultDatasetsDestination
         supportedEventsDestinations
         defaultEventDestination
         supportedStreamsDestinations
         defaultStreamDestination
         
@varunpuranik

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,337 @@
...
+    "description": "The array of inbound endpoints in a device that this connector will be deployed for.",
+    "items": [
+      {
+        "type": "object",
+        "properties": {
+          "endpointType": {
+            "type": "string",
+            "description": "The type of this inbound endpoint."
+          },
+          "assets": {
```
hansgschossmann, line 264: tim, maybe my earlier comment was confusing.

the structure here should be:

inboundEndpoint
    endpointType
    version
    additionalConfiguration
    asset
        datasetConfigurationSchema
         .....
         actionConfigurationSchema
         supportedDatasetsDestinations
         defaultDatasetsDestination
         supportedEventsDestinations
         defaultEventDestination
         supportedStreamsDestinations
         defaultStreamDestination

@varunpuranik

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,337 @@
...
+        {
+          "required": [
+            "schema"
+          ]
+        },
+        {
+          "required": [
+            "schemaFile"
+          ]
+        }
```
hansgschossmann, line 28: I would always go with inline schema, then we have one artifact to read whiich sits in the CR. Specifying a schema file could cause issues re: connectivity and authorization, which end up in IcM.

timtay-microsoft: @varunpuranik suggested allowing file refs elsewhere in this PR because he expects some schemas to be very large. I'd personally rather schemas always be inline as well, though. @varunpuranik  thoughts?

doc/akri_connector/connector-metadata-schema.json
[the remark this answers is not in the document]
timtay-microsoft: > re: TLS - this was meant if a connector does need and evaluate the trustsettings field in the connector_template. this is about connectinng to southbound- or northbound devices.
the SDK which you mentioned is most likely the mqtt TLS, which is configured via very differnt mechanism.

I see. Yes, I was talking about MQTT connection TLS support. I'll add a flag like "isEndpointTlsEnforced" per inboundEndpoint

doc/akri_connector/connector-metadata-schema.json
timtay-microsoft: > supportedDatasetsDestinations
> defaultDatasetsDestination
> supportedEventsDestinations
> defaultEventDestination
> supportedStreamsDestinations
> defaultStreamDestination

I added these as per-asset level properties.

I like changing "additionalConfiguration" to "connectorConfiguration" as well, but I'll hold off on changing it here unless the change is made at the CRD level as well.

doc/akri_connector/connector-metadata-schema.json
timtay-microsoft: I talked with Varun offline and he is fine with forcing inline here. I'll make the change

doc/akri_connector/connector-metadata-schema.json
timtay-microsoft: Talked with the Akri folks and we should be fine to rename "additionalConfiguration" to "connectorConfiguration" here, so I will.

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,346 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "aioConnectorMetadataSchemaVersion": "1.0",
```
varunpuranik, line 4: Is this needed? (line 4)?

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,346 @@
...
+      "description": "The version of this connector. This must use semantic versioning such as '1.2.3'",
+      "pattern": "^(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)",
+      "default": "1.0.0"
+    },
+    "isPreview": {
+      "type": "boolean",
+      "description": "True if this connector is considered a preview version.",
+      "default": "false"
+    },
+    "supportAcsaVolumes": {
```
varunpuranik, line 31: I am not sure if we should have this. What is the scenairo where we need this? 
We can have metadata about mount volumes and the description can say that it works with ACSA. But making this first class in this file seems odd to me.

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,346 @@
...
+          "connectorPackageVersion": {
+            "type": "string",
+            "description": "The version of the Azure IoT Operations connector package"
+          }
+        }
+      }
+    },
+    "required": [
+      "language",
+      "languageVersion"
```
varunpuranik, line 198: Why is language version required? 
Or even language? I would make all of these optional.

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,346 @@
...
+    "$ref": "http://json-schema.org/draft-07/schema#"
+  },
+  "inboundEndpoints": {
+    "type": "array",
+    "description": "The array of inbound endpoints in a device that this connector will be deployed for.",
+    "items": [
+      {
+        "type": "object",
+        "properties": {
+          "endpointType": {
```
varunpuranik, line 226: Should we model the version separately?

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,346 @@
...
+              },
+              "supportedEventDestinations": {
+                "type": "array",
+                "description": "The array of destinations that this asset's events supports.",
+                "items": {
+                  "type": "string",
+                  "enum": ["Mqtt", "BrokerStateStore", "Storage"]
+                }
+              },
+              "defaultEventDestinations": {
```
varunpuranik, line 269: Is this an array or a single default destination...

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,346 @@
...
+                    "$ref": "http://json-schema.org/draft-07/schema#"
+                  }
+                }
+              },
+              "supportedEventDestinations": {
+                "type": "array",
+                "description": "The array of destinations that this asset's events supports.",
+                "items": {
+                  "type": "string",
+                  "enum": ["Mqtt", "BrokerStateStore", "Storage"]
```
varunpuranik, line 266: StateStore is not allowed for Streams and Events, I believe

doc/akri_connector/connector-metadata-schema.json
varunpuranik: I am not aligned with this. I added a comment on the thread.

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,346 @@
...
+    "type": "object",
+    "description": "The AIO requirements to run this connector",
+    "properties": {
+      "aioMinVersion": {
+        "type": "string",
+        "description": "The minimal (inclusive) version that can run this connector."
+      },
+      "aioMaxVersion": {
+        "type": "string",
+        "description": "The maximal (inclusive) version that can run this connector."
```
timtay-microsoft, line 211: "If no known max version, 1.* is an acceptable value. Alternatively, not specifying an aioMaxVersion signals no upper bound"

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,346 @@
...
+      }
+    }
+  },
+  "connectorConfigurationSchema": {
```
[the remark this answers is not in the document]
timtay-microsoft: Sure, will model as key-value pairs instead

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,346 @@
...
+    "description": "The connector-level JSON schemas for the 'additional configuration' field. A JSON schema must be only if that field is used by this connector.",
+    "$ref": "http://json-schema.org/draft-07/schema#"
+  },
+  "inboundEndpoints": {
+    "type": "array",
```
timtay-microsoft: Don't need to be an array. We don't support multiple inboundEndpoints yet

doc/akri_connector/connector-metadata-schema.json
timtay-microsoft: Should be just one. I will make this not an array

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,346 @@
...
+                  },
+                  "managementGroupConfigurationSchema": {
+                    "$ref": "http://json-schema.org/draft-07/schema#"
+                  },
+                  "actionConfigurationSchema": {
+                    "$ref": "http://json-schema.org/draft-07/schema#"
+                  }
+                }
+              },
+              "supportedEventDestinations": {
```
timtay-microsoft: supported should be mandatory, but keep the default optional

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,321 @@
...
+          "type": "string",
+          "description": "The maximal (inclusive) version that can run this connector. If there is no known exact version, wildcards can be used such as '1.*.*'. Alternatively, not specifying an aioMaxVersion signals no upper bound"
+        }
+      }
+    },
+    "connectorConfigurationKeys": {
+      "type": "array",
+      "description": "The array of keys in the connector configuration",
+      "items": {
+        "type": "string"
```
hansgschossmann, line 198: just KVPs will most likely generate generate low fidelity user experience since the user does not know what the setting is about, if there are allowed value ranges, ...
technically as discussed yesterday: no problem with it, bc of timing maybe in the next version :)

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -0,0 +1,321 @@
...
+    },
+    "inboundEndpoints": {
+      "type": "array",
+      "description": "The array of inbound endpoints in a device that this connector will be deployed for.",
+      "items": [
+        {
+          "type": "object",
+          "properties": {
+            "endpointType": {
+              "type": "string",
```
hansgschossmann, line 209: we are not in sync with ADR here

timtay-microsoft: Looks like the latest ADR (still in PR) has:

```yaml
                        endpointType:
                          description: Type of connection endpoint. Formatted as ABC.XYZ:{connectorVersion}.
                          pattern: ^[a-zA-Z]+\.[a-zA-Z]+:[^+#{}/]+$
                          type: string
``` 

So I will add the pattern to this doc and copy the description (but keep the example 'Microsoft.Media:1.2.3')

PR #800, opened 2025-05-19
[Rust] [Connector] Cleanup

vaavva:
This PR:
- removes the now unused data transformer
- adds some more logging around retries
- adds escapes to the retries that aren't critical. Some of these maybe should be put back in later, but for now this will make our first drop more stable


rust/azure_iot_operations_connector/src/data_transformer.rs
```diff
@@ -1,51 +0,0 @@
-// Copyright (c) Microsoft Corporation.
```
cartertinney, line 1: My bad, I thought I had already removed this

cartertinney (approved):

RLeclair (approved):

rust/azure_iot_operations_connector/src/data_transformer.rs
vaavva: no worries!

PR #801, opened 2025-05-19
[Rust] [Connector] Fix status updates in base connector

vaavva:
# Problem
only vectors on statuses are applied as patch merges on the service, all other fields need to be specified as the current state to not be overwritten

# This PR
Updates status reporting APIs to send existing status fields when only trying to do a partial update.
Note that some of this logic may change in the future, depending on how the merging strategy evolves


rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
```diff
@@ -296,10 +296,23 @@ impl DeviceEndpointClient {
 
     /// Used to report the status of just the endpoint,
     /// and then updates the [`Device`] with the new status returned
+    /// # Panics
+    /// if the status mutex has been poisoned, which should not be possible
     pub async fn report_endpoint_status(&mut self, endpoint_status: Result<(), AdrConfigError>) {
-        // Create status with empty device status
+        // If the version of the current status config matches the current version, then include the existing config.
+        // If there's no current config or the version doesn't match, don't report a status since the status for this version hasn't been reported yet
+        let current_config = self.status.read().unwrap().as_ref().and_then(|status| {
```
cartertinney, line 304: This is probably more suited to a `match` isn't it? Or is there a subtlety I don't get that makes this preferable?

Here and the other similar additions.

cartertinney (approved):

rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
vaavva: the `and_then`s simplify out explicitly saying the `None` match arms should return `None`. Using matches I think it would look like this:
```
let current_config = match self.status.read().unwrap().as_ref() {
            Some(status) => {
                match &status.config {
                    Some(config) => {
                        if config.version == self.specification.version {
                            Some(config.clone())
                        } else {
                            None
                        }
                    },
                    None => None,
                }
            },
            None => None,
        };
```
But if that's preferable for Rust optimization, I can change it to this. Or if it was something else that you were suggesting, lmk

cartertinney: Ah, got it. You're right, I was missing something.

RLeclair (approved):

PR #806, opened 2025-05-20
[Rust] [Connector] Add Update support for DeviceEndpoints

vaavva:
This PR:
- adds update support for DeviceEndpointClients
- locks the device specification under rwlock now that it can be updated
- moves receiving updates onto the DeviceEndpointClient so that the specification and status can be updated in place
- shows update support in the sample

Notable limitation:
- Currently an update will be received when a status is reported. This behavior is in flux with the service, so I don't want to over index on de-duping this right now, but might in a follow-up PR if needed
- This PR doesn't add update support for Assets/Datasets - this will be in a follow-up PR so that any comments from this PR can be applied to that implementation as well


cartertinney (approved):

rust/azure_iot_operations_connector/examples/base_connector_sample.rs
```diff
@@ -56,86 +57,86 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+
+                for dataset in asset_client.datasets() {
+                    tokio::task::spawn({
+                        let dataset_clone = dataset.clone();
+                        async move {
+                            run_dataset(dataset_clone).await;
+                        }
+                    });
+                }
+                } else {
```
clecompt-msft, line 135: nit: same issue as the REST PR, formatting is getting wonky in the `select!`.

RLeclair (approved):

rust/azure_iot_operations_connector/examples/base_connector_sample.rs
```diff
@@ -161,7 +162,7 @@ async fn run_dataset(dataset_client: DatasetClient) {
     }
     let mut count = 0;
     loop {
-        tokio::time::sleep(Duration::from_secs(5)).await;
+        tokio::time::sleep(Duration::from_secs(10)).await;
```
olivakar, line 165: any reason for increasing sleep ?

vaavva: just because it runs constantly and there isn't a way to easily pause execution of the pod (that I know about at least), I wanted it to execute a little bit less while still making it easy to validate that things are working

rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
```diff
@@ -205,63 +200,70 @@ impl DeviceEndpointClientCreationObservation {
...
+        device_update_observation: DeviceUpdateObservation,
         connector_context: Arc<ConnectorContext>,
         // TODO: This won't need to return an error once the service properly sends errors if the endpoint doesn't exist
     ) -> Result<Self, String> {
         Ok(DeviceEndpointClient {
             // TODO: get device_endpoint_credentials_mount_path from connector config
-            specification: Arc::new(DeviceSpecification::new(
+            specification: Arc::new(RwLock::new(DeviceSpecification::new(
                 device.specification,
                 "/etc/akri/secrets/device_endpoint_auth",
```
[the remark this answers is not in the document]
vaavva: see the TODO above this :) `// TODO: get device_endpoint_credentials_mount_path from connector config` Once Carter has updated the file mount config, this won't be hardcoded anymore

rust/azure_iot_operations_connector/examples/base_connector_sample.rs
vaavva: gross, fixed. Thanks for catching!

PR #812, opened 2025-05-22
[Rust][Services][Connector] Update to latest `adr-base-service` DTDL for ADR

cartertinney:
- Moved the ADR client implementation to use the new `adr-base-service.json` DTDL instead of the old `device-name-based-operations.json`
- Renamed `EventsAndStreamsDestination` to `EventStreamDestination` to match latest DTDL concepts
- Regenerated code


rust/azure_iot_operations_services/src/azure_device_registry.rs
```diff
@@ -1108,63 +1104,29 @@ impl From<adr_name_gen::AssetDatasetDataPointSchemaElementSchema> for DatasetDat
...
+impl From<base_client_gen::DatasetDestination> for DatasetDestination {
+    fn from(value: base_client_gen::DatasetDestination) -> Self {
         DatasetDestination {
             configuration: value.configuration.into(),
             target: value.target.into(),
         }
     }
 }
 
-impl From<adr_name_gen::DefaultDatasetsDestinationsSchemaElementSchema> for DatasetDestination {
```
vaavva, line 1120: There's a few From impls that you've removed - is there a change that makes them not needed, or were they not needed before as well?

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -514,6 +516,7 @@ where
             .get_device_command_invoker
             .invoke(get_device_request)
             .await
+            .map_err(ErrorKind::from)?
```
vaavva, line 519: what changed here to need an extra error mapping?

vaavva (approved):

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
olivakar: This is for the service error i think

rust/azure_iot_operations_services/src/azure_device_registry.rs
cartertinney: There is no longer a bespoke generated type for a Default Destination. Destinations are now standardized

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
cartertinney: Codegen Invoker now returns Result<Result<Response, AkriServiceError>, AIOProtocolError>.

Basically, the placeholder `ServiceError` is no longer required - codegen gives us a real one now.

cartertinney: > u might need to update docs if there is any other error

Done

PR #834, opened 2025-05-30
[Rust][Services] ADR discovery models + APIs

cartertinney:
- Added new APIs and data models
- Changed QoS enum to be distinct (but convertible to) the QoS from MQTT crate
- Added helpers for conversions of maps
- Changed optionality of nested model structs to better reflect truth of DTDL contract

unit tests in progress


rust/azure_iot_operations_services/src/azure_device_registry/models/device.rs
```diff
@@ -156,33 +220,61 @@ impl From<base_client_gen::DeviceSpecificationSchema> for DeviceSpecification {
...
+    }
+}
+
+impl From<DiscoveredOutboundEndpoints>
+    for discovery_client_gen::DiscoveredDeviceOutboundEndpointsSchema
+{
+    fn from(value: DiscoveredOutboundEndpoints) -> Self {
+        discovery_client_gen::DiscoveredDeviceOutboundEndpointsSchema {
+            assigned: value
+                .assigned
```
vaavva, line 274: can you use the option_map_into here?

rust/azure_iot_operations_services/src/azure_device_registry/models/device.rs
```diff
@@ -156,33 +220,61 @@ impl From<base_client_gen::DeviceSpecificationSchema> for DeviceSpecification {
...
+            outbound: value.outbound.map(Into::into),
+        }
+    }
+}
+
+impl From<base_client_gen::OutboundSchema> for OutboundEndpoints {
+    fn from(value: base_client_gen::OutboundSchema) -> Self {
+        OutboundEndpoints {
+            assigned: value
+                .assigned
```
vaavva, line 259: can you use the option_map_into here?

rust/azure_iot_operations_services/src/azure_device_registry/models/device.rs
```diff
@@ -196,6 +288,24 @@ impl From<base_client_gen::DeviceOutboundEndpoint> for OutboundEndpoint {
     }
 }
 
+impl From<discovery_client_gen::DeviceOutboundEndpoint> for OutboundEndpoint {
```
vaavva, line 291: is this direction of the From needed?

rust/azure_iot_operations_services/src/azure_device_registry/models/asset.rs
```diff
@@ -89,6 +89,54 @@ pub struct AssetSpecification {
     pub version: Option<u64>,
 }
 
+/// Represents the partial specification of an Discovered Asset in the Azure Device Registry service.
+pub struct DiscoveredAssetSpecification {
+    /// URI or type definition ids.
+    pub asset_type_refs: Vec<String>, // if None, we can represent as empty vec. Can currently only be length of 1
```
vaavva, line 95: > Can currently only be length of 1

This was okay to have just in the documentation for received structs, but since this is a struct that will be sent, we should consider adding validation for this. And for all other fields like this. Check the discovered_asset.yaml to find these restrictions (or confirm with Abhipsa if this is something that's fine for the edge to send and then the operator would filter out)

vaavva: Did a quick look and not all fields that can only be length of 1 on the non discovered flow have the same limit for the discovered flow, fyi

rust/azure_iot_operations_services/src/azure_device_registry/models/asset.rs
```diff
@@ -420,6 +588,16 @@ pub enum ActionType {
     Write,
 }
 
+/// Represents the MQTT Quality of Service level.
+/// Currently does not include Quality of Service 2.
+#[derive(Clone, Debug)]
+pub enum Qos {
```
vaavva, line 594:
```suggestion
pub enum QoS {
```
For our enum, can we use the more correct capitalization scheme?

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -735,6 +786,62 @@ where
         }
     }
 
+    /// Creates or updates a discovered device in the Azure Device Registry service.
```
vaavva, line 789: Can you add some information about how to use this for an update? Something like:
To perform an update, the entire existing device specification plus changes must be included. This operation will do a full replace

rust/azure_iot_operations_services/src/azure_device_registry/models/asset.rs
cartertinney: Sure

rust/azure_iot_operations_connector/src/destination_endpoint.rs
```diff
@@ -342,7 +342,7 @@ impl Destination {
                         message: Some(e.to_string()),
                     })?;
                     Destination::Mqtt {
-                        qos: definition_destination.configuration.qos,
+                        qos: definition_destination.configuration.qos.clone().map(Into::into),
```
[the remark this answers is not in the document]
cartertinney: Sure

rust/azure_iot_operations_services/src/azure_device_registry/models/device.rs
cartertinney: no

rust/azure_iot_operations_services/src/azure_device_registry/models/device.rs
cartertinney: No, see other comment

rust/azure_iot_operations_services/src/azure_device_registry/models/device.rs
```diff
@@ -51,15 +57,57 @@ pub struct DeviceSpecification {
     pub version: Option<u64>,
 }
 
-#[derive(Debug, Clone, Default)]
+#[derive(Debug, Clone)]
+/// Represents a discovered device specification in the Azure Device Registry service.
+pub struct DiscoveredDeviceSpecification {
```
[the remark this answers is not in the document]
cartertinney: Should we? I feel like service is the ideal place to do validation, I'd prefer to do as little as possible client-side

rust/azure_iot_operations_services/src/azure_device_registry/models/asset.rs
cartertinney: I"ll just remove it wholesale - I'm not trying to validate this right now.

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -735,6 +786,62 @@ where
...
+    /// * `timeout` - The duration until the client stops waiting for a response to the request, it is rounded up to the nearest second.
+    ///
+    /// Returns tuple containing the discovery ID and version of the discovered device.
+    ///
+    /// # Errors
+    /// [`struct@Error`] of kind [`InvalidRequestArgument`](ErrorKind::InvalidRequestArgument)
+    /// if timeout is 0 or > `u32::max`.
+    ///
+    /// [`struct@Error`] of kind [`AIOProtocolError`](ErrorKind::AIOProtocolError) if:
+    /// - device or inbound endpoint names are invalid.
```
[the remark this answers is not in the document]
cartertinney: i do not like this strange emerging hierarchy of validation, we have to find a way to tighten this up, this is going to be extremely prone to oversight

rust/azure_iot_operations_services/src/azure_device_registry/models/device.rs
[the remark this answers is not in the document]
vaavva: True, that does feel a bit more like a service-side validation since it's more defined on the contents than something simple like empty string. That makes sense - can you validate that if this error is returned, nothing super crazy happens?

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -1529,6 +1717,11 @@ mod tests {
         ));
     }
 
+    // #[tokio::test]
```
vaavva, line 1720: accidentally left in?

rust/azure_iot_operations_services/src/azure_device_registry/models/device.rs
```diff
@@ -51,15 +57,57 @@ pub struct DeviceSpecification {
...
-    /// The 'outboundUnassigned' Field.
-    pub outbound_unassigned: HashMap<String, OutboundEndpoint>,
+    pub outbound: Option<OutboundEndpoints>,
+}
+
+/// Represents the endpoints of a discovered device in the Azure Device Registry service.
+#[derive(Debug, Clone)]
+pub struct DiscoveredDeviceEndpoints {
+    /// The 'inbound' Field.
+    pub inbound: HashMap<String, DiscoveredInboundEndpoint>, // if None, we can represent as empty hashmap. Might be able to change this to a single InboundEndpoint
```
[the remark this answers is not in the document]
vaavva: I think this change got reverted with your updates to the None/empty comments

PR #847, opened 2025-06-02
Add DTMI resolution to ProtocolCompiler

jrdouceur:
This PR adds support for resolving external identifiers in models to the ProtocolCompiler.  This is needed for generating code from models that have been converted from OPC UA companion specs, because the companion specs are hierarchical and reference each other.


timtay-microsoft (approved):

jrdouceur:
@avishekpant, Tim approved this PR, but it seems I need your approval to continue.  Not sure why, though.

avishekpant (approved):

PR #850, opened 2025-06-03
[Rust] [Services] codegen from latest dtdl

vaavva:
Adds
- split between getDevice/getDeviceStatus (and same for Asset)
- descriptions
- error code enum
- removal of <device/asset>Specification layer


cartertinney (approved):
Approved

PR #856, opened 2025-06-04
[Rust] [Services] description updates from ADR DTDL

vaavva:
This PR:
- updates descriptions that were out of sync with the DTDL descriptions after the latest DTDL update

Note:
- There were a few places that I made editorial choices, but this is mostly true to the DTDL. The few changes I made were where we already had something more descriptive (i.e. the DTDL still said something like `The 'config' Field.`) or some slight edits for the audience of this API vs a more cloud-oriented audience (these I all approved with Abhipsa)
- You'll notice that Discovered items don't have any updates - these currently still don't have descriptions in the DTDL, so I won't worry about touching them until they do


rust/azure_iot_operations_services/src/azure_device_registry/models/asset.rs
```diff
@@ -193,9 +193,9 @@ pub struct DiscoveredDatasetDataPoint {
 /// Represents the destination for a dataset.
 #[derive(Clone, Debug)]
 pub struct DatasetDestination {
-    /// The configuration for the destination
+    /// The destination configuration.
```
olivakar, line 196: could this be more descriptive ?

rust/azure_iot_operations_services/src/azure_device_registry/models/asset.rs
```diff
@@ -211,9 +211,9 @@ pub struct DatasetDestination {
 /// Represents the destination for an event or stream.
 #[derive(Clone, Debug)]
 pub struct EventStreamDestination {
-    /// The configuration for the destination
+    /// The destination configuration.
```
olivakar, line 214: same as above

rust/azure_iot_operations_services/src/azure_device_registry/models/device.rs
```diff
@@ -86,9 +86,9 @@ pub struct DiscoveredDeviceEndpoints {
 /// Represents the outbound endpoints of a device in the Azure Device Registry service.
 #[derive(Debug, Clone)]
 pub struct OutboundEndpoints {
-    /// The 'assigned' Field.
+    /// Device messaging endpoint model.
```
olivakar, line 89: any difference in description between assigned and unassigned ?

rust/azure_iot_operations_services/src/azure_device_registry/models/asset.rs
```diff
@@ -226,29 +226,29 @@ pub struct EventStreamDestination {
...
 
-/// A reference to the Device and Endpoint within the device
+/// A reference to the Device and Endpoint within the device (connection information) used by brokers to connect that provides data points for this asset.
 #[derive(Clone, Debug)]
 pub struct DeviceRef {
-    /// The name of the device
+    /// Name of the device resource.
     pub device_name: String,
-    /// The endpoint name of the device
+    /// The name of endpoint to use.
```
cartertinney, line 234: Feel like we've lost some information about the endpoint here

rust/azure_iot_operations_services/src/azure_device_registry/models/asset.rs
```diff
@@ -350,13 +350,13 @@ pub struct DiscoveredManagementGroupAction {
 /// Represents a stream for an asset.
 #[derive(Clone, Debug)]
 pub struct Stream {
-    /// Destinations for a stream.
+    /// Destinations for a Stream.
```
cartertinney, line 353: Capitalized or not? Inconsistent, even within this struct.

rust/azure_iot_operations_services/src/azure_device_registry/models/asset.rs
```diff
@@ -458,22 +458,22 @@ pub struct ManagementGroupStatus {
 pub struct ActionStatus {
     /// The configuration error of the management group action.
     pub error: Option<ConfigError>,
-    /// The name of the management group action.
+    /// The name of the action. Must be unique within the status.managementGroup[i].actions array. This name is used to correlate between the spec and status management group action information.
     pub name: String,
-    /// The request message schema reference for the management group action.
+    /// The request message schema reference object.
```
cartertinney, line 463: Does "object" provide any semantic value here?

cartertinney (approved):

rust/azure_iot_operations_services/src/azure_device_registry/models/device.rs
vaavva: I thought this one was pretty weird too! But the new descriptions from the DTDL (what I have here) still at least felt better than what we had before, and I'm not sure that I know enough to come up with something else

rust/azure_iot_operations_services/src/azure_device_registry/models/asset.rs
vaavva: don't necessarily disagree, but given the possible subtle nuances of differing language here, I'd rather match the DTDL exactly for this one.

rust/azure_iot_operations_services/src/azure_device_registry/models/asset.rs
vaavva: there were a lot of changes in the 6-5 dtdl update specific to capitalization changes, so I'm just matching it exactly

rust/azure_iot_operations_services/src/azure_device_registry/models/asset.rs
vaavva: This is directly what's from the DTDL - I don't want to differ from it too much unless we already had something way better or there's something that isn't relevant for our audience

PR #859, opened 2025-06-05
[Rust][Services] ADR Client 6/5 DTDL Update

vaavva:
(ignore my branch name, guess I was on autopilot)
This PR:
- updates the ADR Client with the latest 6/5 DTDL. Description updates from this DTDL will be added in the descriptions PR (#856 ) all together

### DIFF
--- a/rust/azure_iot_operations_services/src/azure_device_registry.rs
+++ b/rust/azure_iot_operations_services/src/azure_device_registry.rs
@@ -4,7 +4,6 @@
 //! Types for Azure Device Registry operations.
 
 use core::fmt::Debug;
-use std::collections::HashMap;
 
 use azure_iot_operations_mqtt::interface::AckToken;
 use azure_iot_operations_protocol::{common::aio_protocol_error::AIOProtocolError, rpc_command};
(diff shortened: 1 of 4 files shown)


RLeclair (approved):

rust/azure_iot_operations_services/src/azure_device_registry.rs
```diff
@@ -173,38 +170,18 @@ impl From<discovery_client_gen::CodeSchema> for base_client_gen::CodeSchema {
     }
 }
 
-impl From<StatusConfig> for base_client_gen::DeviceStatusConfigSchema {
+impl From<StatusConfig> for base_client_gen::ConfigStatus {
```
ewertons, line 173: From ConfigStatus to StatusConfig. Interesting.
I was going to ask if StatusConfig is a good name, but I think this is what Carl was sharing about earlier today.

ewertons (approved):

rust/azure_iot_operations_services/src/azure_device_registry.rs
vaavva: yeah, it's interesting - we originally had StatusConfig to match their DeviceStatusConfigSchema, but then they swapped it...not sure if we should take this swap or not - might be clearer? I'll make the change - never bad to be more consisten with the service

vaavva: Realized as well that I refer to this as the config status in all of my doc, so that's an extra good reason to make the change. Updated

PR #885, opened 2025-06-13
[Go] Error metadata workaround

clecompt-msft:


timtay-microsoft (approved):

go/protocol/command_executor.go
```diff
@@ -348,7 +348,12 @@ func (ce *CommandExecutor[Req, Res]) build(
 	rpub.CorrelationData = pub.CorrelationData
 	rpub.Topic = pub.ResponseTopic
 	rpub.MessageExpiry = pub.MessageExpiry
-	maps.Copy(rpub.UserProperties, errutil.ToUserProp(resErr))
+
+	// Generate user properties according to the error, but don't overwrite
+	// existing user-specified metadata.
```
vaavva, line 353: I'm not super familiar with Go, but does this skip adding the sdk properties if they're already in the user specified list, or does it keep both? I think there aren't guarantees around ordering of user properties in the MQTT spec, so to be safe we must only have one instance of each header (that we care/know about at least)

clecompt-msft: The Go SDK uses hashmaps for the headers, so it's only possible to have one copy of any given header. This change effectively just flips the priority order.

PR #896, opened 2025-06-17
[Rust][Services] DTDL changes for ADR Client

olivakar:


eng/dtdl/adr-base-service.json
```diff
@@ -17,19 +17,32 @@
     {
       "@type": "Command",
       "name": "getDevice",
+      "description": "Retrieve a device by device name and inbound endpoint name.",
```
vaavva, line 20: aren't there updates for the other DTDL as well? Or will that be part of a separate PR?

rust/azure_iot_operations_services/src/azure_device_registry/models/device.rs
```diff
@@ -49,19 +49,19 @@ pub struct Device {
 #[derive(Debug, Clone)]
 /// Represents a discovered device in the Azure Device Registry service.
 pub struct DiscoveredDevice {
-    /// The 'attributes' Field.
+    /// A set of key-value pairs that contain custom attributes set by the customer.
     pub attributes: HashMap<String, String>, // if empty hashmap, we can represent as None on generated model
-    /// The 'endpoints' Field.
+    /// Connection endpoint URL a device can use to connect to a service.
```
vaavva, line 54: this feels like the wrong description for this field. @abhipsaMisra is this accurate? (Oliva, don't hold the PR on this comment)

vaavva (approved):

PR #902, opened 2025-06-19
[Rust][Connector Scaffolding] Connector scaffolding 

RLeclair:
This PR:
- Adds a connector scaffolding template based on a pull model with a fixed interval.
- Adds a dockerfile to build a container out of it
- Adds a readme with instructions

I tested this against against an AIO deployment and it works out of the box.


rust/sample_connector_scaffolding/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+FROM mcr.microsoft.com/azurelinux/distroless/debug:3.0
+
+# Add busybox shortcuts for rational debugging
+RUN ["busybox", "--install"]
+
+# Add a symlink to enable SSL certs
+RUN mkdir /etc/ssl && ln -s /etc/pki/tls/certs /etc/ssl/certs
+
+# Copy the built application into the working directory
+COPY ./target/release/sample_connector_scaffolding .
```
clecompt-msft, line 11: This is missing the `chmod` fix and `distroless/base` stage from the updated REST dockerfile.

rust/sample_connector_scaffolding/src/main.rs
```diff
@@ -0,0 +1,699 @@
...
+        //     DatasetNotification::Updated => {
+        //         log::info!("{dataset_log_identifier} Dataset definition updated");
+        //     }
+        //     DatasetNotification::Deleted => {
+        //         log::info!("{dataset_log_identifier} Dataset deleted, ending dataset handler");
+        //         return;
+        //     }
+        //     DatasetNotification::UpdatedInvalid => {
+        //         log::info!("{dataset_log_identifier} Dataset definition invalid, waiting for a valid update");
+        //         loop {
```
vaavva, line 520: can the parent match statement just be a loop instead of having the loop with the same match again inside?

rust/sample_connector_scaffolding/src/main.rs
```diff
@@ -0,0 +1,699 @@
...
+        //             }
+        //         }
+        //     }
+        // }
+    };
+
+    // This variable keeps track of the latest reported schema.
+    let mut current_schema = None;
+
+    let mut timer = tokio::time::interval(DEFAULT_SAMPLING_INTERVAL);
```
vaavva, line 543:
```suggestion
    // this could be configured in the dataset_configuration instead if desired
    let mut timer = tokio::time::interval(DEFAULT_SAMPLING_INTERVAL);
```

rust/sample_connector_scaffolding/src/main.rs
```diff
@@ -0,0 +1,699 @@
...
+
+                // Create a data structure with the sampled data
+                let data = Data {
+                    payload: bytes,
+                    content_type: "application/json".to_string(),
+                    custom_user_data: vec![],
+                    timestamp: Some(HybridLogicalClock::new()),
+                };
+
+                // TODO: Update to pass-through
```
vaavva, line 632: need to do this for this PR, otherwise it won't compile anymore

rust/sample_connector_scaffolding/src/main.rs
```diff
@@ -0,0 +1,699 @@
...
+                // Transform the data and infer the message schema using the derived_json module. This works for JSON data only.
+                let Ok((data, message_schema)) = derived_json::transform(data, &local_dataset_definition) else {
+                    log::error!("{dataset_log_identifier} Failed to transform data");
+
+                    // Report a status error of the dataset.
+                    match dataset_client.report_status(Err(AdrConfigError {
+                        message: Some("Failed to transform data".to_string()),
+                        ..Default::default()
+                    })).await {
+                        Ok(()) => log::error!("{dataset_log_identifier} Dataset status reported as error"),
```
vaavva, line 642:
```suggestion
                        Ok(()) => log::debug!("{dataset_log_identifier} Dataset status reported as error"),
```

rust/sample_connector_scaffolding/src/main.rs
```diff
@@ -0,0 +1,699 @@
...
+                        Err(e) => log::error!("{dataset_log_identifier} Failed to report dataset status: {e}"),
+                    }
+                    // If we fail to report the message schema, we will not be able to forward it
```
vaavva, line 645:
```suggestion
                    // If we fail to generate the message schema, we will not be able to report it
```
not sure if this is the right comment, but this step didn't report the message schema, so there's definitely some change needed here

rust/sample_connector_scaffolding/src/main.rs
```diff
@@ -0,0 +1,699 @@
...
+                    is_dataset_asset_ready = false;
```
vaavva, line 646: also not sure if failing to generate the message schema/transform data should require a dataset update or if it could be just one malformed payload. I'd err on the side of continuing to sample without setting `is_dataset_asset_ready` to false

rust/sample_connector_scaffolding/src/main.rs
```diff
@@ -0,0 +1,699 @@
...
+                if current_schema.is_none() || current_schema.as_ref() != Some(&message_schema) {
+                    // Note, this operation already retries internally.
+                    log::info!("{dataset_log_identifier} Reporting message schema");
+                    match dataset_client.report_message_schema(message_schema.clone()).await {
+                        Ok(_) => {
+                            log::debug!("{dataset_log_identifier} Successfully reported message schema");
+                            current_schema = Some(message_schema);
+                        }
+                        Err(e) => {
+                            log::debug!("{dataset_log_identifier} Failed to report message schema, attempting in next interval: {e}");
```
vaavva, line 674:
```suggestion
                            log::error!("{dataset_log_identifier} Failed to report message schema, attempting in next interval: {e}");
```

clecompt-msft (approved):

rust/sample_applications/sample_connector_scaffolding/src/main.rs
```diff
@@ -0,0 +1,514 @@
...
+            }
+        }
+    }
+}
+
+/// Receives a device endpoint and creates a device handler for it.
+///
+/// # Arguments
+/// * `device_endpoint_client_creation_observation` - The device endpoint client creation observation.
+async fn receive_device_endpoints(
```
cartertinney, line 116: Is this not just a device_endpoint_handler?

rust/sample_applications/sample_connector_scaffolding/src/main.rs
```diff
@@ -169,137 +164,130 @@ async fn receive_device_endpoints(
...
+        // If knowledge of a successful connection can't be determined at this level, communication will need to be added from the location of
+        // the connection logic to this level to report the device and endpoint statuses.
+
+        // Notify any lower components that the device endpoint is in a ready state
+        device_endpoint_ready_watcher_tx.send_if_modified(send_if_modified_fn(true));
+        // if there was an error, notify any lower components that the device endpoint is not in a ready state while we wait for an update
+        // device_endpoint_ready_watcher_tx.send_if_modified(send_if_modified_fn(false));
+    }
+
+    // if the connection is successful or the device wasn't enabled and there weren't configuration errors, report the device and endpoint statuses.
```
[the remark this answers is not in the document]
vaavva: this comment was removed, but I looked through the rest of the file and fixed all other comments like this

rust/sample_applications/sample_connector_scaffolding/src/main.rs
```diff
@@ -0,0 +1,514 @@
...
+}
+
+/// Handles the device endpoint and receives the asset creation notifications with which it will
+/// create asset handlers.
+///
+/// # Arguments
+/// * `device_endpoint_log_identifier` - A string identifier for the device endpoint, used for logging.
+/// * `device_endpoint_client` - The device endpoint client.
+async fn device_handler(
+    device_endpoint_log_identifier: String,
```
cartertinney, line 148: I think we should probably bite the bullet even in the scaffolding and use a `SampleConnectorContext` or something here that contains the log identifier since:

1) When OTEL comes in, you'll need it to be here to show how metrics work
2) Artifacts can be shared on it, even if they're unused in this scaffolding, the option is there
3) It gets passed around in exactly the same way as this log identifier, so might as well just use the more scalable pattern

rust/sample_applications/sample_connector_scaffolding/src/main.rs
```diff
@@ -0,0 +1,517 @@
...
+    application::ApplicationContextBuilder, common::hybrid_logical_clock::HybridLogicalClock,
+};
+use tokio::sync::watch;
+
+const DEFAULT_SAMPLING_INTERVAL: Duration = Duration::from_millis(10000); // Default sampling interval in milliseconds
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // Initialize the logger
+    // IMPLEMENT: Use a more sophisticated logger configuration in production
```
cartertinney, line 62: Is this an implement? Or is this one an actual TODO on us?

rust/sample_applications/sample_connector_scaffolding/src/main.rs
```diff
@@ -0,0 +1,517 @@
...
+}
+
+/// Handles sampling of data from the dataset.
+///
+/// # Arguments
+/// * `dataset_log_identifier` - A string identifier for the dataset, used for logging.
+/// * `dataset_client` - The dataset client.
+/// * `asset_ready_watcher_rx` - A watcher for the asset readiness state.
+/// * `device_endpoint_ready_watcher_rx` - A watcher for the device endpoint readiness state.
+#[allow(unused_assignments)] // IMPLEMENT: Remove once variables are being used
```
cartertinney, line 344: Why not just use underscore prefixes?

cartertinney (approved):

rust/sample_applications/sample_connector_scaffolding/src/main.rs
vaavva: agree - I'm going to leave that as a follow-up PR for the sake of time today though

rust/sample_applications/sample_connector_scaffolding/src/main.rs
vaavva: because it's just one of the assignments that would be inside an if statement in a real implementation. There is a change I can make to remove this though, I'll do that instead

rust/sample_applications/sample_connector_scaffolding/src/main.rs
vaavva: changed back to TODO

rust/sample_applications/sample_connector_scaffolding/src/main.rs
vaavva: sync'd offline, won't make this change at least until OTEL is integrated

PR #915, opened 2025-07-07
Merge main changes into feature akri

olivakar:


rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -58,7 +58,7 @@ where
         Arc<base_client_gen::SetNotificationPreferenceForDeviceUpdatesCommandInvoker<C>>,
     create_or_update_discovered_device_command_invoker:
         Arc<discovery_client_gen::CreateOrUpdateDiscoveredDeviceCommandInvoker<C>>,
-    device_update_notification_dispatcher: Arc<Dispatcher<(Device, Option<AckToken>)>>,
+    device_update_notification_dispatcher: Arc<Dispatcher<(Device, Option<AckToken>), String>>,
```
vaavva, line 61: wasn't the whole goal of the dispatcher change to be able to use the DeviceEndpointRef and DeviceEndpointAssetRef as the hashmap key instead of using the hash function to a String?

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -466,11 +466,11 @@ where
                                 Ok(()) => {
                                     log::debug!("Device Update Notification dispatched for device {device_name:?} and inbound endpoint {inbound_endpoint_name:?}");
                                 }
-                                Err(DispatchError::SendError(tokio::sync::mpsc::error::SendError((payload, _)))) => {
-                                    log::warn!("Device Update Observation has been dropped. Received Device Update Notification: {payload:#?}");
+                                Err(DispatchError { data, kind: DispatchErrorKind::SendError }) => {
+                                    log::warn!("Device Update Observation has been dropped. Received Device Update Notification: {data:#?}");
```
vaavva, line 470: small nit that's been driving me crazy when debugging, can you change this to not be a pretty print anymore while you're in here?
```suggestion
                                    log::warn!("Device Update Observation has been dropped. Received Device Update Notification: {data:?}");
```

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -466,11 +466,11 @@ where
...
                                 }
-                                Err(DispatchError::NotFound((receiver_id, (payload, _)))) => {
-                                    log::warn!("Device Endpoint is not being observed. Received Device Update Notification: {payload:#?} for {receiver_id:?}");
+                                Err(DispatchError { data, kind: DispatchErrorKind::NotFound(receiver_id) }) => {
+                                    log::warn!("Device Endpoint is not being observed. Received Device Update Notification: {data:#?} for {receiver_id:?}");
```
vaavva, line 473:
```suggestion
                                    log::warn!("Device Endpoint is not being observed. Received Device Update Notification: {data:?} for {receiver_id:?}");
```

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -466,11 +466,11 @@ where
                                 Ok(()) => {
                                     log::debug!("Device Update Notification dispatched for device {device_name:?} and inbound endpoint {inbound_endpoint_name:?}");
                                 }
-                                Err(DispatchError::SendError(tokio::sync::mpsc::error::SendError((payload, _)))) => {
-                                    log::warn!("Device Update Observation has been dropped. Received Device Update Notification: {payload:#?}");
+                                Err(DispatchError { data, kind: DispatchErrorKind::SendError }) => {
```
vaavva, line 469: since before we were only printing the payload and not the ack token, would we want to keep that functionality? Might be something like this? Not sure on the compiler syntax
```suggestion
                                Err(DispatchError { (payload, _), kind: DispatchErrorKind::SendError }) => {
```
here and for the other 3 spots in this file

rust/azure_iot_operations_services/src/state_store/client.rs
```diff
@@ -626,11 +629,13 @@ where
                                     Ok(()) => {
                                         log::debug!("Key Notification dispatched: {key_notification:?}");
                                     }
-                                    Err(DispatchError::SendError(_)) => {
-                                        log::warn!("Key Notification Receiver has been dropped. Received Notification: {key_notification:?}",);
+
+                                    Err(DispatchError { data, kind: DispatchErrorKind::SendError }) => {
+                                        log::warn!("Key Notification Receiver has been dropped. Received Notification: {data:?}");
```
vaavva, line 634: same here, probably don't want to print the ack token info

rust/azure_iot_operations_services/src/azure_device_registry/client.rs
```diff
@@ -242,37 +242,27 @@ where
     /// Convenience function to get all observed device & inbound endpoint names to quickly unobserve all of them before cleaning up
     #[must_use]
     pub fn get_all_observed_device_endpoints(&self) -> Vec<(String, String)> {
```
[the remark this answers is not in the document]
olivakar: https://dev.azure.com/msazure/One/_workitems/edit/33795008

PR #922, opened 2025-07-10
Add ADR 23, SDK Support for Property

jrdouceur:


jrdouceur:

clecompt-msft:
This pattern is very similar to DSS. Has any thought been considered towards merging the two?

vaavva (commented):
This is a really interesting pattern! I have a lot of thoughts on how to make this even more powerful without much extra work (since we've already implemented a lot of it in our service client implementations), and some areas that I want to make sure we don't miss out on needed functionality - please set up some time with me!

doc/dev/adr/0023-property-support.md
```diff
@@ -0,0 +1,211 @@
...
+
+The Consumer issues requests to the Maintainer, and it processes change notifications it receives from the Maintainer.
+
+The values of all Properties within an Interface must be readable atomically, writable atomically, and notifiable atomically.
+User code is not under any obligation to respect this atomicity, but the envoys are obligated to provide an API that readily supports atomicity if the application demands it.
+
+There are five specific actions that are relevant to Properties:
+
+* Action *write*:
+  * The Consumer sends a `write` request to the Maintainer, specifying new values for one or more Properties.
```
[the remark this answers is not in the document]
jrdouceur: I had heard that supporting multiple properties atomically was a requirement, which is what led me to this design.  I don't know much detail, but I understood there to be cases that demand consistent values across multiple properties, particularly for use with OPC UA.

doc/dev/adr/0023-property-support.md
```diff
@@ -0,0 +1,211 @@
...
+The *write*, *read*, *watch*, and *unwatch* actions have behaviors that align with the RPC communication pattern.
+The *notify* action has a behavior that aligns with the Telemetry communication pattern.
+
+## Property envoy
+
+A Property envoy is an assemblage of four Command envoys and one Telemetry envoy.
+It is parameterized by two types:
+
+* `TProp` &mdash; A structure type (class/struct/object) that collates related Properties
+  * There is one field in the `TProp` structure per Property in the collation.
```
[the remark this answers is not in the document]
jrdouceur: I agree that this topic in particular seems to warrant an interactive discussion.  I sent out an invitation for later today.  I'll make a few comments now, though....

> does this mean that all properties have to be known at the time of creating a DTDL?

Yes, but this is nothing new.  A DTDL model must explicitly specify all Commands, all Telemetries, and all Properties statically.  This is no direct way to support dynamic contents in DTDL.  One can define a DTDL Map, which supports a dynamic set of keys, but the Map values must all have the same schema type.  A completely generic structure with arbitrary names and arbitrary value schemas would essentially be saying that the data is unmodeled.

Although I wrote this ADR with the expectation that it would be targetable from DTDL via the ProtocolCompiler, it is intended to stand on its own as an SDK feature usable in non-codegen scenarios.  The goal is to provide functional commonality that helps to ensure consistent behavior across programming languages, much as our RPC and Telemetry envoys do.  In support of this goal, I intend to add Property-specific syntax to the METL testing language and to write METL test cases for Property.

With that in mind, I don't really see how -- even in non-codegen cases -- we could cleanly support run-time dynamic Property types, except in an interpreted language like JavaScript or Python.  Technically, .NET offers a way to construct and read types that are defined at runtime, but the mechanism is horribly awkward to use.  And if Rust offers such a mechanism, it is not one I am familiar with.

For what it's worth, I did consider using an array of names instead of a struct of Booleans for the second type param, but I went with the latter for two reasons:  First, an array admits the possibility of runtime errors in the Property names, whereas parallel structs do not incur this risk.  Second, in codegen scenarios, the ProtocolCompiler can generate both structs from the same set of DTDL definitions, which ensures that they are appropriately parallel.

> If this more locked down pattern of non-dynamic properties is desired for other solutions, I would propose that being a codegen wrapper (or protocol wrapper if we don't want it just in codegen) on top of this pattern I'm proposing
>
> I'm wondering if we're thinking about this fairly differently

Quite possibly.  The intent of this feature is to provide functionality analogous to what RPC/Command and Telemetry provide, but with Property semantics, meaning stateful values with explicit read, write, and notify-on-update.  My originally considered approach was the alternative stated in the doc's intro: "Another design option is for the SDKs to provide native support for Property that is parallel to — and independent of — their support for RPC and Telemetry."  Perhaps if it had been defined this way, the intended semantics for the design would have been clearer.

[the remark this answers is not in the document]
jrdouceur: I should also note that there is nothing in this proposal that restricts it to one instance per Interface.  For Telemetry, it is even possible from DTDL to generate either one Telemetry envoy per DTDL Telemetry or one Telemetry envoy for all DTDL Telemetries in an Interface.  (The distinction is implied by whether the telemetry topic pattern contains the "{telemetryName}" token.)  For Property as described in this ADR, the SDK feature also allows for either of these two uses.  (However, the ProtocolCompiler support I'm designing does not allow for the non-combined case, because it would add a fair bit of complexity to support a scenario that is not in our near-term plan.)

doc/dev/adr/0023-property-support.md
```diff
@@ -0,0 +1,312 @@
...
+
+The Consumer issues requests to the Maintainer, and it processes change notifications it receives from the Maintainer.
+
+The values of all Properties within an Interface must be readable atomically, writable atomically, and notifiable atomically.
+User code is not under any obligation to respect this atomicity, but the envoys are obligated to provide an API that readily supports atomicity if the application demands it.
+
+There are five specific actions that are relevant to Properties:
+
+* Action *write*:
+  * The Consumer sends a `write` request to the Maintainer, specifying new values for one or more Properties.
```
timtay-microsoft, line 35: The verbiage here strikes me as a bit odd. "Consumers" don't generally write. Should we refer to them as "Users" instead? Or "Actors"?

doc/dev/adr/0023-property-support.md
```diff
@@ -0,0 +1,312 @@
...
+  * The semantics of `TStatVal` values are determined by the service that implements the property.
+
+The *read* action is performed by issuing a 'read' Command request.
+
+* The request payload is an instance of `TCtrl`.
+  * For statically itemized properties, the optional fields in the `TCtrl` request object are `true` for each field whose Property is to be read.
+  * For dynamically itemized properties, the elements in the `TCtrl` request array indicate the names of the Properties to read.
+* The response payload is an instance of `TProp`.
+  * For statically itemized properties, the optional fields in the `TProp` response object contain values for any Properties that have been read.
+  * For dynamically itemized properties, the keys in the `TProp` response map indicate which Properties have been read, and the corresponding values are what was read.
```
vaavva, line 197: For static/dynamic, is there a scenario where some values that were requested weren't returned? (I think that feels possible), and if so, do we need some way to communicate why it wasn't returned? Ex. item doesn't exist vs timeout vs disk error

doc/dev/adr/0023-property-support.md
```diff
@@ -0,0 +1,312 @@
...
+
+The ProtocolCompiler will be enhanced to support the DTDL Property content type.
+Generated code will target the new Property envoy classes in the SDKs.
+This requires an additive change to the DTDL Mqtt extension, and since version 4 of this extension has not officially shipped, this change will be rolled into the pending Mqtt extension version 4.
+
+## Property roles and semantics
+
+Just as RPC encompasses the two roles of Executor and Invoker, and as Telemetry encompasses the two roles of Sender and Receiver, the Property pattern encompasses two roles: Maintainer and Consumer.
+
+The Maintainer's responsibilities are to hold the Property state, to read and update the state upon request, and to configurably distribute notifications of state changes.
```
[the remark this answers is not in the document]
vaavva: I realized what part of this concern is as well for me - there's the difference between the over the wire protocol vs what the envoys do vs what the application does - I think that would be the better semantic split to describe in this ADR, even more than just SDK vs customer application responsibilities.

doc/dev/adr/0023-property-support.md
```diff
@@ -0,0 +1,312 @@
...
+* The response payload is an instance of `TProp`.
+  * For statically itemized properties, the optional fields in the `TProp` response object contain values for any Properties that have been read.
+  * For dynamically itemized properties, the keys in the `TProp` response map indicate which Properties have been read, and the corresponding values are what was read.
+
+The *observe* action is performed by issuing an 'observe' Command request.
+
+* The request payload is an instance of `TCtrl`.
+  * For statically itemized properties, the optional fields in the `TCtrl` request object are `true` for each field whose Property is to be added to the notify list.
+  * For dynamically itemized properties, the elements in the `TCtrl` request array indicate the names of the Properties to add to the notify list.
+  * The maintainer may keep a single notify list across all consumers, or it may keep a separate notify list for each consumer; in the latter case, the maintainer must be provided with a way to identify the consumer that requested the observation.
```
vaavva, line 204: I would propose that this is functionality that we provide in our SDK (keeping the list of where notifications should be sent), and then having the customer component just deciding whether to use their client id as the scope for setting/receiving notifications or some other shared key as the scope. I still feel that supporting non-consumer specific notifications is a very concerning/confusing scenario, since another consumer can unobserve notifications that you wanted, but I understand this is a scenario we need to support

doc/dev/adr/0023-property-support.md
```diff
@@ -0,0 +1,312 @@
...
+  * For dynamically itemized properties, the elements in the `TCtrl` request array indicate the names of the Properties to remove from the notify list.
+  * The maintainer may keep a single notify list across all consumers, or it may keep a separate notify list for each consumer; in the latter case, the maintainer must be provided with a way to identify the consumer that requested the observation.
+* The response payload is an instance of `TStat`.
+  * For statically itemized properties, each optional field in the `TStat` response object indicates the status of the unobserve operation for the Property named by the field.
+  * For dynamically itemized properties, each value in the `TStat` response map indicates the status of the unobserve operation for the Property named by the key.
+  * The semantics of `TStatVal` values are determined by the service that implements the property.
+  * The maintainer's response may additionally provide status values for the observation state of properties other than those named in the 'unobserve' request.
+
+The *notify* action is performed by sending a Telemetry with a payload that is an instance of `TProp`.
```
vaavva, line 223: If we support shared notification lists (where one consumer can affect the notifications that they and other consumers will receive), I think it is important to have a notification for when something becomes unobserved (and potentially for when something becomes observed as well for symmetry, although I don't think the ramifications of not having that one are as large)

doc/dev/adr/0023-property-support.md
```diff
@@ -0,0 +1,312 @@
...
+  * For dynamically itemized properties, the elements in the `TCtrl` request array indicate the names of the Properties to add to the notify list.
+  * The maintainer may keep a single notify list across all consumers, or it may keep a separate notify list for each consumer; in the latter case, the maintainer must be provided with a way to identify the consumer that requested the observation.
+* The response payload is an instance of `TStat`.
+  * For statically itemized properties, each optional field in the `TStat` response object indicates the status of the observe operation for the Property named by the field.
+  * For dynamically itemized properties, each value in the `TStat` response map indicates the status of the observe operation for the Property named by the key.
+  * The semantics of `TStatVal` values are determined by the service that implements the property.
+  * The maintainer's response may additionally provide status values for the observation state of properties other than those named in the 'observe' request.
```
vaavva, line 209: I really liked your idea that
> it seemed to me that there should be some mechanism to find which items are currently on the watch list, and rather than adding another Command for this, it can be done by sending an empty list of additions in a "watch" Command.

I think this concept would be worth calling out here!

doc/dev/adr/0023-property-support.md
```diff
@@ -0,0 +1,312 @@
...
+The *read* action is performed by issuing a 'read' Command request.
+
+* The request payload is an instance of `TCtrl`.
+  * For statically itemized properties, the optional fields in the `TCtrl` request object are `true` for each field whose Property is to be read.
+  * For dynamically itemized properties, the elements in the `TCtrl` request array indicate the names of the Properties to read.
+* The response payload is an instance of `TProp`.
+  * For statically itemized properties, the optional fields in the `TProp` response object contain values for any Properties that have been read.
+  * For dynamically itemized properties, the keys in the `TProp` response map indicate which Properties have been read, and the corresponding values are what was read.
+
+The *observe* action is performed by issuing an 'observe' Command request.
```
vaavva, line 199: please add to this doc whether the notification topic would be set as part of this command or determined by the maintainer some other way. (Of course, the envoy would probably set the notification topic to the same thing on every request, similar to how it's done in RPC, but I think it's good to define how the over the wire protocol for this would work)

doc/dev/adr/0023-property-support.md
```diff
@@ -0,0 +1,312 @@
...
+  * The maintainer's response may additionally provide status values for the observation state of properties other than those named in the 'observe' request.
+
+The *unobserve* action is performed by issuing an 'unobserve' Command request.
+
+* The request payload is an instance of `TCtrl`.
+  * For statically itemized properties, the optional fields in the `TCtrl` request object are `true` for each field whose Property is to be removed from the notify list.
+  * For dynamically itemized properties, the elements in the `TCtrl` request array indicate the names of the Properties to remove from the notify list.
+  * The maintainer may keep a single notify list across all consumers, or it may keep a separate notify list for each consumer; in the latter case, the maintainer must be provided with a way to identify the consumer that requested the observation.
+* The response payload is an instance of `TStat`.
+  * For statically itemized properties, each optional field in the `TStat` response object indicates the status of the unobserve operation for the Property named by the field.
```
vaavva, line 218: would this describe everything that was successfully unobserved, or would this response describe the state of the current (still active) observations? I think there are clarity arguments for either option, unfortunately

doc/dev/adr/0023-property-support.md
```diff
@@ -0,0 +1,312 @@
...
+Constraints for statically itemized properties:
+
+* `TProp` &mdash; A structure type (class/struct/object) that collates related Properties
+  * There is one field in the `TProp` structure per Property in the collation.
+  * Each field may have an arbitrary type.
+  * Each field has a value that is optional/nullable.
+* `TCtrl` &mdash; a structure type whose field names match those in `TProp`
+  * Each field has type Boolean.
+* `TStat` &mdash; a structure type whose field names match those in `TProp`
+  * Each field has type `TStatVal`, which is &mdash; consistently across all fields &mdash; one of Boolean, integer, string, or an enumeration that is convertable to/from integer or string.
```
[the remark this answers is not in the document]
vaavva: I think what I'm really proposing is that "an enumeration that is convertible to/from integer or string" could just mean "serializable", which if that's what you intended, then we may already be at the functionality that I would want.

doc/dev/adr/0023-property-support.md
```diff
@@ -0,0 +1,312 @@
...
+      "name": "Foo",
+      "schema": "integer"
+    },
+    {
+      "@type": "Property",
+      "name": "Bar",
+      "schema": "string"
+    },
+    {
+      "@type": [ "Property", "Fragmented" ],
```
vaavva, line 275: Should dynamically itemized properties be defined in the same chunk of DTDL as the statically itemized properties since they won't be generated together (but the statically itemized properties would be)?

doc/dev/adr/0023-property-support.md
```diff
@@ -0,0 +1,312 @@
+# ADR 23: SDK Support for Property
+
+## Context
+
+The AIO SDKs currently support two communication patterns: RPC and Telemetry.
+There is no support for the stateful communication pattern we call Property.
+There is an upcoming need to support this pattern because it is used heavily in OPC UA companion specifications, which AIO intends to support during [Br].
+
```
vaavva, line 8: Adding a high level comment - this pattern feels SO valuable, and it's already clear because we're using a _very_ similar pattern already in the ADR and DSS clients (that we do share a chunk of code for already as well!). So, a large goal that I would like to see for this pattern would be to make it generalizable enough to allow it to be used for these already defined scenarios, as well as (of course) the scenarios that this ADR was proposed for.

vaavva (commented):
Varun asked me to document my feedback on this PR from our meeting/emails - I think most of what I've commented here we've already discussed at least a little bit, but now it'll be tracked in one central place!

jrdouceur:
> Varun asked me to document my feedback on this PR from our meeting/emails - I think most of what I've commented here we've already discussed at least a little bit, but now it'll be tracked in one central place!

Thanks, this is really helpful.  I'll respond to some of these here, and I'll also include an itemization of these concerns in the slides I'm starting to put together for next week's design review.

doc/dev/adr/0023-property-support.md
```diff
@@ -0,0 +1,312 @@
...
+## Property envoy
+
+A Property envoy is an assemblage of four Command envoys and one Telemetry envoy.
+It is parameterized by three types: `TProp`, `TCtrl`, and `TStat`.
+These types must satisfy exactly one of the following two sets of constraints.
+
+Constraints for statically itemized properties:
+
+* `TProp` &mdash; A structure type (class/struct/object) that collates related Properties
+  * There is one field in the `TProp` structure per Property in the collation.
```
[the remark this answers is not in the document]
jrdouceur: The particular line you attached this comment to, "There is one field in the `TProp` structure per Property in the collation." is a usage description.  If you add a field to the structure, it is a Property in the collation; if you don't, it is not.

doc/dev/adr/0023-property-support.md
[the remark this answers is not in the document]
jrdouceur: > I do still think that the application should be able to define their own return type so that they could have a more robust error format

This was my intent with "an enumeration that is convertable to/from integer or string".

> I think what I'm really proposing is that "an enumeration that is convertible to/from integer or string" could just mean "serializable", which if that's what you intended, then we may already be at the functionality that I would want.

This works in Rust -- thanks to serde -- but not in any other language.  Since other languages to not cleanly decouple structure from serialization format, there is no way to simply say "serializable".  But conversion to/from integer or string is well defined across languages and formats.

doc/dev/adr/0023-property-support.md
jrdouceur: Given the flexibility that is currently proposed regarding status values, I think the application is free to imply any choice of default value to non-present values.

doc/dev/adr/0023-property-support.md
```diff
@@ -0,0 +1,312 @@
...
+  * Each field may have an arbitrary type.
+  * Each field has a value that is optional/nullable.
```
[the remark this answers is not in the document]
vaavva: I don't think there's anything additional that we'd need to do in the SDK to support this, as the details of the field value type would be within the application logic. This is more of a comment about including this in documentation so that customers don't get confused about why they can't clear an Optional value (although I suppose that might only be the case if they follow our guidelines - they could easily have each write request be a complete PUT instead of a PATCH like we describe here and then not need the additional Option)

doc/dev/adr/0023-property-support.md
```diff
@@ -0,0 +1,312 @@
...
+* `TCtrl` &mdash; a structure type whose field names match those in `TProp`
+  * Each field has type Boolean.
+* `TStat` &mdash; a structure type whose field names match those in `TProp`
+  * Each field has type `TStatVal`, which is &mdash; consistently across all fields &mdash; one of Boolean, integer, string, or an enumeration that is convertable to/from integer or string.
+  * Each field has a value that is optional/nullable.
+
+Constraints for dynamically itemized properties:
+
+* `TProp` &mdash; A map/dictionary type, or a structure with a single field that is a map/dictionary type
+  * The key type is string.
```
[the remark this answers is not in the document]
jrdouceur: Strictly speaking, I think string should be fully sufficient for Property,  What it is not sufficient for is the generalization of this functionality to use cases beyond Property, such as ADR.

doc/dev/adr/0023-property-support.md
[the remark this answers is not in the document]
vaavva: In Rust, we actually implement this independent of using serde or not (and I believe Go switched to this model at some point as well). But as long as you would be fine with us implementing this in Rust in this way, then I don't have concerns

doc/dev/adr/0023-property-support.md
jrdouceur: This was one of the considerations that led Alex to ask for error headers in Command responses.  He wanted to be able to return a payload and also communicate that the payload might not be exactly as requested for some reason or other.  I think the error header mechanism could provide some support for this case.  I worry that generalizing to handle per-Property success values and concurrent per-Property failure statuses could incur a risk of complicating the common case.

doc/dev/adr/0023-property-support.md
jrdouceur: > I would propose that this is functionality that we provide in our SDK

I don't disagree, but I think it is a wholly separable component, maybe a helper class/module that can be used by services that want this functionality.  I don't know if we want to bundle it into our current dev schedule, though.  This is not my call, of course.

> I still feel that supporting non-consumer specific notifications is a very concerning/confusing scenario

It may be, but it is in fact the semantics of OPC UA, so we cannot avoid supporting it (as you noted).

doc/dev/adr/0023-property-support.md
timtay-microsoft: It's a minor thing and I don't feel too strongly about it all. I was hesitant to even leave this comment since I can't think of a definitively better word for this "consumer" concept

doc/dev/adr/0023-property-support.md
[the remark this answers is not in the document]
jrdouceur: I don't think this is really enforceable in any programming language I'm familiar with.  So, I think this could be considered to be a usage recommendation.  It's a pretty strong recommendation, though, if anyone wants to switch to modeled Interfaces at some point, because the ProtocolCompiler will not be able to generate compatible data structures that do not follow these rules.

doc/dev/adr/0023-property-support.md
jrdouceur: In fact, static and dynamic Properties *should* be generated together.  This was a weakness of the ADR that I have fixed with the latest commit.

doc/dev/adr/0023-property-support.md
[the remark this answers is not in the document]
jrdouceur: As a purely academic point, I think this actually can be enforceable in languages that allow metaprogramming.  It could probably be enforced in C++, and perhaps also in Rust using the macro language.  However, it is not enforceable in general.

PR #924, opened 2025-07-10
[Rust][Connector] Removed deprecated 3P metrics artifacts

cartertinney:
Meso is now responsible for 1P and 3P, we no longer need to send it to multiple places


rust/azure_iot_operations_connector/src/deployment_artifacts/connector.rs
```diff
@@ -86,11 +86,6 @@ pub struct ConnectorArtifacts {
     pub http_log_endpoint: Option<String>,
     /// OTEL http/https trace endpoint.
     pub http_trace_endpoint: Option<String>,
-
-    /// OTEL 3P metric endpoint.
-    pub metric_endpoint_3p: Option<String>,
```
vaavva, line 91: github won't let me comment in the right place, but should these still have 1p qualifiers?
`grpc_metric_collector_1p_ca_mount`, etc

vaavva (approved):

rust/azure_iot_operations_connector/src/deployment_artifacts/connector.rs
cartertinney: Good question for Akri folks that I intend to ask tomorrow

vaavva: in that case, let's stay consistent with those for now

PR #925, opened 2025-07-10
[Rust][Connector] Add tokio rwlock to ensure all dataset status updates reach service

vaavva:
This PR changes the sync rwlock to a tokio rwlock and holds the lock for longer when writing to protect against multiple dataset clients updating the asset status concurrently. This ensures only one status can be modified and reported at a time so that all status updates get sent to the service. (the service only accepts a full status replace, they don't do any patch handling of status updates). This may add some latency around reporting statuses as they can't happen concurrently, but this is better than losing status information. Changes in this PR are mostly restructuring to allow for this mutex to be held in the correct places

Note that this same change does not need to be made on the device client, because the device status can only be reported from the device (unlike the asset status, which can be reported from multiple dataset clients)

### DIFF
--- a/rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
+++ b/rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
@@ -791,8 +791,14 @@ impl AssetClient {
     ) -> Result<(), azure_device_registry::Error> {
         let mut status_write_guard = self.status.write().await;
         let version = self.specification.read().unwrap().version;
-        // TODO: remove this fn call and just have the functionality here
-        let new_status = Self::internal_asset_status(&status_write_guard, status, version);
+        // get current or cleared (if it's out of date) asset status as our base to modify only what we're explicitly trying to set
+        let mut new_status = Self::current_status_to_modify(&status_write_guard, version);
+        // no matter whether we kept other fields or not, we will always fully replace the config status
+        new_status.config = Some(azure_device_registry::ConfigStatus {
+            version,
+            error: status.err(),
+            last_transition_time: Some(chrono::Utc::now()),
+        });
 
         log::debug!("Reporting asset status from app for {:?}", self.asset_ref);
         // send status update to the service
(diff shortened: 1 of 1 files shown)


rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
```diff
@@ -940,12 +1013,9 @@ impl AssetClient {
     /// Note that this is the value of the last reported status (or the status
     /// from when the asset was first received if one hasn't been reported yet),
     /// so it is possible that this is not the latest value that the ADR service has
-    ///
-    /// # Panics
-    /// if the status mutex has been poisoned, which should not be possible
     #[must_use]
-    pub fn status(&self) -> adr_models::AssetStatus {
-        (*self.status.read().unwrap()).clone()
+    pub async fn status(&self) -> adr_models::AssetStatus {
```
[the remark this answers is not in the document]
vaavva: yeah, don't love it either. Right now, we don't have a specific use of this in mind for the applications, but want to be transparent and give them the option to have it if they want it for some reason. I'm open to the blocking wait, but I do feel like it could cause large delays in an async application in a not super transparent way

rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
```diff
@@ -706,31 +668,106 @@ impl AssetClient {
...
-            .await;
+            .await
+            {
+                log::error!(
+                    "Failed to report error Asset status for new Asset {:?}: {e}",
+                    asset_client.asset_ref
+                );
+            }
+        }
+        drop(status_write_guard);
```
RLeclair, line 770: Does this need to be dropped here if we're exiting the function?

rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
```diff
@@ -1572,6 +1494,37 @@ impl DatasetClient {
...
+        dataset_name: &str,
+        dataset_status: Result<(), AdrConfigError>,
+    ) {
+        if let Some(curr_dataset_status) =
+            asset_status_to_update
+                .datasets
+                .as_mut()
+                .and_then(|datasets| {
+                    datasets
+                        .iter_mut()
```
RLeclair, line 1510: Does this need to be mutable?

RLeclair (approved):

rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
vaavva: yeah, it's because the mutex lives on the AssetClient, so it's a little weird

rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
vaavva: yep - because we then set `curr_dataset_status` that we get from it's error

rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
```diff
@@ -706,31 +668,106 @@ impl AssetClient {
             asset_deletion_token,
         };
 
-        // if there are any config errors when creating the datasets, collect them all so we can report them at once
-        let mut dataset_config_errors = Vec::new();
+        // lock the status write guard so that no other threads can modify the status while we update it
+        // (not possible in new, but allows use of Self:: helper fns)
+        let mut status_write_guard = asset_client.status.write().await;
```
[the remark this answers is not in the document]
vaavva: switched to scope

PR #926, opened 2025-07-11
[Rust][Connector] Making receiving of Asset / Dataset notifications cancellable

RLeclair:
This PR:
- Addresses a bug where receiving a notification on an asset or dataset would not get fully processed if the receive flow was cancelled.
- Example: If receiving in a select branch and while processing the asset update a different branch takes over, we will lose the update.
- Changes unobserve to happen in a separate task.

### DIFF
--- a/rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
+++ b/rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
@@ -831,7 +831,8 @@ impl AssetClient {
         // lock the status write guard so that no other threads can modify the status while we update it
         let mut status_write_guard = self.status.write().await;
         // if there are any config errors when parsing the asset, collect them all so we can report them at once
-        let mut new_status: adr_models::AssetStatus = Self::current_status_to_modify(&status_write_guard, updated_asset.version);
+        let mut new_status: adr_models::AssetStatus =
+            Self::current_status_to_modify(&status_write_guard, updated_asset.version);
         let mut status_updated = false;
 
         // update datasets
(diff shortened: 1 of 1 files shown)


rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
```diff
@@ -795,132 +920,64 @@ impl AssetClient {
...
-                        &default_dataset_destinations,
-                    ) {
-                        // If an error is returned, continue to process other datasets even if one isn't valid. Don't give this one to
-                        // the application since we can't forward data on it. If there's an update to the
-                        // definition, they'll get the create notification for it at that point if it's valid
-                        dataset_config_errors.push(e);
-                    }
+                    );
+                    // Asset update has been fully processed, mark as seen.
+                    self.asset_update_watcher_rx.mark_unchanged();
```
vaavva, line 956: no update would have been put into the channel in this flow - is this mark_unchanged() needed? Or is it just for cleanup safety so that we always continue to return deleted after this point? In that case, would it be better to close the watcher_tx channel?

rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
```diff
@@ -795,132 +920,64 @@ impl AssetClient {
...
-                *unlocked_specification = AssetSpecification::from(updated_asset);
-
-                ClientNotification::Updated
             },
+            // An error here is not possible since the asset client holds the sender end
+            Ok(()) = self.asset_update_watcher_rx.changed() => {
+                // We need to mark as changed in case `handle_update` is cancelled so we can enter
+                // this branch again
+                self.asset_update_watcher_rx.mark_changed();
+                let updated_asset = self.asset_update_watcher_rx.borrow().clone();
```
vaavva, line 965: annoying that this adds 3 extra clones on the updated asset (I'd expect minimum 1 extra to save in the channel), but I don't think there's anything we can do to reduce that, right?

rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
RLeclair: It is the first case, I can't close it without dropping the channel though

rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
RLeclair: For the three cases:
- When sending (`self.asset_update_observation.recv_notification()` branch) we have to clone, it is necessary to save the value as we received it. I could avoid the clone by sending inside of `handle_update` but I would have to send the modified `updated_asset` in there.
- In (`self.asset_update_watcher_rx.changed()` here) I have to clone it because of a similar reason to the above, the updated_asset gets modified and I can't do operations on it on a reference.
- The third case is to save initially correct? I have to do a copy there as well because of the `from` on line 664. (`let specification = AssetSpecification::from(asset);`)

PR #927, opened 2025-07-11
[Rust][Connector] Make `BaseConnector` ingest `ConnectorArtifacts`

cartertinney:
Partial solution to the problem of artifact portability.

May need to combine connector arguments into some kind of context or options, depending on the eventual OTEL metrics integration.

Also still need to unify ADR artifacts with everything else.

### DIFF
--- a/rust/azure_iot_operations_connector/src/base_connector.rs
+++ b/rust/azure_iot_operations_connector/src/base_connector.rs
@@ -65,7 +65,7 @@ impl BaseConnector {
             .to_mqtt_connection_settings("0")
             .map_err(|e| e.to_string())?;
         let session_options = SessionOptionsBuilder::default()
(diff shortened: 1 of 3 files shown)


rust/azure_iot_operations_connector/src/base_connector.rs
```diff
@@ -58,66 +57,58 @@ impl BaseConnector {
     /// with exponential backoff. This allows for new configuration to be deployed to fix any
     /// errors without needing to restart the connector.
     #[must_use]
-    pub fn new(application_context: ApplicationContext) -> Self {
+    pub fn new(
+        connector_artifacts: ConnectorArtifacts,
+        application_context: ApplicationContext,
```
clecompt-msft, line 62: nit: "context"-like params typically go first in the argument list.

vaavva (approved):

PR #928, opened 2025-07-11
[Rust][Connector] Allowing underscores to be used in inbound endpoint names

RLeclair:
Allows underscores to be used in inbound endpoint names

Tested with rest connector.

### DIFF
--- a/rust/azure_iot_operations_connector/src/deployment_artifacts/azure_device_registry.rs
+++ b/rust/azure_iot_operations_connector/src/deployment_artifacts/azure_device_registry.rs
@@ -437,12 +437,10 @@ impl TryFrom<String> for DeviceEndpointRef {
 
         // TODO: Add a warning in case the format is not as expected
         match value.split_once('_') {
-            Some((device_name, inbound_endpoint_name)) => {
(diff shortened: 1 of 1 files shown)


rust/azure_iot_operations_connector/src/deployment_artifacts/azure_device_registry.rs
```diff
@@ -432,20 +432,11 @@ impl TryFrom<String> for DeviceEndpointRef {
 
     fn try_from(value: String) -> Result<Self, Self::Error> {
         // The below assumes the format is always {device_name}_{inbound_endpoint_name} with no additional
-        // `_` in the names.
-        // Kubernetes does not allow `_` in the names so this should be safe.
+        // `_` in the device name. The inbound endpoint name may have additional `_`.
```
vaavva, line 435:
```suggestion
        // `_` in the device name. The inbound endpoint name may have additional `_`.
        // Kubernetes does not allow `_` in the device name so this should be safe.
```
seems still worth calling out

vaavva (approved):

cartertinney (approved):

PR #929, opened 2025-07-11
[Rust][Connector] Cancel Safe Status Reporting

vaavva:
This PR ensures that statuses that need to be reported in recv_notification functions are cancel safe. For the Asset recv_notification function, this means that we don't do any mutable actions until we can no longer be cancelled, and then do the mutable actions at that point. For the dataset recv_notification function, we spawn reporting the status into its own task so that it doesn't introduce a cancellation point.

### DIFF
--- a/rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
+++ b/rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
@@ -1320,25 +1320,51 @@ impl DatasetClient {
         &self,
         status: Result<(), AdrConfigError>,
     ) -> Result<(), azure_device_registry::Error> {
+        log::debug!("Reporting dataset {:?} status from app", self.dataset_ref);
+        Self::internal_report_status(
+            &self.asset_status,
+            &self.asset_specification,
+            status,
+            &self.dataset_ref,
+            &self.connector_context,
+            &self.asset_ref,
+            "DatasetClient::report_status",
+        )
+        .await
+    }
+
+    async fn internal_report_status(
+        asset_status_mutex: &Arc<tokio::sync::RwLock<adr_models::AssetStatus>>,
+        asset_specification: &Arc<std::sync::RwLock<AssetSpecification>>,
+        desired_dataset_status: Result<(), AdrConfigError>,
+        dataset_ref: &DatasetRef,
+        connector_context: &Arc<ConnectorContext>,
+        asset_ref: &AssetRef,
+        log_identifier: &str,
+    ) -> Result<(), azure_device_registry::Error> {
         // get current or cleared (if it's out of date) asset status as our base to modify only what we're explicitly trying to set
-        let mut status_write_guard = self.asset_status.write().await;
+        let mut status_write_guard = asset_status_mutex.write().await;
         let mut new_status = AssetClient::current_status_to_modify(
             &status_write_guard,
-            self.asset_specification.read().unwrap().version,
+            asset_specification.read().unwrap().version,
         );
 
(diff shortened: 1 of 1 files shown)


rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
```diff
@@ -981,6 +984,26 @@ impl AssetClient {
         let mut unlocked_specification = self.specification.write().unwrap(); // unwrap can't fail unless lock is poisoned
         *unlocked_specification = AssetSpecification::from(updated_asset);
 
+        // update dataset_hashmap now that this task can't be cancelled
+        self.dataset_hashmap = temp_dataset_hashmap;
```
cartertinney, line 988: Is there a risk of a race condition overwrite here? this doens't look mutexted

cartertinney (approved):

rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
```diff
@@ -835,10 +835,12 @@ impl AssetClient {
             Self::current_status_to_modify(&status_write_guard, updated_asset.version);
         let mut status_updated = false;
 
+        // make changes to a copy of the dataset_hashmap so that this function is cancel safe
+        let mut temp_dataset_hashmap = self.dataset_hashmap.clone();
         // update datasets
         // remove the datasets that are no longer present in the new asset definition.
         // This triggers deletion notification since this drops the update sender.
-        self.dataset_hashmap.retain(|dataset_name, _| {
+        temp_dataset_hashmap.retain(|dataset_name, _| {
```
RLeclair, line 843: Just to confirm here, let's say we have dataset 1 here that will not be retained. I was thinking of a scenario as follows:
1. On creating a temporary dataset we now have two deletion tokens for dataset 1
2. When we retain on the temporary dataset we drop the copied deletion token which triggers a deletion on the dataset
3. Say we get cancelled 
4. On the second run we will once again create a temporary dataset for dataset 1 with two deletion tokens. 

On step 4, can we create two deletion tokens after one has been cancelled (I believe we use a drop guard so we lose it on step 2)?

rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
```diff
@@ -928,16 +931,16 @@ impl AssetClient {
...
                             (
                                 received_dataset_definition.clone(),
                                 dataset_update_watcher_tx,
                             ),
                         );
 
-                        // error is not possible since the receiving side of the channel is owned by the AssetClient
-                        let _ = self.dataset_creation_tx.send(new_dataset_client);
+                        // save new dataset client to be sent on self.dataset_creation_tx after the task can't get cancelled
+                        new_datasets.push(new_dataset_client);
```
RLeclair, line 943: Nit
```suggestion
                        new_dataset_clients.push(new_dataset_client);
```

RLeclair (approved):

rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
vaavva: it's not a drop guard, but the sending side of the watch channel, so the temp dataset hashmap creates a copy of the sending side of the watch channel, so because the original still has a copy of the watch channel, it doesn't go out of scope and trigger deletion until the watch channel goes out of scope in both the real dataset_hashmap (right before the update is returned to the application) as well as the temp_dataset (in line 843)

PR #931, opened 2025-07-14
[Rust] Clean up Redundant Clones

vaavva:
Did a quick pass with [Clippy Lint Redundant Clone](https://rust-lang.github.io/rust-clippy/master/#redundant_clone) set to warn and fixed all of the correct warnings that came up. We can't leave this on by default because it isn't super stable and detects some needed clones, but can be used as a local tool to help with reducing extra `clone()`s. Note that this also doesn't catch everything, so there may be more that could be cleaned up, and this still needs to be manually thought about, even if this warning is on

### DIFF
--- a/rust/azure_iot_operations_mqtt/src/session/receiver.rs
+++ b/rust/azure_iot_operations_mqtt/src/session/receiver.rs
@@ -1134,7 +1134,7 @@ mod tests {
         ); // Type 4
 
         // Register a new filter again
-        let topic_filter8 = topic_filter7; // Type 4
+        let topic_filter8 = topic_filter7.clone(); // Type 4
         let filter_rx8 = manager
             .lock()
             .unwrap()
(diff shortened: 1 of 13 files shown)


rust/azure_iot_operations_mqtt/src/session/receiver.rs
```diff
@@ -1134,7 +1134,7 @@ mod tests {
         ); // Type 4
 
         // Register a new filter again
-        let topic_filter8 = topic_filter7.clone(); // Type 4
```
cartertinney, line 1137: I would prefer to exclude this change for clarity reasons - I understand it is technically unnecessary, but having this clone reduces cognitive overhead in test maintenance.

vaavva: there were a few others in this file that clippy detected that needed to stay as well - I'll revert this one then!

cartertinney (approved):

PR #934, opened 2025-07-15
[Go] Support the aio-persistence flag

clecompt-msft:


go/services/statestore/options.go
```diff
@@ -35,6 +35,9 @@ type (
 	// WithManualAck allows notifications to be manually acknowledged.
 	WithManualAck bool
 
+	// WithPersist indicates that the key should be stored to disk.
+	WithPersist bool
```
timtay-microsoft, line 39: I'm unfamiliar with Go, but is the default value here false?

clecompt-msft: Correct.

avishekpant (approved):

go/protocol/telemetry_sender.go
```diff
@@ -139,7 +145,10 @@ func (ts *TelemetrySender[T]) Send(
 	if err := opts.CloudEvent.toMessage(pub); err != nil {
 		return err
 	}
-	pub.Retain = opts.Retain
+	pub.Retain = opts.Retain || opts.Persist
```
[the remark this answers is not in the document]
vaavva: sync'd offline, this makes total sense

PR #935, opened 2025-07-16
Add environment variables for Azure IoT Operations and resource group…

asergaz:
… creation

### DIFF
--- a/.env
+++ b/.env
@@ -11,27 +11,6 @@ export -n AIO_TLS_CA_FILE
 export -n AIO_SAT_FILE
 export -n AIO_TLS_CERT_FILE
 export -n AIO_TLS_KEY_FILE
-export -n LOCATION
-export -n RESOURCE_GROUP
-export -n CLUSTER_NAME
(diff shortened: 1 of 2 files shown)


.env
```diff
@@ -11,6 +11,27 @@ export -n AIO_TLS_CA_FILE
...
+export -n RESOURCE_GROUP
+export -n CLUSTER_NAME
+export -n STORAGE_ACCOUNT
+export -n SCHEMA_REGISTRY
+export -n SCHEMA_REGISTRY_NAMESPACE
+
+# Environment variables used by install-aio-sdk.sh
+
+# LOCATION: An Azure region close to you. For the list of currently Azure IoT Operations supported regions, see https://learn.microsoft.com/azure/iot-operations/overview-iot-operations#supported-regions.
+export LOCATION=westus
```
dominicbetts, line 24: Probably best to use values that are obviously placeholders. Storage account names must be unique, so it'll fail if more than one person uses myiotopsstorage. This may be the case for the cluster name and schema registry as well.

PR #941, opened 2025-07-17
[Rust] Add report_message_schema_reference

clecompt-msft:
This change factors out the schema reference reporting from the schema registration, to support scenarios where a dataset's schema has already been registered externally.


vaavva (approved):
LGTM

PR #952, opened 2025-07-22
[ADR] RPC streaming design doc

timtay-microsoft:
TODO: https://github.com/Azure/iot-operations-sdks/pull/1107#discussion_r2577922091


doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,163 @@
...
+ - RPC executor receives command request with "__streamResp", but that executor doesn't understand streaming requests because it uses an older protocol version
+   - Command executor responds with "not supported protocol" error code
+ - RPC executor receives command request with "__streamResp", and the executor understands that it is a streaming request (protocol versions align) but that particular command doesn't support streaming
+   - RPC executor treats it like a non-streaming command, but adds the "__isLastResp" flag to the one and only response
+ - RPC invoker tries to invoke a non-streaming command that the executor requires streaming on
+   - Atypical case since codegen will prevent this
+   - But, for the sake of non-codegen users, a new error code "StreamingRequired" would be returned by the executor
+     - Or should this just be "invalid header" error since the executor expects the "__streamResp" header?
+ - timeout per response vs overall?
+
```
vaavva, line 155: a thought that I had - should there be support for the invoker to stop receiving more stream messages for a given command? Obviously, it could just stop listening, but the executor would still keep sending them. Three scenarios here:
1. If the invoker dies, should there be some way for the executor to be able to detect this so that it doesn't keep sending more stream messages (probably not - for starters, I think this would require QoS 2, but also, if the invoker just temporarily loses network connectivity, it would want the broker to have queued up any messages that happened in the meantime. I don't think there would be a way to distinguish that from a lost session (where we could say that there's no point to sending more stream messages))
2. For something where the command request is "send me this video file", then it doesn't make much sense to decide to cancel it part way. I could see this _maybe_ being desired, especially if a stream is significantly long, but it doesn't seem like a super common scenario to me?
3. However, for a command request like "stream this security camera feed to me", I'm not sure there would ever be a "last message", so the invoker would _need_ some way to cancel this

Now, we could say that streaming only supports finite streams, and then it's less of a big issue, but I think it's worth considering if infinite streams are an expected use of the pattern.

We could also potentially say that infinite streams should actually fall into the observation pattern (proposed in @jrdouceur's property envoy ADR), but the read/write portions of that pattern don't make sense for infinite streaming, so that would have to be considered. Either way, I think there is some potential overlap between streaming and the observation pattern, and we should definitely coordinate on whether we can easily define them as two separate patterns or whether maybe they should even be combined?
I'm slightly leaning towards there's probably some separation between the two because of the correlation id(?), and the easiest to have them be more separate in implementation to me seems streaming should be for something finite, and observations should be for infinite listening, but that would require some changes to observation semantically to support something that's less of a state change and more of a stream frame?

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -117,24 +127,47 @@ TODO which existing client works well for long-running commands? Mem mon ("Repor
...
 - The command executor receives a command **without** "__streamResp" flag set to "true"
   - The command must be responded to without streaming
 
+### Cancellation support
+
+To avoid scenarios where long-running streaming responses are no longer wanted, we will want to support cancelling RPC calls. This feature is moreso applicable for RPC streaming, but the design allows for it to work for non-streaming RPC as well.
+
+#### Invoker side
+
+- The command invoker may cancel a normal or streaming RPC call at an arbitrary time by sending an MQTT message with:
```
vaavva, line 141: For streaming, I think this will be a little bit easier because there's already the concept of some messages with pause points in between, but I worry this might be a bit more challenging for a normal RPC call. For what I'm thinking, in Rust:

For a streaming RPC call,
the invoker application would call invoke, and instead of `await`ing until the response is returned/timeout, it would get the receiving end of a channel, where it receives each message until it gets the one with the isLastResponse (at which point, maybe we also close the channel to indicate no more messages, tbd). It would be fairly trivial to also return some one-shot channel or cancellation token for the application to indicate to the envoy to cancel the request at any point. We could also have the application dropping/closing the receiving end of the channel automatically indicate cancellation (not sure if we'd want to be more explicit though)

For the executor side (still for a streaming rpc call), on recv(), the executor would receive the request and the sending side of a channel to send all of the responses. If we receive the cancelRpc message from the invoker, it would be easy enough to close the envoy's receiving end of that channel to indicate not to send any more messages (and if we needed that notice to come sooner than the next time they send the next message, we could consider providing some cancellation token or something as well, although that would require more explicit monitoring of something extra, so I'd initially lean against that)

However, for normal RPC,
The invoker in Rust calls invoke() and then this function doesn't return until the response is received. There isn't a time for us to give the invoker a cancellation token, so we'd either have to have them pass in a cancellation token themselves or have some way to call a different fn with the correlation id to cancel - however, both of these would have to happen _while_ they're `await`ing the original invoke, so it's a bit weird for them to still be listening on the `await` and also sending a cancellation? In my opinion, this is what the timeout is for and introducing another way to cancel these requests seems a bit odd (granted, I understand this may be a rust specific oddity). Another option that I don't think would work - I'm not sure there's a way for us to detect that the application stopped listening on the invoke.await reliably to use that along.

On the executor side, kind of similar but not as weird - the application calls recv() and gets the request info as well as a struct that has a function to call to send the response. We could provide some cancellation token with the request info for the application to be monitoring. However, the expectation would be that the application is doing some execution without pauses until it's ready to respond, so even if we provided a cancellation token, there potentially aren't as many points where it pauses execution (like in streaming where it would have to "pause" to send the part of the response) where it would be checking for this cancellation.


Summary, I think it _could_ be possible to support this for normal RPC calls, but if isn't as useful, I'm not sure the extra complexities of adding it would be worth it. The streaming scenario seems much more likely to desire this

clecompt-msft:
Recording meeting notes:
* Agreed that response ID is an application-level concept that should be included in the payload or user properties rather than as part of this protocol.
* Agreed that from a wire-protocol perspective we will not require ordering (and therefore still need indexes). In the future we could consider (optionally) trying to enforce ordering on the client side, but that does not seem to be an initial requirement.
* Need a section covering message expiry / timeouts / acks and how these all interplay with the various features (with a callout for cancellation specifically here).
* Leaning towards having completely separate invoker/executor types and taking a strong stance that a single command is strongly tied to a single type.
* Request to combine metadata into a single header.
* Leaning towards caching not being viable, due to the existence of long-running (and potentially infinite) streams. Since this violates some assumptions/guarantees of what we consider commands, we may want to treat streaming as a separate concept (which leans in further to having them be separate types).

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,282 @@
...
+    where TReq : class
+    where TResp : class
+{
+    /// <summary>
+    /// A streaming command was invoked
+    /// </summary>
+    /// <remarks>
+    /// The callback provides the stream of requests and requires the user to return one to many responses.
+    /// </remarks>
+    public required Func<IAsyncEnumerable<StreamingExtendedRequest<TReq>>, CancellationToken, IAsyncEnumerable<StreamingExtendedResponse<TResp>>> OnStreamingCommandReceived { get; set; }
```
varunpuranik, line 93: How will the SDK add the header to the response messages?

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,282 @@
...
+  - Even though the streaming command classes are separate from the existing RPC classes, they should also offer the same features around topic string pre/suffixing, custom topic token support, etc.
+
+Upon receiving a MQTT message that contains a streaming request, the streaming executor should notify the application layer that the first message in a request stream was received. Once the executor has notified the user that the final message in a request stream was received, the user should be able to provide a stream of responses. Upon receiving each response in that stream from the user, the executor will send an MQTT message for each streamed response with:
+  - The same correlation data as the original request
+  - The topic as specified by the original request's response topic field
+  - The appropriate streaming metadata [see above](#streaming-user-property)
+  - The serialized payload as provided by the user's response object
+  - Any user-definied metadata as specified in the ```ExtendedResponse```
+
+Unlike normal RPC, the streaming command executor will not provide any cache support. This is because streams may grow indefinitely in length and size.
```
jrdouceur, line 157: In normal RPC, caching of non-idempotent requests is considered [necessary for correctness](https://github.com/Azure/iot-operations-sdks/blob/main/doc/reference/command-cache.md#cache-behavior-for-rpc-correctness).  Is there a reason this is not the case for streaming RPC?  Are we assuming that streaming requests are always idempotent?

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,282 @@
...
+    /// </summary>
+    /// <remarks>
+    /// The callback provides the stream of requests and requires the user to return one to many responses.
+    /// </remarks>
+    public required Func<IAsyncEnumerable<StreamingExtendedRequest<TReq>>, CancellationToken, IAsyncEnumerable<StreamingExtendedResponse<TResp>>> OnStreamingCommandReceived { get; set; }
+}
+
+```
+
+With this design, commands that use streaming are defined at codegen time. Codegen layer changes will be defined in a separate ADR, though.
```
jrdouceur, line 98: The ProtocolCompiler will certainly need substantial additions to handle this design, but I think the DTDL additions could be nearly trivial.  The structural similarities are so strong that I think the current DTDL Command would be sufficient.  We'd just need to add a simple co-type, such as Streaming, to indicate that the codegen target is the SDK's streaming RPC API rather than the normal RPC API.

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,282 @@
...
+  - The serialized payload as provided by the user's request object
+  - Any user-definied metadata as specified in the ```ExtendedRequest```
+
+Once the stream of requests has finished sending, the streaming command invoker should expect the stream of responses to arrive on the provided response topic with the provided correlation data and the streaming user property.
+
+The command invoker will acknowledge all messages it receives that match the correlation data of a known streaming command
+
+#### Executor side
+
+A streaming command executor should start by subscribing to the expected command topic
```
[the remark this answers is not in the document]
timtay-microsoft: RPC streaming will utilize shared subscriptions the same way that vanilla RPC does. I can make this explicit in the doc for clarity

doc/dev/adr/0025-rpc-streaming.md
timtay-microsoft: I re-worked this a bit so that users can pass along custom user properties per message as part of the returned IAsyncEnumerable. 

The SDK will iterate on that enumerable internally and, when the iterator offers a new response, the SDK will construct the MQTT message with both the user-defined properties and the standards headers (__stream header)

doc/dev/adr/0025-rpc-streaming.md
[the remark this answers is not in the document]
timtay-microsoft: Good point. I hadn't considered request messages being delivered to different executors as the request stream flows. 

I think it would simplify the design a lot if we imposed a constraint here saying that all requests in a stream must go to the same executor (and all the responses from that executor go back to the initial invoker). Allowing other executors to take over a stream hasn't been a feature ask AFAIK.

To achieve this, it sounds like we should borrow the executorId concept used in vanilla RPC, but make it mandatory for the invoker to specify the executorId at invoke time and make it mandatory that the executor side's executorId is its client Id. That should make it pretty ironclad that the invoker sends all the requests to the same executor. This would also mean not using shared subscriptions since each executor will have a unique topic anyways.

I'll update the doc

doc/dev/adr/0025-rpc-streaming.md
timtay-microsoft: It's more-so that implementing those same guarantees in RPC streaming is less feasible and maybe less relevant.

Less feasible because:
- Re-playing a stream response may require storing an unboundedly large set of response payloads
- We allow executors to send a response upon receiving a request (as opposed to waiting for all requests before sending any responses). This complicates if/when we send cached responses. 
  - For example, if an executor receives a new stream where the first 3 request messages are equivalent to a stream in cache, but the 4th message deviates. If we already sent cached responses for the first 3 requests, do we ask the user to provide the next response?

Less relevant because:
- Users that want to report the same payload shape over time are more likely to be reporting the status of something, so a cached response may become outdated quickly
- A user may be less interested in receiving all of the responses from a cache in the first place (especially in status reporting streams)

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,399 @@
...
+
+    /// <summary>
+    /// Get the user properties associated with a cancellation request started with <see cref="CancelAsync(Dictionary{string, string}?, CancellationToken)"/>.
+    /// </summary>
+    /// <returns>The user properties associated with a cancellation request</returns>
+    /// <remarks>
+    /// If the stream has not been cancelled, this will return null. If the stream has been cancelled, but no user properties were
+    /// provided in that cancellation request, this will return null.
+    /// </remarks>
+    Dictionary<string, string>? GetCancellationRequestUserProperties();
```
vaavva, line 302: why not have these as some information provided with the cancellation? (Or is this just a dotnet limitation)

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,399 @@
...
+### Disconnection scenario considerations
+
+- Invoker side disconnects unexpectedly while sending requests
+  - Upon reconnection, the request messages queued in the session client should send as expected
+  - If no reconnection, the streaming RPC will timeout
+- Invoker side disconnects unexpectedly while receiving responses
+  - The broker should hold all published responses for as long as the invoker's session lives and send them upon reconnection
+  - If the invoker's session is lost, then the RPC will timeout
+- Executor side isn't connected when invoker sends first request
+  - Invoker will receive a "no matching subscribers" puback
```
vaavva, line 363: this is broker dependent behavior, so we can't count on receiving this puback (some brokers will just return a successful puback)

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,399 @@
...
+### Protocol layer details
+
+#### Invoker side
+
+- The command invoker may cancel a streaming command while streaming the request or receiving the stream of responses by sending an MQTT message with: 
+  - The same MQTT topic as the invoked method
+  - The same correlation data as the invoked method 
+  - Streaming metadata with the ["cancel" flag set](#streaming-user-property)
+  - No payload
+- The command invoker should still listen on the response topic for a response from the executor which may still contain a successful response (if cancellation was received after the command completed successfully) or a response signalling that cancellation succeeded ("Canceled" error code)
```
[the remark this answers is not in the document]
timtay-microsoft: I may not have made this clear enough, but there is no "success" response to a cancellation request.

The receiving end of a cancellation request (invoker or executor) can either:
  - Send the "canceled" error code 
  - Send nothing if it already sent the final message in its sending stream and already received the final message in its receiving stream ("You can't cancel this because it is already done")

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,399 @@
...
+
+```0:false:false:10000```: The first (and not last) message in a request stream where the RPC should timeout beyond 10 seconds
+
+```3:true:false```: The third and final message in a response stream
+
+```0:true:false:1000```: The first and final message in a request stream where the RPC should timeout beyond 1 second
+
+```0:true:true:0```: This request stream has been canceled. Note that the values for ```index```, ```isLast```, and ```<rpc timeout milliseconds>``` are ignored here.
+
+```0:true:true```: This response stream has been canceled. Note that the values for ```index``` and ```isLast``` are ignored here.
```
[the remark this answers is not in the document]
timtay-microsoft: In so far as they won't be presented to the user as though they were messages in the stream. Our protocol layer would take these messages and turn them into signaling a cancellation token to let the user's application layer know to stop processing this stream.

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,400 @@
...
+    /// <summary>
+    /// Invoke a streaming command on a particular streaming command executor
+    /// </summary>
+    /// <param name="requests">The stream of requests to send. This stream must contain at least one request.</param>
+    /// <param name="streamMetadata">The metadata for the request stream as a whole.</param>
+    /// <param name="additionalTopicTokenMap">Topic tokens to substitute in the request topic.</param>
+    /// <param name="streamExchangeTimeout">The timeout between the beginning of the request stream and the end of both the request and response stream.</param>
+    /// <param name="cancellationToken">Cancellation token. Signalling this will also make a single attempt to notify the executor of the cancellation.</param>
+    /// <returns>The stream of responses.</returns>
+    public async Task<IStreamContext<StreamingExtendedResponse<TResp?>> InvokeStreamingCommandAsync(
```
koepalex, line 96: Can we please use `IAsyncEnumerable<StreamingExtendedResponse<TResp?>>` as return value?

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,400 @@
...
+   - This would force users to always call the ```InvokeCommandWithStreaming``` API on the command invoker side and that returned object isn't as easy to use for single responses
+ - Treat streaming RPC as the same protocol as RPC
+   - This introduces a handful of error cases such as:
+     - Invoker invokes a method that it thinks is non-streaming, but the executor tries streaming responses
+     - Executor receives a streaming command but the user did not set the streaming command handler callback (which must be optional since not every command executor has streaming commands)
+   - API design is messy because a command invoker/executor should not expose streaming command APIs if they have no streaming commands
+   - Caching behavior of normal RPC doesn't fit well with streamed RPCs which may grow indefinitely large
+
+
+## Appendix
```
koepalex, line 389: Can each streaming message contain application level errors?

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,400 @@
...
+ - When exposed to the user, each request and response includes an index of where it was in the stream
+ - Allow for multiple separate commands to be streamed simultaneously
+ - Allow for invoker and/or executor to cancel a streamed request and/or streamed response at any time
+ - Allow for invoker + executor to send their requests/responses at arbitrary* times
+   - For instance, executor may send 1 response upon receiving 1 request, or it may wait for the request stream to finish before sending the first response
+   - Alternatively, this allows the invoker to send a request upon receiving a response
+   - *The only limitation is that the invoker must initiate the RPC streaming with a request
+ - Allow for invoker/executor to end their own request/response stream gracefully at any time
+   - For instance, if the executor doesn't know if a response will be the last one prior to sending it, the executor should still be capable of ending the response stream later without sending another fully-fledged payload
+
```
timtay-microsoft, line 21: Do we want to make it a requirement that a hypothetical chunking implementation could be built on this proposed streaming API?

doc/dev/adr/0025-rpc-streaming.md
timtay-microsoft: I believe the reason I got away from that design was because we need a way to cancel the stream using the returned type, but we will follow up on that again

doc/dev/adr/0025-rpc-streaming.md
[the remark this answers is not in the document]
timtay-microsoft: Circling back on this now, I may have misunderstood this question at the time.

From what I can recall, Valerie's point here is correct in that

```
Invoker sends cancellation with cancel flag set -> executor responds with error code cancelled
Executor sends cancellation with cancel flag set -> invoker response with error code cancelled
Executor sends cancellation with cancel flag set -> invoker response with "stream cancelled successfully" flag set
```

Should be reduced down to just

```
Invoker sends cancellation with cancel flag set -> executor responds with error code cancelled
Executor sends cancellation with cancel flag set -> invoker response with error code cancelled
```

doc/dev/adr/0025-rpc-streaming.md
vaavva: I think that would be a good goal!

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,400 @@
...
+## Non-requirements
+
+ - Different payload shapes per command response
```
vaavva, line 24:
```suggestion
 - Different payload shapes per command response/request
```

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,400 @@
...
+    /// </summary>
+    /// <param name="requests">The stream of requests to send. This stream must contain at least one request.</param>
+    /// <param name="streamMetadata">The metadata for the request stream as a whole.</param>
+    /// <param name="additionalTopicTokenMap">Topic tokens to substitute in the request topic.</param>
+    /// <param name="streamExchangeTimeout">The timeout between the beginning of the request stream and the end of both the request and response stream.</param>
+    /// <param name="cancellationToken">Cancellation token. Signalling this will also make a single attempt to notify the executor of the cancellation.</param>
+    /// <returns>The stream of responses.</returns>
+    public async Task<IStreamContext<StreamingExtendedResponse<TResp?>> InvokeStreamingCommandAsync(
+      IAsyncEnumerable<StreamingExtendedRequest<TReq>> requests,
+      StreamRequestMetadata? streamRequestMetadata = null,
```
vaavva, line 98: would this metadata (here and in the executor fn as well) be per item in the stream or only one for the entire stream?

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,400 @@
...
+
+```<uint>:<boolean>:<boolean>:<uint>```
+
+where the field ```:<rpc timeout milliseconds>``` is only present in request stream messages and may be omitted if the RPC has no timeout.
+
+For example:
+
+```0:false:false:10000```: The first (and not last) message in a request stream where the RPC should timeout beyond 10 seconds
+
+```3:true:false```: The third and final message in a response stream
```
vaavva, line 145: what happens if the executor sends a response with isLast = true, but the invoker either has another request in flight or would like to send more requests on the same stream?

vaavva: or should the executor not send isLast=true until the invoker has sent a request with isLast=true?

doc/dev/adr/0025-rpc-streaming.md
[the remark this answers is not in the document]
vaavva: would it be important to know whether this has been received before a previous response though?
ex. I receive 1:false:false then 3:true:true, then I know that I might get 2:false:false soon?

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,400 @@
...
+We need to provide timeout support for our streaming APIs to avoid scenarios such as:
+
+- The invoker side is stuck waiting for the final response in a stream because it was lost or the executor side crashed before sending it.
+- The executor side is stuck waiting for the final request in a stream because it was lost or the invoker side crashed before sending it.
+
+#### Decision
+
+We will allow configuration on the invoker's side of a timeout for the RPC as a whole and a timeout of each message in the request and/or response stream.
+
+##### RPC level timeout
```
vaavva, line 217: tiny nit to make things clearer - can we call this the Stream level timeout? I kept forgetting whether RPC was for the entire exchange or for a single RPC request

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,400 @@
...
+
+To enable this, each message in the request stream will include a value in the ```<rpc timeout milliseconds>``` portion of the ```__stream``` user property. This header should be sent in all request stream messages in case the first N request messages are lost due to timeout or otherwise.
```
vaavva, line 219: Since this timeout could get rather large for streams that are expected to be long-running, thoughts on this being in seconds to reduce bytes on the wire? And since the executor would start the timer once it receives the first message, it's not going to match on both sides to millisecond level precision anyways. And seconds would align more with the message expiry property that we use for other rpc timeouts

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,400 @@
...
+
+The invoker side will start a countdown from this value after receiving the first PUBACK that ends with throwing a timeout exception to the user if the final stream response has not been received yet. The invoker should not send any further messages beyond this timeout.
+
+The executor side will start a countdown from this value after receiving the first PUBLISH in the request stream. At the end of the countdown, if the executor has not sent the final response in the response stream, the executor should return the ```timeout``` error code back to the invoker. The executor should also notify the user callback to stop.
```
vaavva, line 223: is there any point to sending the timeout error code back to the invoker after the timer has expired? Because the executor doesn't start the timer until after it receives the first message, the timer on the executor will always expire after the invoker stops listening for responses. I do like the completeness of explicitly "closing the stream", but I'm not sure it would ever be "received"

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,400 @@
...
+
+The invoker side will start a countdown from this value after receiving the first PUBACK that ends with throwing a timeout exception to the user if the final stream response has not been received yet. The invoker should not send any further messages beyond this timeout.
+
+The executor side will start a countdown from this value after receiving the first PUBLISH in the request stream. At the end of the countdown, if the executor has not sent the final response in the response stream, the executor should return the ```timeout``` error code back to the invoker. The executor should also notify the user callback to stop. 
+
+Any request stream or response stream messages that are received by the executor/invoker after they have ended the timeout countdown should be acknowledged but otherwise ignored. This will require both parties to track correlationIds for timed out streams for a period of time beyond the expected end of the RPC so that any post-timeout messages are not treated as initiating a new stream.
```
vaavva, line 225: I think for the invoker it doesn't need to track anything other than the current correlationIds, because if it gets a response for something outside of that set (whether because the stream expired or was never tracked), it would always be (acked and) ignored.

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,400 @@
...
+
+If the request stream omits the timeout value in the ```__stream``` user property, the invoker and executor should treat the RPC as not having a timeout.
+
+This design does make the invoker start the countdown sooner than the executor, but the time difference is negligible in most circumstances.
+
+##### Message level timeout
+
+We will allow users to set the message expiry interval of each message in a request/response stream. By default, though, we will set each message expiry interval equal to the RPC level timeout value.
```
vaavva, line 233: as a side note, this behavior can inform what the cache ttl on the executor should be for a correlation id (for when it ignores requests that are known to have already expired) - it must be at least 2x the rpc level timeout plus some buffer, since the broker could redeliver a request up to that point.

vaavva, line 233: will we allow a message expiry interval > rpc level timeout? (I think we probably shouldn't allow this)

vaavva, line 233: Should the message level timeout value be used at all by the receiving invoker/executor, or is it purely used here for broker redelivery? I would assume the invoker definitely wouldn't do anything by the received value. But I would lean towards the executor not doing anything with this value either, and having it be completely just for broker redelivery, because I don't think we can generically assign meaning to it separate from the rpc level timeout - one request isn't necessarily correlated to one response, so there's nothing to say to stop doing if one message in the stream "times out", and we know the invoker is still listening for responses as a whole until the rpc level timeout

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,400 @@
...
+
+Both the invoker and executor stream messages _must_ include a message expiry interval. The receiving end will use this value as the de-dup cache length for each cached message. Vanilla RPC has the same requirement as explained [here](../../reference/command-timeouts.md#input-values).
```
vaavva, line 235: while it makes sense for the executor (and I guess the invoker now too) to have a de-dup cache for received messages (so that it doesn't send duplicates to the customer), it would be slightly different on the executor from vanilla rpc because it wouldn't resend any responses when receiving a duplicate request, right?

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,400 @@
...
+  - This is the approach that gRPC takes, but... 
+    - It doesn't account for scenarios where the invoker/executor dies unexpectedly (since gRPC relies on a direct connection between invoker and executor)
+- Use the message expiry interval of the first received message in a stream to indicate the RPC level timeout
+  - Misuses the message expiry interval's purpose and could lead to broker storing messages for extended periods of time unintentionally
+
+### Cancellation support
+
+To avoid scenarios where long-running streaming requests/responses are no longer wanted, we will want to support cancelling streaming RPC calls. 
+
+Since sending a cancellation request may fail (message expiry on broker side), the SDK API design should allow for the user to repeatedly call "cancel" and should return successfully once the other party has responded appropriately.
```
vaavva, line 252: would we want the user to have to repeatedly call "cancel", or should the SDK be the one managing sending multiple cancellation requests until a cancellation response has been received?

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,400 @@
...
+    /// </remarks>
+    Dictionary<string, string>? GetCancellationRequestUserProperties();
+
+    /// <summary>
+    /// True if this stream exchange has timed out. If a stream has timed out, <see cref="CancellationToken"/> will trigger as well.
+    /// </summary>
+    bool HasTimedOut { get; internal set; }
+
+    /// <summary>
+    /// True if this stream exchange has been canceled by the other party. If a stream has been cancelled, <see cref="CancellationToken"/> will trigger as well.
```
vaavva, line 310: tiny nit to align spelling on canceled vs cancelled at least in the code :) not sure if there's one spelling or the other that we've already been using in the SDKs to align to

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,400 @@
...
+
+[see timeout support for more details on timeout scenarios](#timeout-support)
+
+#### Invoker side
+
+The streaming command invoker will first subscribe to the appropriate response topic prior to sending any requests
+
+Once the user invokes a streaming command, the streaming command invoker will send one to many MQTT messages with:
+  - The same response topic
+    - This response topic must be prefixed with 'clients/{mqtt client id of invoker}' like in vanilla RPC
```
vaavva, line 163: I don't think this is a requirement, just the default behavior if it isn't overridden

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,400 @@
...
+  - The same correlation data
+  - The user property "$partition" set to a value of the client Id of the MQTT client sending this invocation
+    - This ensures that the broker always routes the messages in the stream to the same executor
+  - The appropriate streaming metadata [see above](#streaming-user-property)
+  - The serialized payload as provided by the user's request object
+  - Any user-definied metadata as specified in the ```ExtendedStreamingRequest```
+  - QoS 1
```
vaavva, line 170: I'm assuming we probably want the other standard rpc metadata as well, for completeness might be worth listing out here too since you are already listing some of them: content type, format indicator, message expiry, timestamp, sourceId, protocol version
https://github.com/Azure/iot-operations-sdks/blob/10f55dcab314135a35e7ef362c223d02d066b98c/doc/reference/message-metadata.md#request-message

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,400 @@
...
+
+where the field ```:<rpc timeout milliseconds>``` is only present in request stream messages and may be omitted if the RPC has no timeout.
+
+For example:
+
+```0:false:false:10000```: The first (and not last) message in a request stream where the RPC should timeout beyond 10 seconds
+
+```3:true:false```: The third and final message in a response stream
+
+```0:true:false:1000```: The first and final message in a request stream where the RPC should timeout beyond 1 second
```
vaavva, line 147: I think because of https://github.com/Azure/iot-operations-sdks/pull/952/changes#islast-message-being-its-own-message, this wouldn't be a valid user property value - so there must always be at least 2 messages for a full stream: at minimum the first message and an empty message indicating isLast

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,400 @@
...
+
+Once the user-supplied stream of request messages has ended, the streaming command invoker should send one final message to the same topic/with the same correlation data with no payload and with the 'isLast' flag set in the '__stream' metadata bundle.  
+
+Upon receiving an MQTT message in the response stream with the 'isLast' flag set in the '__stream' metadata, the streaming command invoker should notify the user that the stream of responses has ended. This particular message should not contain any payload or other user properties, so the message _should not_ be propagated to the user as if it were part of the response stream. [See here for more details on why this ```isLast``` flag is an independent message](#islast-message-being-its-own-message).
+
+By default, the streaming command invoker will acknowledge all request messages it receives as soon as they are given to the user. Users may opt into manual acknowledgements, though. Opting into manual acknowledgements allows the user time to "process" each response as necessary before forgoing re-delivery from the broker if the invoker crashes unexpectedly.
```
vaavva, line 178: if the invoker crashes unexpectedly, it won't be able to process the rest of the stream anyways since it will have lost the context of the request(s) that it sent

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,400 @@
...
+  - Even though the streaming command classes are separate from the existing RPC classes, they should also offer the same features around topic string pre/suffixing, custom topic token support, etc.
+  - The executor should use a shared subscription so that, if there are multiple executors, only one of them receives each stream
+
+Upon receiving a MQTT message that contains a streaming request, the streaming executor should notify the application layer that the first message in a request stream was received. Once the executor has notified the user that the first message in a request stream was received, the user should be able to provide a stream of responses. Upon receiving each response in that stream from the user, the executor will send an MQTT message for each streamed response with:
+  - The same correlation data as the original request
+  - The topic as specified by the original request's response topic field
+  - The appropriate streaming metadata [see above](#streaming-user-property)
+  - The serialized payload as provided by the user's response object
+  - Any user-definied metadata as specified in the ```ExtendedStreamingResponse```
+  - QoS 1
```
vaavva, line 194: similar comment as on the invoker, would be good to specify if any of the additional standard rpc response metadata is used (for example, are the error properties supported in the same way or not?) https://github.com/Azure/iot-operations-sdks/blob/10f55dcab314135a35e7ef362c223d02d066b98c/doc/reference/message-metadata.md#response-message

doc/dev/adr/0025-rpc-streaming.md
vaavva: nevermind, this is clarified in the executor side section to be the case - will close this comment!

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,400 @@
...
+
+By default, the streaming command executor will acknowledge all response messages it receives as soon as they are given to the user. Users may opt into manual acknowledgements, though. Opting into manual acknowledgements allows the user time to "process" each response as necessary before forgoing re-delivery from the broker if the executor crashes unexpectedly.
+
+Also unlike normal RPC, the streaming command executor will not provide any re-play cache support. This is because streams may grow indefinitely in length and size so re-playing a response stream isn't feasible.
+
+The streaming command executor will provide de-dupe caching of received requests to account for QoS 1 messages potentially being re-delivered. The streaming command invoker will de-dup using a the combination of the correlationId of the stream and the index of the message within that stream. The de-dup cache entries for a stream should be cleared once the stream has finished (gracefully or otherwise).
```
vaavva, line 204: is this just a copy and paste typo, or is there something from the invoker that should be relevant here?
And from the timeout section, shouldn't the cache be maintained for some time after the timeout to ignore any additional requests that come in late?
```suggestion
The streaming command executor will provide de-dupe caching of received requests to account for QoS 1 messages potentially being re-delivered. The streaming command executor will de-dup using the combination of the correlationId of the stream and the index of the message within that stream. The de-dup cache entries for a stream should be cleared once the stream has finished (gracefully or otherwise) plus some buffer if it has no rpc level timeout or after 2x the rpc timeout plus buffer to not treat post-timeout messages as initiating a new stream.
```

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,400 @@
...
+
+- Invoker side disconnects unexpectedly while sending requests
+  - Upon reconnection, the request messages queued in the session client should send as expected
+  - If no reconnection, the streaming RPC will timeout
+- Invoker side disconnects unexpectedly while receiving responses
+  - The broker should hold all published responses for as long as the invoker's session lives and send them upon reconnection
+  - If the invoker's session is lost, then the RPC will timeout
+- Executor side isn't connected when invoker sends first request
+  - Depending on broker behavior, invoker will receive a "no matching subscribers" puback
+    - Seems like a scenario we would want to retry?
```
vaavva, line 364: probably best to give back to the user to decide whether to retry or not - if things just weren't configured properly, there might never be an executor available and we might be retrying forever

doc/dev/adr/0025-rpc-streaming.md
```diff
@@ -0,0 +1,400 @@
...
+
+### Protocol versioning
+
+By maintaining RPC streaming as a separate communication pattern from normal RPC, we will need to introduce an independent protocol version for RPC streaming. It will start at ```1.0``` and should follow the same protocol versioning rules as the protocol versions used by telemetry and normal RPC.
+
+## Alternative designs considered
+
+ - Allow the command executor to decide at run time of each command if it will stream responses independent of the command invoker's request
+   - This would force users to always call the ```InvokeCommandWithStreaming``` API on the command invoker side and that returned object isn't as easy to use for single responses
+ - Treat streaming RPC as the same protocol as RPC
```
vaavva, line 381: being able to support chunking as part of regular RPC (and using streaming under the hood) might require bringing some of these alternatives back into the discussion, but let's see!

doc/dev/adr/0025-rpc-streaming.md
vaavva: related question for both invoker and executor, because I remember .net might be slightly different than rust - do the current .net invoker and executor both expose manual acking? I know rust doesn't for either

PR #967, opened 2025-07-30
[.NET] Update nullability of SR client return types

timtay-microsoft:
And update sample to show how to catch "not found" errors


dotnet/src/Azure.Iot.Operations.Services/SchemaRegistry/ISchemaRegistryClient.cs
```diff
@@ -17,7 +17,7 @@ public interface ISchemaRegistryClient : IAsyncDisposable
     /// <param name="timeout">>An optional timeout for the operation, which specifies the maximum time allowed for the request to complete.</param>
     /// <param name="cancellationToken">A token that can be used to cancel the operation before completion if needed.</param>
     /// <returns>Information about the requested schema</returns>
-    Task<SchemaInfo?> GetAsync(string schemaId, string version = "1.0.0", TimeSpan? timeout = default!, CancellationToken cancellationToken = default!);
+    Task<SchemaInfo> GetAsync(string schemaId, string version = "1.0.0", TimeSpan? timeout = default!, CancellationToken cancellationToken = default!);
```
timtay-microsoft, line 20: With the recent DTDL changes, our codegen layer now returns a non-nullable SchemaInfo here and in the put API. It used to return a nullable payload so that the "not found" case could return null, but now the codegen layer throws an exception instead of returning null

maximsemenov80 (approved):

PR #976, opened 2025-08-05
[Rust] Support AIO peristence

clecompt-msft:
This change adds support for the AIO persistence flag to the Rust SDK in the following places: session (connect), telemetry send, and state-store set. It does not add it to the generic MQTT publish, since the publish options do not currently extend the underlying MQTT options.

Minor breaking changes:
* Added a field to `SetOptions`, requiring adding `default` where it was not already in use.
* Made the fields of `SessionOptions` private, requiring the (existing) builder to be used.
* `aio_broker_features` is now an optional struct (pass `None` instead of `false` to disable).


rust/azure_iot_operations_protocol/src/telemetry/sender.rs
```diff
@@ -280,6 +284,9 @@ impl<T: PayloadSerialize> MessageBuilder<T> {
                     .validate(&serialized_payload.content_type, &cloud_event.spec_version)?;
             }
         }
+        if self.persist == Some(true) && self.retain == Some(false) {
+            return Err("Persist cannot be used without retain".to_string());
```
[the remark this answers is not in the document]
clecompt-msft: This is incorrect - the validation ensures that retain has not *explicitly been set* to false when persist is true (which is invalid and would result in non-intuitive behavior), whereas the documentation describes the behavior if only persist was explicitly set.

rust/azure_iot_operations_mqtt/src/session/wrapper.rs
```diff
@@ -48,19 +48,31 @@ pub struct SessionPubReceiver(managed_client::SessionPubReceiver);
...
+    reconnect_policy: Box<dyn ReconnectPolicy>,
     /// Maximum number of queued outgoing messages not yet accepted by the MQTT Session
     #[builder(default = "100")]
-    pub outgoing_max: usize,
+    outgoing_max: usize,
     /// Indicates if the Session should use features specific for use with the AIO MQTT Broker
     #[builder(default = "true")]
-    pub aio_broker_features: bool,
+    aio_broker_features: bool,
+    /// Indicates if the Session should use AIO persistence
```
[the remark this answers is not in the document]
clecompt-msft: Implied by the AIO descriptor.

rust/azure_iot_operations_services/tests/state_store_client_test.rs
```diff
@@ -408,6 +408,7 @@ async fn state_store_set_conditions_network_tests() {
                     SetOptions {
                         expires: Some(Duration::from_secs(10)),
                         set_condition: SetCondition::OnlyIfDoesNotExist,
+                        ..Default::default()
```
vaavva, line 411: would there be any test we can add for the new field? I don't think so, but figured I'd check

rust/azure_iot_operations_services/examples/lock_client.rs
```diff
@@ -194,6 +195,7 @@ async fn lock_client_2_operations(
     let shared_resource_key_set_options = SetOptions {
         set_condition: SetCondition::Unconditional,
         expires: Some(Duration::from_secs(15)),
+        ..Default::default()
```
vaavva, line 198: this would be relevant the same way for lease/lock, right? There isn't any reason that we'd obfuscate this at this layer?

rust/azure_iot_operations_mqtt/src/session/wrapper.rs
```diff
@@ -74,11 +86,14 @@ impl Session {
 
         // Add AIO metric to user properties when using AIO MQTT broker features
         // TODO: consider user properties from being supported on SessionOptions or ConnectionSettings
-        let user_properties = if options.aio_broker_features {
-            vec![("metriccategory".into(), "aiosdk-rust".into())]
+        let mut user_properties = if options.aio_broker_features {
+            vec![("metriccategory".to_string(), "aiosdk-rust".to_string())]
```
vaavva, line 90: could you move
```
if options.persist {
            user_properties.push(("aio-persistence".to_string(), true.to_string()));
        }
```
into this if block since we won't set aio-persistence if aio_broker_features is false? Save the tiniest amount of processing in the other case :D

vaavva (approved):
Does this need to be exposed as an option for connector destinations? (probably out of scope for this pr and more of a design discussion)

rust/azure_iot_operations_mqtt/src/session/wrapper.rs
```diff
@@ -48,19 +48,31 @@ pub struct SessionPubReceiver(managed_client::SessionPubReceiver);
...
-    pub aio_broker_features: bool,
+    aio_broker_features: bool,
+    /// Indicates if the Session should use AIO persistence
+    #[builder(default = "false")]
+    aio_persistence: bool,
+}
+
+impl SessionOptionsBuilder {
+    fn validate(&self) -> Result<(), String> {
+        if self.aio_persistence == Some(true) && self.aio_broker_features == Some(false) {
```
cartertinney, line 71: This works, but perhaps broker_features should be it's own struct. Before, there was just the one broker feature, now perhaps there should be an optional struct. Would reduce validation and all these knock-on effects of it.

rust/azure_iot_operations_mqtt/src/session/wrapper.rs
```diff
@@ -51,16 +51,24 @@ pub struct SessionPubReceiver(managed_client::SessionPubReceiver);
 #[builder(pattern = "owned")]
 pub struct SessionOptions {
     /// MQTT Connection Settings for configuring the [`Session`]
-    pub connection_settings: MqttConnectionSettings,
+    connection_settings: MqttConnectionSettings,
```
cartertinney, line 54: Why did we get rid of these `pub`s? Not that i'm strictly against it, but why?

rust/azure_iot_operations_mqtt/src/session/wrapper.rs
```diff
@@ -72,10 +80,15 @@ impl Session {
...
-        // Add AIO metric to user properties when using AIO MQTT broker features
+        // Add AIO metric and features to user properties when using AIO MQTT broker features
         // TODO: consider user properties from being supported on SessionOptions or ConnectionSettings
-        let user_properties = if options.aio_broker_features {
-            vec![("metriccategory".into(), "aiosdk-rust".into())]
+        let user_properties = if let Some(features) = options.aio_broker_features {
+            let mut user_properties =
+                vec![("metriccategory".to_string(), "aiosdk-rust".to_string())];
+            if features.persistence {
+                user_properties.push(("aio-persistence".to_string(), true.to_string()));
```
cartertinney, line 89: Is my understanding correct that this enables the feature for the conection, but the more granular usage from the `services` implementation below does it per-message?

rust/azure_iot_operations_mqtt/src/session/wrapper.rs
```diff
@@ -72,10 +80,15 @@ impl Session {
...
         let sat_file = options.connection_settings.sat_file.clone();
 
-        // Add AIO metric to user properties when using AIO MQTT broker features
+        // Add AIO metric and features to user properties when using AIO MQTT broker features
         // TODO: consider user properties from being supported on SessionOptions or ConnectionSettings
-        let user_properties = if options.aio_broker_features {
-            vec![("metriccategory".into(), "aiosdk-rust".into())]
+        let user_properties = if let Some(features) = options.aio_broker_features {
+            let mut user_properties =
+                vec![("metriccategory".to_string(), "aiosdk-rust".to_string())];
```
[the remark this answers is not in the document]
clecompt-msft: The latter is the correct intent here, yes.

rust/azure_iot_operations_protocol/src/telemetry/sender.rs
```diff
@@ -187,8 +187,13 @@ pub struct Message<T: PayloadSerialize> {
     #[builder(default = "None")]
     cloud_event: Option<CloudEvent>,
     /// Indicates whether the message should be retained or not.
-    #[builder(default = "false")]
+    #[builder(default = "self.persist == Some(true)")]
```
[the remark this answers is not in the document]
cartertinney: wow i just can't read

PR #985, opened 2025-08-12
[Rust][Codegen] Expand allowed content-type values

clecompt-msft:
Rust codegen follow-up to #983 to support expanded content-types in generated Rust encodings.


codegen/src/Azure.Iot.Operations.ProtocolCompilerLib/T4/communication/rust/Serialization/t4/RustSerialization.tt
```diff
@@ -19,6 +19,20 @@ use azure_iot_operations_protocol::common::payload_serialize::{
 <# } #>
 use super::<#=this.schemaClassName.GetFileName(TargetLanguage.Rust)#>::<#=this.schemaClassName.GetTypeName(TargetLanguage.Rust)#>;
 
+const <#=this.schemaClassName.GetVariableName(TargetLanguage.Rust).ToUpperInvariant()#>_CONTENT_TYPE: &str = "<#=this.contentType#>";
```
jrdouceur: This will work as a hack, but it would be much better to modify CodeName.cs to add support for SCREAMING_SNAKE_CASE, add a new GetConstantName() method, and then replace this with:
```
<#=this.schemaClassName.GetConstantName(TargetLanguage.Rust, "content", "type")#>
```

vaavva (approved):
If JD's good with the protocol compiler changes, then I'm good with the output!

PR #990, opened 2025-08-18
Add ADR 23 (take 2), Codegen Support for Property

jrdouceur:


doc/dev/adr/0023-property-support.md
```diff
@@ -0,0 +1,306 @@
...
+| --- | --- | --- | --- |
+| Read | "sample/property/read" | | { "Foo": 33, "Bar": "hello" } |
+| Write | "sample/property/write" | { "Bar": "bye" } | |
+
+In general, a model that does not contain a "{propertyName}" token generates at most two Commands, one for reading and one for writing.
+If no Property in the Interface is writable, only a read Command is generated.
+
+The read Command has no request payload, and the write Command has no response payload.
+Fields in the read response are not optional and not nullable; all values are read on each read Command.
+Fields in the write request are optional; only fields that are present and have non-null values will be written.
```
vaavva, line 131: I assume here you mean that there's no additional optionality/nullability added by codegen to allow partial reading, but the property could be optional/nullable at the customer level if desired still, right? And similar for write, there's an additional level of optionality added - a field could still be cleared by a write request if the underlying DTDL property is optional. I don't have a good suggestion for how to clarify these lines, but I think it would be good to clarify something around this:

Fields in the write request have an additional layer of optionality over the DTDL model; only fields that are present and have non-null values will be written. For example, 
if the DTDL is equivalent (apologies for the Rust-ness) to:
```
Foo: Option<int>
Bar: String
```
then write would take
```
Foo: Option<Option<int>>
Bar: Option<String>
```
and if the current value is
```
Foo: Some(3)
Bar: "hello"
```
then sending
```
Foo: Some(None)
Bar: None
```
would end in
```
Foo: None
Bar: "hello"
```

doc/dev/adr/0023-property-support.md
```diff
@@ -0,0 +1,306 @@
...
+| --- | --- | --- |
+| PropertyResult | Object | Result |
+| PropertyValue | Field | NormalResult |
+| ReadError | Field | ErrorResult |
+| WriteError | Field | ErrorResult |
+
+These adjunct types behave nearly identically, *mutatis mutandis*, to the analogous types defined for Command in ADR19.
+Specifically:
+
+* PropertyResult defines the type sent on the wire in a read response.
```
vaavva, line 150: PropertyResult/PropertyValue were a bit hard for me to link properly to read vs write - can you add the takeaways from these sentences to the table? I think that would help me understand this much better!
ex. extra column specifying whether that line corresponds to the read response with no error/write response, etc

I was thinking about WriteResult/ReadResult to align more with both the Error names and the Command adjunct types, but it looks like that would be too specific. (although maybe ReadErrorResult/WriteErrorResult would add better symmetry)

doc/dev/adr/0023-property-support.md
```diff
@@ -0,0 +1,306 @@
...
+* PropertyResult also defines the type sent on the wire in a write response, except there is no payload when there is no error.
+* PropertyValue defines the type returned by the Read API when there is no error.
+* PropertyValue also defines the type accepted by the Write API for writable Properties.
```
vaavva, line 153: isn't this slightly different since there's an extra layer of optionality on write requests vs read responses?

doc/dev/adr/0023-property-support.md
```diff
@@ -0,0 +1,306 @@
...
+| PropertyResult | Object | Result |
+| PropertyValue | Field | NormalResult |
+| ReadError | Field | ErrorResult |
+| WriteError | Field | ErrorResult |
+
+These adjunct types behave nearly identically, *mutatis mutandis*, to the analogous types defined for Command in ADR19.
+Specifically:
+
+* PropertyResult defines the type sent on the wire in a read response.
+* PropertyResult also defines the type sent on the wire in a write response, except there is no payload when there is no error.
```
vaavva, line 151: This is interesting, especially if there are also distinct read/write errors. Generally, until this point, it seems like the DTDL very closely describes/resembles what is sent/possible to be sent over the wire. But now there are some fields that will never be present - such as the write response, that would never have the PropertyValue or the ReadError, but the WriteError is also still optional.

Since we generate the code, I suppose we won't have weirdness around these types showing up on the customer surface though

doc/dev/adr/0023-property-support.md
```diff
@@ -0,0 +1,306 @@
...
+          "name": "propValue",
+          "schema": "integer"
+        }
+      }
+    }
+  ]
+}
+```
+
+A dynamically itemized Property placeholder is defined using co-type Fragmented and a schema that is a Map whose contents are dynamically added Properties.
```
vaavva, line 250: is the `Fragmented` co-type needed, or can we infer this if the schema is a Map? Or does `Fragmented` allow for a different scenario that doesn't include `Fragmented` and treats the Map as any other field that gets fully defined/replaced per write?

doc/dev/adr/0023-property-support.md
```diff
@@ -0,0 +1,306 @@
...
+The Map keys are the names of the dynamic Properties, and the Map values are the Properties themselves.
+
+The Fragmented adjunct type indicates that the Map values have analogous constraints to the Property values of normal static Properties.
+Specifically, Map values in a read response are not nullable, and all values in the Map should be returned on each read Command.
+Map values in a write request are optional and nullable:
+
+* If a key is omitted from the Map in a write request, the value is not written.
+* If a key is included and the value is non-null, the dynamic Property is added if not yet present and updated if already present.
```
vaavva, line 258: This makes sense for Map - is there any plan to support Array as well, or not at this time since the duplicate story is much trickier or should be fully replaced instead of having this `patch` type operation?

doc/dev/adr/0023-property-support.md
```diff
@@ -0,0 +1,306 @@
...
+          }
+        ]
+      }
+    }
+  ]
+}
+```
+
+If a modeler wishes to return a Map of error conditions so that each condition can be paired with a dynamic Property in the Map, this must be defined manually.
+In other words, the schema of the Error must be an Object containing a Map field.
```
vaavva, line 306: It would be helpful to have some sample codegen output for this scenario in! Not even a full output, but just some expected API surface

vaavva (commented):
Some comments, but overall this looks like the right direction to me!

doc/dev/adr/0023-property-support.md
```diff
@@ -0,0 +1,306 @@
...
+
+The DTDL Mqtt extension places no restrictions &mdash; other than basic syntactical constraints &mdash; on the set of tokens used in MQTT topic patterns.
+Topic tokens recognized by the ProtocolCompiler are defined in the [topic-structure](https://github.com/Azure/iot-operations-sdks/blob/main/doc/reference/topic-structure.md) document.
+The sets of tokens differ between RPC and Telemetry, and the following set of tokens is hereby defined for Property:
+
+| Topic token | Description | Required |
+| --- | --- | --- |
+| `{modelId}` | The identifier of the service model, which is the full DTMI of the Interface | optional |
+| `{maintainerId}` | Identifier of the maintainer, by default the MQTT client ID | optional |
+| `{consumerClientId}` | The MQTT client ID of the endpoint that requests to read or write a Property | optional |
```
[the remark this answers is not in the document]
jrdouceur: This was intended to align with the tokens "{executorId}" and "{invokerClientId}" in the [Command topic pattern](https://github.com/Azure/iot-operations-sdks/blob/main/doc/reference/topic-structure.md#command).  The difference is because on the client side (invoker, consumer), the token will always be replaced by the actual MQTT client ID.  But on the service side (executor, maintainer), the MQTT client ID is merely the default replacement value, but it can be overridden by user code if desired.  I don't recall the reason behind this difference, but the difference accounts for the different token strings.

doc/dev/adr/0023-property-support.md
```diff
@@ -0,0 +1,306 @@
...
+| `{action}` | "read" or "write" | optional |
+
+>[!NOTE]
+> The string value of the `propertyTopic` property is not required to contain the `{propertyName}` token, but Property communication differs depending on whether this token is present:
+>
+> * If the `propertyTopic` property contains a `{propertyName}` token, each Property is assigned a separate MQTT pub/sub topic, and each is read or written separately.
+> * If the `propertyTopic` property does not contain a `{propertyName}` token, all Properties in the Interface are grouped into a collection that is read or written with a combined payload. This does not require updating all Properties with each write operation, but it provides the option of writing multiple Properties in a single consistent update.
+
+>[!NOTE]
+> The action token is optional because a model might have no writable properties, in which case all requests are implicitly read requests.
```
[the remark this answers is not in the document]
jrdouceur: Yes, but not in the DTDL language extension.  The ProtocolCompiler interprets the topic strings and knows whether the Property is writable, so it can throw an exception if the topic string would result in ambiguous publication topics.

doc/dev/adr/0023-property-support.md
```diff
@@ -0,0 +1,306 @@
...
+  "payloadFormat": "Json/ecma/404",
+  "propertyTopic": "sample/property/{propertyName}/{action}",
+  "contents": [
+    {
+      "@type": "Property",
+      "name": "Foo",
+      "schema": "integer"
+    },
+    {
+      "@type": "Property",
```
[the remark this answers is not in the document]
jrdouceur: Writability is part of the core DTDL language.  You can "writable" in the [list](https://github.com/Azure/opendigitaltwins-dtdl/blob/master/DTDL/v4/DTDL.v4.md#property) of properties (lowercase 'p') defined for Property (uppercase 'P').  By contrast, NormalResult is an adjunct type defined in the Mqtt extension.  DTDL extensions are not able to add properties to core types (like Property); they can only define adjunct types.

It is possible for an adjunct type to add its own properties, so we could have done something like this:

```json
  "@type": [ "Property", "PotentiallyWritable" ],
  "actuallyWritable": true
```
But this is a bit silly for Boolean properties.  We do this for cases in which additional information beyond a yes/no is needed, such as the string-valued "propertyTopic" property added by the Mqtt adjunct type.

doc/dev/adr/0023-property-support.md
jrdouceur: I'll take a shot at modifying the table and see how it comes out.

Regarding names, I kind of wanted to avoid the term "result" since we're not supporting arbitrary functions that return results.  It is the Write action or the Read action that can yield an error.

doc/dev/adr/0023-property-support.md
jrdouceur: I think it is needed; the explanation of the semantics is in the subsequent paragraph.  And yes, this is to distinguish from cases in which the user models a normal static Property that happens to have a Map for its schema.

doc/dev/adr/0023-property-support.md
jrdouceur: No plan for Array.  If the user wants an Array as a Property schema, this will work just like any other Property schema (including a non-Fragmented Map).  But the Fragmented Map is specifically here to support dynamically itemized Properties, which need keys to identify the actual run-time Property names.

doc/dev/adr/0023-property-support.md
jrdouceur: How about if I just provided an example in DTDL of an error type that included a Map?  Do you think that would be sufficient?

doc/dev/adr/0023-property-support.md
jrdouceur: Added some clarifying text.

doc/dev/adr/0023-property-support.md
jrdouceur: Added a column to the table.

doc/dev/adr/0023-property-support.md
jrdouceur: Added asterisks and subsequent note.

doc/dev/adr/0023-property-support.md
jrdouceur: Added example DTDL error type definition.

doc/dev/adr/0023-property-support.md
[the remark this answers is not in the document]
jrdouceur: That was a mis-aligned example I gave above.  I should have written something like this:

```json
  "@type": [ "Field", "SomeKindOfResult" ],
  "whichKind": "Normal"
```

Not that it really matters, though.  It's still a bit silly.

doc/dev/adr/0023-property-support.md
```diff
@@ -0,0 +1,340 @@
...
+    }
+  ]
+}
+```
+
+This model generates three Commands:
+
+| Command name | MQTT topic | Example request | Example response |
+| --- | --- | --- | --- |
+| ReadFoo | "sample/property/Foo/read" | | { "Foo": 33 } |
```
clecompt-msft, line 103: Do we need the property name in the read/write body when it is specified in the command/topic? That seems redundant (though I could see an argument for consistency between this and the grouped format).

doc/dev/adr/0023-property-support.md
clecompt-msft, line 1: Do you want to add a section or note about incorporating streaming as an observe pattern, once it is available?

doc/dev/adr/0023-property-support.md
jrdouceur: We don't need the name per se, but we do need *some* field in the body.  Due to restrictions of some formats (AVRO and Protobuf, specifically), the payload must be an object, not a primitive type like a string or number.  So, the minimum requirement is an object with a single field, and that field needs a name.  For consistency, we use the same name as the aggregate format uses.  This also aligns with the way Telemetry is defined.

doc/dev/adr/0023-property-support.md
```diff
@@ -0,0 +1,340 @@
...
+Mechanisms for modeling errors in Command responses are defined in [ADR 19](./0019-codegen-user-errs.md).
+The present ADR adds closely analogous mechanisms for modeling errors in Property read/write responses.
+Specifically, the following four adjunct types are added.
+
+| New Adjunct Type | Material Cotype | Analogous ADR19 Type | Field is in message |
+| --- | --- | --- | --- |
+| PropertyResult | Object | Result | |
+| PropertyValue | Field | NormalResult | Read response when no error |
+| ReadError | Field | ErrorResult | Read response when error |
+| WriteError | Field | ErrorResult | Write response when error |
```
vaavva, line 148: Would something like this be more complete?
```suggestion
| New Adjunct Type | Material Cotype | Analogous ADR19 Type | Field is in message |
| --- | --- | --- | --- |
| PropertyResult | Object | Result | Represents full read response, Represents full write response |
| PropertyValue | Field | NormalResult | Read response when no error, Write request |
| ReadError | Field | ErrorResult | Read response when error |
| WriteError | Field | ErrorResult | Write response when error |
```

doc/dev/adr/0023-property-support.md
vaavva: I think even just some psuedo code would be helpful!

vaavva: This is super helpful, thank you!

PR #995, opened 2025-08-21
Containerized `statestore-cli`

yophilav:
- Containerized `statestore-cli`
- Fix renaming mishaps in the readme docs.


tools/statestore-cli/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+FROM mcr.microsoft.com/azurelinux/base/rust:$RUST_VERSION AS build
+
+# Move source code from build context into /app folder
+COPY --link / /app
+WORKDIR /app
+
+# Install additional deps
+RUN tdnf install -y openssl-devel clang
+
+RUN cargo build -p "statestore-cli" --release
```
RLeclair, line 12: Any reason why we can't copy the built application in?

yophilav: It's a good practice to keep the container sterilized. In this container, we build then copy the bin to the final image to be used.

RLeclair (approved):

PR #996, opened 2025-08-22
Update topic-structure.md with new tokens for Property communication

jrdouceur:


doc/reference/topic-structure.md
```diff
@@ -75,6 +76,57 @@ deployments/gen/3/sensors/west/building22/sensor-temp-1628/telemetry
 > * If the `telemetryTopic` property contains a `{telemetryName}` token, each Telemetry is assigned a separate MQTT pub/sub topic, and each is published separately. This may be more efficient for Telemetries that have no logical or temporal relationship to each other.
 > * If the `telemetryTopic` property does not contain a `{telemetryName}` token, all Telemetries in the Interface are grouped into a collection that is published in a combined payload. This does not require any given payload to express values for all Telemetries, but it provides the option of including multiple Telemetries in a single published message.
 
+## Property
```
RLeclair, line 79: I believe this file should be split into two separate documents: one focusing on topic structure at the protocol level, and another detailing accepted protocol tokens at the codegen level. According to [ADR 1](https://github.com/Azure/iot-operations-sdks/blob/main/doc/dev/adr/0001-generalized-topic-tokens.md) we do not recognize any specific tokens at the protocol level - including them in the topic structure documentation could be confusing to someone not using codegen

The `topic-structure.md` for the protocol should primarily contain information from the ADR. Any references to specific topic tokens such as `{modelId}` should be documented in the codegen file, including the recently added `Property` section

Is this something we can do in this PR? We can also leave a TODO or a note at the top and I can get to separating them at a later time

clecompt-msft: I took this whole document to be describing codegen structure, per ADR1, since most of it is focused on the tokens rather than the syntax. It might be enough to just put a note to that effect in this file.

RLeclair: Thanks for the input—agree that the current content is codegen-focused, so adding a note at the top in this PR sounds great

I’d also like to open a follow-up PR later to introduce a Protocol Topic Structure reference that pulls the details from ADR 1. This would give readers who aren’t using codegen a clear source for protocol-level topic rules and our generalized topic token pattern

In that same follow-up, I’d propose renaming this file to something like codegen-topic-structure-pattern.md so its scope is obvious

jrdouceur: I have no objection to subsequent PRs that add clarity and/or improve the alignment of this content with the current state of the system design.  My intent with this PR was merely to avoid adding rot to the present document vis-a-vis the Telemetry pattern.

PR #998, opened 2025-08-26
[Codegen][Rust] Use &str in generated executor_id helper

clecompt-msft:


jrdouceur:
The change looks good, but I think you'll need to re-run gen.sh in a bunch of places in the repo, and then maybe update some call sites in code that uses the generated code.

PR #1000, opened 2025-08-28
Enable defining usage of dataset/event/datapoint/etc-specific fields in connector metadata docs

timtay-microsoft:
Often in our connectors, we expect the user to deploy a device definition with an endpoint address or an asset with an event that has an eventNotifier field that matches some expected shape.

This change allows the connector metadata doc writer to enumerate which of these fields are necessary, and what they should look like if they are necessary

This PR also:
 - adds fields that specifies if assets and endpoints are enabled by default 
 - adds field that specifies what authentication types are supported by each endpoint
 - nests the default/supported destination information inside the relevant supportedModeling field
   - this makes it more obvious that you don't need to specify a default/supported stream/dataset/event destination if your connector does not support modeling streams/datasets/events


doc/akri_connector/connector-metadata-schema.json
```diff
@@ -126,14 +126,45 @@
...
+      "required": ["input"]
     }
   },
   "properties": {
     "aioConnectorMetadataSchemaVersion": {
       "type": "string",
       "description": "The version of AIO connector metadata schema that this connector metadata file adheres to.",
       "pattern": "^(0|[1-9]\\d*)\\.(0|[1-9]\\d*)(?:-((?:0|[1-9A-Za-z-][0-9A-Za-z-]*)(?:\\.(?:0|[1-9A-Za-z-][0-9A-Za-z-]*))*))?(?:\\+([0-9A-Za-z-]+(?:\\.[0-9A-Za-z-]+)*))?$",
-      "default": "3.0-preview"
+      "default": "4.0-preview"
```
RLeclair, line 167: Just checking if there isn't something like `$ref: #/aioConnectorMetadataSchemaVersion` that can be used here

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -355,88 +390,28 @@
...
-                  ]
-                }
+            "supportedAuthenticationTypes": {
+              "type": "array",
+              "items": {
+                "type": "string"
               },
-              "additionalProperties": false,
-              "required": ["supportedDestinations"]
+              "description": "The set of authentication types that this connector supports. For example, [\"usernamePassword\", \"anonymous\"]"
```
RLeclair, line 398: We currently only have one of three authentication methods available, is there a way to validate we're using one of those?

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -467,8 +442,18 @@
                             "alternativeTypeName" : { 
                               "$ref" : "#/definitions/alternativeTypeName",
                               "description": "The alternative name to the type \"dataPoints\" within an event"
+                            },
+                            "fields": {
+                              "type": "object",
+                              "description" : "Describes if this connector expects these fields to be provided in the asset defintion when deploying the connector. Also describes the expected shape of these values if they are supported.",
```
RLeclair, line 448:
```suggestion
                              "description" : "Describes if this connector expects these fields to be provided in the data point definition when deploying the connector. Also describes the expected shape of these values if they are supported.",
```

Is this specific for data points?

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -480,10 +465,46 @@
                     "eventConfigurationSchema": {
                       "$ref": "http://json-schema.org/draft-07/schema#",
                       "description": "The JSON schema for the \"eventConfigurationSchema\" field on an event."
+                    },
+                    "fields": {
+                      "type": "object",
+                      "description" : "Describes if this connector expects these fields to be provided in the asset defintion when deploying the connector. Also describes the expected shape of these values if they are supported.",
```
RLeclair, line 471:
```suggestion
                      "description" : "Describes if this connector expects these fields to be provided in the event definition when deploying the connector. Also describes the expected shape of these values if they are supported.",
```

Similar to above. Technically it is the asset definition but we might want to be more specific that these are fields on the event

RLeclair: Also applies for the below descriptions

RLeclair (approved):

doc/akri_connector/connector-metadata-schema.json
timtay-microsoft: It's possible to codify the allowed values here to be something like ["usernamePassword", "anonymous", ...], but that would limit users to our understanding of authentication. I'd rather we keep this open-ended so that we support future authentication mechanisms

doc/akri_connector/connector-metadata-schema.json
timtay-microsoft: Not that I kind find, unfortunately

doc/akri_connector/connector-metadata-schema.json
timtay-microsoft: Good idea. This was copy-pasted, but I can make these more specific

PR #1004, opened 2025-09-04
Implement ADR 23 for .NET

jrdouceur:
This PR implements [ADR 23](https://github.com/Azure/iot-operations-sdks/blob/main/doc/dev/adr/0023-property-support.md), Property support, for C#.  This includes optional aggregation, error modeling, dynamically itemized properties, and the cross product of these three features.

Most changes are to the SchemaGenerator and the serialization T4 templates, which will be re-usable across target languages.

Changes to the EnvoyGenerator and the communication T4 templates are almost exclusively related to .NET code generation.  Other languages will follow.

Only JSON serialization is fully supported.  There is sketchy support for AVRO serialization, mainly just enough to enable the ProtocolCompiler to continue to build.  Proper support for AVRO will await future development effort.


jrdouceur:
> I think the Rust support would make this easier for me to review! I skimmed this, but I'm not deeply familiar with all of the codegen code to say that I strongly reviewed this. I can definitely pull the changes and generate the rust code to look over the output once you've got it though!

@timtay-microsoft, do you expect that you'll have a chance to review the current PR this week?  If so, I will await your review before proceeding with Rust.  If not, I'll move forward with Rust, and I'll just hope there is no latent issue that forces a major rewrite.

timtay-microsoft:
@jrdouceur I should have time to look at this today or tomorrow, sorry for the delay

jrdouceur:
> Also, the client side has a dispose method, but doesn't implement IAsyncDisposable

Whoops!  Good catch.  Fixed in latest commit.

jrdouceur:
@avishekpant or @cartertinney -- could one of you add an approval to this PR?  These changes relate to generating .NET code, but because of some incidental code rearrangement in the generated Rust, this needs a Rust approval in addition to the .NET approval that Tim provided.  If I can get this approveda and merged in, it will be easier for me to start a new PR for targeting Rust.

vaavva (approved):
Apologies for the delay! Approving the rust changes (I've noticed these changes a lot lately, I think there's some variance between some cargo fmt settings alphabetizing based on the json name vs the rust name and I think it keeps causing flip-flopping check-ins :( )

PR #1007, opened 2025-09-09
Address more connector metadata schema feedback

timtay-microsoft:
- Allow for parameterized tokens such as {assetName} and {namespace} in the default MQTT topic + default dss key, + default storage path
- Un-nest the "supportedModeling" types

### DIFF
--- a/doc/akri_connector/connector-metadata-schema.json
+++ b/doc/akri_connector/connector-metadata-schema.json
@@ -1,8 +1,8 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "$id": "https://json.schemastore.org/aio-connector-metadata-5.0-preview.json",
+  "$id": "https://json.schemastore.org/aio-connector-metadata-6.0-preview.json",
   "type": "object",
-  "title": "JSON schema for Azure IoT Operations Connector Metadata 5.0-preview",
+  "title": "JSON schema for Azure IoT Operations Connector Metadata 6.0-preview",
   "description": "Schema that defines how to write a connector metadata document when writing connectors for Azure IoT Operations solutions",
   "definitions": {
     "modelLimits": {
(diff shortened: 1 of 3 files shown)


doc/akri_connector/connector-metadata-schema.json
```diff
@@ -66,7 +66,7 @@
         },
         "key": {
           "type": "string",
-          "description": "The default broker state store key that will be published to"
+          "description": "The default broker state store key that will be published to. Values are allowed to contain deploy-time parameters like 'dss/{deviceName}/{inboundEndpointName}'. The supported parameter tokens are: {deviceName}, {inboundEndpointName}, {assetName}, {connectorClientId}, {namespace}, and {endpointType}."
```
timtay-microsoft, line 69: @hansgschossmann you mentioned "namespace" in the meeting yesterday as a potential parameter here. Just to confirm, but you were referring to the Kubernetes namespace that connector is deployed in, correct?

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -390,307 +390,279 @@
               "type": "string",
               "description": "The version of the protocol of this inbound endpoint."
             },
-            "supportedModeling": {
+            "events": {
```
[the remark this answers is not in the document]
timtay-microsoft: I will probably tackle that later. The SDK as a whole should probably consume that change all at once

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -34,7 +34,7 @@
         },
         "topic": {
           "type": "string",
-          "description": "The default MQTT topic that will be published to"
+          "description": "The default MQTT topic that will be published to. Values are allowed to contain deploy-time parameters like 'mqtt/{deviceName}/{inboundEndpointName}/myMessages'. The supported parameter tokens are: {deviceName}, {inboundEndpointName}, {assetName}, {connectorClientId}, {datasetName}, {eventName}, {streamName}, {managementGroupName}, {datasetDatapointName}, {eventDatapointName}, {namespace}, and {endpointType}."
```
vaavva, line 37: datapoints wouldn't have destinations, so I'm not sure there's a meaningful value that could replace those tokens. Also not sure if you want to call out that {datasetName} could only be used as a token for a dataset?

timtay-microsoft: Good point x 2

PR #1020, opened 2025-09-18
Connector metadata schema version 7.0-preview

timtay-microsoft:
Add support for event groups

### DIFF
--- a/doc/akri_connector/connector-metadata-schema.json
+++ b/doc/akri_connector/connector-metadata-schema.json
@@ -1,6 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "$id": "https://json.schemastore.org/aio-connector-metadata-6.0-preview.json",
+  "$id": "https://json.schemastore.org/aio-connector-metadata-7.0-preview.json",
   "type": "object",
   "title": "JSON schema for Azure IoT Operations Connector Metadata 6.0-preview",
   "description": "Schema that defines how to write a connector metadata document when writing connectors for Azure IoT Operations solutions",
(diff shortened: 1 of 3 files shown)


marcschier (approved):

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -564,7 +567,7 @@
                 "limits": { "$ref": "#/definitions/modelLimits" },
                 "managementGroupConfigurationSchema": {
```
vaavva, line 568: management groups now have a dataSource

timtay-microsoft:
> > * Basically, "events" were renamed to "eventGroups", and "dataPoints" within an event are now "events" within an event group
> 
> This seems like a very reasonable change to me, but are we sure this is acceptable with regard to the terminology used in asset definitions? It may be worth checking with @chrisgclayton to confirm.

These changes are coming from changes in the asset definitions. If you are curious: https://msazure.visualstudio.com/One/_git/akri/pullrequest/13484910?path=/adr-service-generated/dtdl/adr-base-service.json

jrdouceur (approved):

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -390,56 +390,59 @@
...
+                    "dataSource": {
                       "$ref": "#/definitions/fieldNecessity"
                     },
                     "typeRef": { "$ref": "#/definitions/fieldNecessity" }
                   },
-                  "required": ["eventNotifier", "typeRef"],
+                  "required": ["dataSource", "typeRef"],
                   "additionalProperties": false
                 },
                 "destinations": {
```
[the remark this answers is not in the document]
timtay-microsoft: To be clear, it looks like event groups don't have a concept of a destination, so I just need to move this destinations block out of the "event groups" section and into the "events" section

vaavva (approved):

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -390,82 +390,85 @@
               "type": "string",
               "description": "The version of the protocol of this inbound endpoint."
             },
-            "events": {
+            "eventGroups": {
```
mrtalley, line 393: Unrelated to your current changes, but Val and I just realized we don't have a `description` for the inbound endpoint. Is it possible to add one? There's a field in the connector template for this, but we have nowhere to specify it in the CM

timtay-microsoft: Sure thing. Added.

doc/akri_connector/connector-metadata-schema.json
```diff
@@ -390,82 +390,85 @@
...
+                    "dataSource": {
                       "$ref": "#/definitions/fieldNecessity"
                     },
                     "typeRef": { "$ref": "#/definitions/fieldNecessity" }
                   },
-                  "required": ["eventNotifier", "typeRef"],
+                  "required": ["dataSource", "typeRef"],
                   "additionalProperties": false
-                },
-                "destinations": {
```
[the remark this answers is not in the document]
timtay-microsoft: > Just to confirm - the expectation is that the destination specified at the event level also applies to the defaultEventDestination and defaultDestination (in eventGroup) right?

If you are asking about how a connector behaves, that may be better answered by @vaava, but my understanding is that the more specific default destination (per a specific event vs per every event in an event group) will take precedence over the more general default destination. It wouldn't be additive.

PR #1023, opened 2025-09-22
[Rust][Connector][Services] ADR EventGroups Support (and other 2510 asset ADR updates)

vaavva:
This PR:
- Updates the DTDL for the ADR service for changes coming for Assets for 2510. This (mainly) includes the addition of EventGroups, but there are also a few other minor field adjustments
- Regenerates ADR service code off of the new DTDL changes (cleaned up a few stale files as well)
- Updates the ADR Client Service SDK to handle the new DTDL
- Updates the Connector SDK to handle the new DTDL
  - Updates how events are handled in a DataOperation - they now use an EventSpecification struct that contains the Event definition and that event's EventGroupSpecification. Also changes the "name" of events to be a tuple to be able to dedupe Events on different EventGroups with the same name
  - Updates status reporting to include the eventGroup layer (although eventGroups don't have their own status reporting)
  - Updates the DataOperationRef and DataOperation to formats that better handle eventGroups
  - Updates base connector sample (most changes are some log redundancy simplifications that are only sorta tied to the SDK changes)
 
This PR DOES NOT:
- support eventGroup.default_events_destinations. This field is not used to inform the destination of an event currently
- support separate eventGroup notifications - any updates to an eventGroup definition will be reported on each of its Events and cannot be received separately. Note that this limitation does not negatively affect this scenario: updates to event1 on eventGroupA do not cause a notification on event2 on eventGroupA

State of testing:
- Service is not available yet, so none of this has been able to be tested yet beyond compilation. Before this is released, these changes must be tested with the service

### DIFF
--- a/rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
+++ b/rust/azure_iot_operations_connector/src/base_connector/managed_azure_device_registry.rs
@@ -3353,7 +3353,6 @@ impl DataOperationDefinition {
 /// this trait allows individual data operation types (e.g., `Dataset`, `Event`, `Stream`)
 /// to define their own behavior while still conforming to a common interface.
 trait DataOperation: Into<DataOperationDefinition> {
-    // trait DataOperation: Into<DataOperationDefinition> {
     type HashName: PartialEq + Eq + Hash + Clone + std::fmt::Debug;
     fn kind() -> DataOperationKind;
     fn hash_name(&self) -> &Self::HashName;


eng/dtdl/adr-base-service.json
```diff
@@ -1944,6 +1981,12 @@
             "elementSchema": "dtmi:com:microsoft:akri:EventStreamDestination"
           }
         },
+        {
+          "@type": "Field",
+          "name": "description",
+          "description": "Human-readable description of the asset.",
+          "schema": "string"
```
jrdouceur, line 1988: I don't know if this matters, but we often use localized strings for "description" and "displayName", so this might be better as a Map from string (language codes) to string (language-specific description strings), rather than merely a string.

eng/dtdl/adr-base-service.json
```diff
@@ -1953,6 +1996,12 @@
           "description": "Reference to the device that provides data for this asset. Must provide device name & endpoint on the device to use.",
           "schema": "dtmi:com:microsoft:akri:AssetDeviceRef;1"
         },
+        {
+          "@type": "Field",
+          "name": "displayName",
+          "description": "Human-readable display name.",
+          "schema": "string"
```
jrdouceur, line 2003: Ditto above, mutatis mutandis.

jrdouceur (approved):
DTDL appears sane.  Feel free to ignore my comments about localization if you think single strings are better in this case.

eng/dtdl/adr-base-service.json
vaavva: The DTDL updates are coming from the Service, so we shouldn't be modifying them at all in our repo. Although this is good feedback, feel free to surface it to the ADR team to think about for the next release. Going to resolve these comments for now since they wouldn't take effect in this PR unless higher up takes this change

timtay-microsoft:
>Also changes the "name" of events to be a tuple to be able to dedupe Events on different EventGroups with the same name

Is this an ask from someone? I'm a bit out-of-the-loop, but this strikes me as incorrect. If the ADR modeling allows you to define the same event in 2 different groups, I wouldn't expect our SDK to treat them as one

timtay-microsoft:
Apologies since I can't read rust that well, but does an event still map to a single MQTT message/single DSS key?

PR #1034, opened 2025-09-29
Update connector metadata schema to version 8.0-preview

timtay-microsoft:
- Fix bug where multiple supported architectures could not be specified
- Specify pointer to schema doc rather than specifying version of the schema to use
    - This gives most JSON editors the ability to check schema validity while writing it

### DIFF
--- a/doc/akri_connector/connector-metadata-schema.json
+++ b/doc/akri_connector/connector-metadata-schema.json
@@ -1,8 +1,8 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "$id": "https://json.schemastore.org/aio-connector-metadata-7.0-preview.json",
+  "$id": "https://json.schemastore.org/aio-connector-metadata-8.0-preview.json",
   "type": "object",
-  "title": "JSON schema for Azure IoT Operations Connector Metadata 7.0-preview",
+  "title": "JSON schema for Azure IoT Operations Connector Metadata 8.0-preview",
   "description": "Schema that defines how to write a connector metadata document when writing connectors for Azure IoT Operations solutions",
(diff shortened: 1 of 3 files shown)


doc/akri_connector/connector-metadata-schema.json
```diff
@@ -134,11 +134,10 @@
...
   },
   "properties": {
-    "aioConnectorMetadataSchemaVersion": {
+    "$schema": {
       "type": "string",
-      "description": "The version of AIO connector metadata schema that this connector metadata file adheres to.",
-      "pattern": "^(0|[1-9]\\d*)\\.(0|[1-9]\\d*)(?:-((?:0|[1-9A-Za-z-][0-9A-Za-z-]*)(?:\\.(?:0|[1-9A-Za-z-][0-9A-Za-z-]*))*))?(?:\\+([0-9A-Za-z-]+(?:\\.[0-9A-Za-z-]+)*))?$",
-      "default": "6.0-preview"
+      "description": "The address of the AIO connector metadata schema that this connector metadata file adheres to.",
+      "default": "https://raw.githubusercontent.com/Azure/iot-operations-sdks/refs/heads/main/doc/akri_connector/connector-metadata-schema.json"
```
varunpuranik, line 140: This URL does not have the version of the schema. 
Which means this will resolve to the LATEST version of the metadata schema. 
We need to have per-version files for the schema. 
And maybe we should consider having both the version and the ref. With the ref being optional.

varunpuranik: Alright, fair point. But then why have pointers to 2 different locations? Why not the latest one in the schemastore?

PR #1036, opened 2025-10-01
Fix connector metadata not allowing multiple array entries

timtay-microsoft:
Update sample to validate this change


jrdouceur (approved):
JSON Schema looks good

mrtalley (approved):

PR #1063, opened 2025-11-05
[Rust][MQTT] Adding connection settings adapter for new MQTT client

RLeclair:
This PR:
- Adds new MQTT client adapter integration
- Adds `to_azure_mqtt_connect_parameters` function to convert `MqttConnectionSettings` to the configuration needed for the new client
- Returns a tuple of `ClientOptions` and `AzureMqttConnectParameters`. The former is used once during client initialization, while the latter can be reused for subsequent connections
- Within `AzureMqttConnectParameters`, `authentication_info()` reads the SAT token on-demand. This ensures we always use the current token since it may have changed between the initial configuration and actual connection attempts
- Adds conversions for all configuration parameters needed by the new MQTT client
- Adds comprehensive unit tests (missing SAT token test case)

### DIFF
--- a/rust/azure_iot_operations_mqtt/src/azure_mqtt_adapter.rs
+++ b/rust/azure_iot_operations_mqtt/src/azure_mqtt_adapter.rs
@@ -201,11 +201,11 @@ pub struct AzureMqttConnectParameters {
     /// Keep alive duration
     pub keep_alive: Duration,
     /// Connection transport configuration
-    pub connection_transport_config: azure_mqtt::client::ConnectionTransportConfig,
+    pub connection_transport_config: ConnectionTransportConfig,
     /// Connect options
-    pub connect_options: azure_mqtt::packet::ConnectOptions,
+    pub connect_options: ConnectOptions,
     /// Connect properties
-    pub connect_properties: azure_mqtt::packet::ConnectProperties,
+    pub connect_properties: ConnectProperties,
     // Optional SAT file path for authentication, saved here to be read later
     sat_file: Option<String>,
 }
@@ -241,7 +241,7 @@ impl MqttConnectionSettings {
     ///
     /// # Parameters
     /// - `user_properties`: User properties to include in the connect properties
-    /// - `outgoing_max`: Outgoing message queue size
+    /// - `outgoing_max`: Outgoing message queue size (100 is a good default value)
     ///
     /// # Errors
     /// Returns [`ConnectionSettingsAdapterError`] if any conversion fails
@@ -249,13 +249,7 @@ impl MqttConnectionSettings {
         self,
         user_properties: Vec<(String, String)>,
         outgoing_max: usize,
-    ) -> Result<
-        (
-            azure_mqtt::client::ClientOptions,
-            AzureMqttConnectParameters,
-        ),
-        ConnectionSettingsAdapterError,
-    > {
+    ) -> Result<(ClientOptions, AzureMqttConnectParameters), ConnectionSettingsAdapterError> {
         let client_options = ClientOptions {
             client_id: Some(self.client_id),
             queue_size: outgoing_max,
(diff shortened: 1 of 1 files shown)


rust/azure_iot_operations_mqtt/src/azure_mqtt_adapter.rs
```diff
@@ -0,0 +1,575 @@
...
+    // Receive Maximum
+    let receive_maximum =
+        NonZero::new(receive_max).ok_or_else(|| ConnectionSettingsAdapterError {
+            msg: "receive_max must be > 0".to_string(),
+            field: ConnectionSettingsField::ReceiveMax(receive_max),
+            source: None,
+        })?;
+
+    Ok(ConnectProperties {
+        session_expiry_interval: SessionExpiryInterval::Duration(session_expiry_secs),
```
vaavva: If this is a Duration, why do we need to translate it to a u32 first?

rust/azure_iot_operations_mqtt/src/azure_mqtt_adapter.rs
```diff
@@ -0,0 +1,575 @@
...
+
+    Ok(ca_certs)
+}
+
+fn tls_config(
+    ca_file: Option<String>,
+    cert_file: Option<String>,
+    key_file: Option<String>,
+    key_password_file: Option<String>,
+) -> Result<(Option<ClientCert>, Vec<X509>), anyhow::Error> {
```
vaavva: can you add a comment documenting what these return types represent? (client_cert, ca_trust_bundle)
Just for readability in the future

rust/azure_iot_operations_mqtt/src/azure_mqtt_adapter.rs
```diff
@@ -0,0 +1,575 @@
...
+    ca_certs.append(certs);
+
+    if ca_certs.is_empty() {
+        Err(TlsError::new("No CA certs available in CA File"))?;
+    }
+
+    ca_certs.sort();
+    ca_certs.dedup();
+
+    Ok(ca_certs)
```
vaavva: do we need any comment here about CA revocation checks? Not sure if that's necessary or not

vaavva (approved):

rust/azure_iot_operations_mqtt/src/azure_mqtt_adapter.rs
```diff
@@ -0,0 +1,575 @@
...
+    /// # Parameters
+    /// - `user_properties`: User properties to include in the connect properties
+    /// - `outgoing_max`: Outgoing message queue size
+    ///
+    /// # Errors
+    /// Returns [`ConnectionSettingsAdapterError`] if any conversion fails
+    pub fn to_azure_mqtt_connect_parameters(
+        self,
+        user_properties: Vec<(String, String)>,
+        outgoing_max: usize,
```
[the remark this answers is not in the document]
RLeclair: 100 is what we were using for rumqttc so I'll add a comment with that

rust/azure_iot_operations_mqtt/src/azure_mqtt_adapter.rs
```diff
@@ -0,0 +1,575 @@
...
+    ) -> Result<
+        (
+            azure_mqtt::client::ClientOptions,
```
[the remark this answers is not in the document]
RLeclair: Great catch! I missed replacing a couple of paths there

rust/azure_iot_operations_mqtt/src/azure_mqtt_adapter.rs
RLeclair: There was a comment in rumqttc but in tls config, I've added it back here!

rust/azure_iot_operations_mqtt/src/azure_mqtt_adapter.rs
```diff
@@ -0,0 +1,573 @@
...
+        self,
+        user_properties: Vec<(String, String)>,
+        outgoing_max: usize,
+    ) -> Result<(ClientOptions, AzureMqttConnectParameters), ConnectionSettingsAdapterError> {
+        let client_options = ClientOptions {
+            client_id: Some(self.client_id),
+            queue_size: outgoing_max,
+        };
+
+        let connect_options =
```
cartertinney: Flattened this in my latest PR since you said it was confusing

PR #1070, opened 2025-11-07
[Rust][MQTT] Session Exit Handle

vaavva:
The full Session Exit Handle flow/logic might need some more adjustments, but this at least fixes a lot of the compiler issues around this scenario

### DIFF
--- a/rust/azure_iot_operations_mqtt/src/session/session.rs
+++ b/rust/azure_iot_operations_mqtt/src/session/session.rs
@@ -270,6 +270,8 @@ impl Session {
             } else {
                 match self
                     .connect_handle
+                    // This can't fail because the connect_handle is always Some() on new, and run consumes self
+                    // And connect_handle is always re-set to Some() before the next loop iteration.
                     .take()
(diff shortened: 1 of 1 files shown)


rust/azure_iot_operations_mqtt/src/session/session.rs
```diff
@@ -242,44 +245,91 @@ impl Session {
...
+                //             // have already exited.
+                //             log::error!("Error sending auth code to SAT auth task: {e:?}");
+                //         }
+                //     }
+                // }
+            } else {
+                match self
+                    .connect_handle
+                    .take()
+                    .unwrap()
```
cartertinney: Leave a comment indicating this cannot fail because a handle will always be present when running this function

cartertinney (approved):

rust/azure_iot_operations_mqtt/src/session/session.rs
vaavva: added!

PR #1085, opened 2025-11-14
[Rust][MQTT][Protocol] MQTT Client integration to protocol and METL tests

vaavva:
This PR:
- updates the protocol crate (excluding network tests and samples) to the new MQTT Client/MQTT Crate updates
- adds support in the MQTT crate to inject/sniff network traffic instead of needing to mock the session
- updates METL tests to work directly with the network packets
- fixes some outdated METL tests

This PR does not:
- fully finish topic handling in protocol, this will be done as a follow-up item

### DIFF
--- a/rust/azure_iot_operations_protocol/tests/metl/telemetry_receiver_tester.rs
+++ b/rust/azure_iot_operations_protocol/tests/metl/telemetry_receiver_tester.rs
@@ -44,6 +44,7 @@ struct ReceivedTelemetry {
     source_id: Option<String>,
 }
 
+// TODO: this struct could probably be removed
 pub struct TelemetryReceiverTester {}
 
 impl TelemetryReceiverTester {
--- a/rust/azure_iot_operations_protocol/tests/metl/telemetry_sender_tester.rs
+++ b/rust/azure_iot_operations_protocol/tests/metl/telemetry_sender_tester.rs
@@ -34,6 +34,7 @@ const TEST_TIMEOUT: time::Duration = time::Duration::from_secs(10);
 
 type SendResultReceiver = oneshot::Receiver<Result<(), AIOProtocolError>>;
 
+// TODO: this struct could probably be removed
 pub struct TelemetrySenderTester {}
 
 impl TelemetrySenderTester {
--- a/rust/azure_iot_operations_mqtt/Cargo.toml
+++ b/rust/azure_iot_operations_mqtt/Cargo.toml
@@ -26,7 +26,6 @@ openssl = "0.10.66"                                                             
 rand = "0.8.5"
 thiserror.workspace = true
 tokio.workspace = true
-tokio-util.workspace = true
 
 [dev-dependencies]
 azure_iot_operations_mqtt = { path = ".", features = ["test-utils"] }
--- a/rust/azure_iot_operations_protocol/tests/metl/command_executor_tester.rs
+++ b/rust/azure_iot_operations_protocol/tests/metl/command_executor_tester.rs
@@ -34,6 +34,7 @@ use crate::metl::test_payload::TestPayload;
 
 const TEST_TIMEOUT: time::Duration = time::Duration::from_secs(10);
 
+// TODO: this struct could probably be removed
 pub struct CommandExecutorTester {}
 
 impl CommandExecutorTester {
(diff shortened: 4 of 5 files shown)


eng/test/test-cases/SanityCheck/CommandExecutor/CommandExecutorCatchWronglyInApplication.yaml
```diff
@@ -4,12 +4,12 @@ description:
...
     CommandExecutor initialized with a topic namespace that is invalid.
   expect: >-
-    Erroneous test case expects CommandExecutor to throw exception incorrectly indicating error in application.
+    Erroneous test case expects CommandExecutor to throw exception incorrectly indicating error is remote.
 
 prologue:
   executors:
   - topic-namespace: "invalid/{modelId}"
   catch:
     error-kind: invalid configuration
```
jrdouceur: I don't understand the changes to the sanity checks.  There should already be cases to ensure that the IsRemote field is being checked by the METL infrastructure; these cases are to check that the InApplication field is being checked (per the titles of the test cases).

vaavva: ah, in that case I can remove these tests entirely - in-application isn't a field that metl can check for anymore, so these tests would always pass (when they should fail)

vaavva: nope, all languages! It got removed from METL (and the SDKs) here: https://github.com/Azure/iot-operations-sdks/pull/592/files, but the sanity tests didn't get updated

rust/azure_iot_operations_protocol/src/common/payload_serialize.rs
```diff
@@ -137,7 +163,7 @@ impl PayloadSerialize for BypassPayload {
         };
         Ok(BypassPayload {
             content_type: ct,
-            format_indicator: format_indicator.clone(),
+            format_indicator: *format_indicator,
```
cartertinney: If it implements `Copy` now, this is still more complex than it needs to be - don't even bother taking a reference in the params and simplify it further.

rust/azure_iot_operations_protocol/tests/metl/telemetry_sender_tester.rs
```diff
@@ -476,138 +471,95 @@ where
...
-                    }
-                }
-                _ => {
-                    panic!("expected metadata but found no properties in published message");
+            for (key, value) in &expected_message.metadata {
+                let found = published_message
+                    .other_properties
+                    .user_properties
+                    .iter()
+                    .find(|&k| k.0 == **key);
```
cartertinney: Keys are not guaranteed to be unique in MQTT.

rust/azure_iot_operations_protocol/tests/metl/telemetry_receiver_tester.rs
```diff
@@ -44,22 +44,13 @@ struct ReceivedTelemetry {
...
 }
 
-pub struct TelemetryReceiverTester<C>
-where
-    C: ManagedClient + Clone + Send + Sync + 'static,
-    C::PubReceiver: Send + Sync + 'static,
-{
-    managed_client: PhantomData<C>,
-}
+pub struct TelemetryReceiverTester {}
```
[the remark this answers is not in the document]
vaavva: I don't think so, but I didn't want to touch METL that drastically more than I needed to. If you've got strong feelings, I'll look into pulling this out

eng/test/test-cases/SanityCheck/CommandExecutor/CommandExecutorCatchWronglyInApplication.yaml
vaavva: this changed quite a bit more than I was expecting - please let me know if this is what you would have expected to happen or not!

vaavva: @jrdouceur, I'm going to merge this PR since it's just into our feature branch, but I'll follow up with you on an email thread to make sure this resolves how we expect before this would go to main!

PR #1087, opened 2025-11-14
Draft for Cloud Events headers

maximsemenov80:
This pull request introduces several new classes to support configuration and telemetry for Azure IoT Operations Connector, focusing on CloudEvents metadata generation, asset and connector configuration, and telemetry sender creation. The changes lay out the foundational structure for managing device, endpoint, asset, and dataset information, and for integrating CloudEvents-compliant telemetry into the connector SDK. Many methods and validation steps are marked as TODOs, indicating upcoming implementation work.

### DIFF
--- a/dotnet/src/Azure.Iot.Operations.Connector/CloudEvents/AioCloudEvent.cs
+++ b/dotnet/src/Azure.Iot.Operations.Connector/CloudEvents/AioCloudEvent.cs
@@ -1,0 +1,86 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Azure.Iot.Operations.Protocol.Telemetry;
+
+namespace Azure.Iot.Operations.Connector.CloudEvents;
+
+/// <summary>
+/// Azure IoT Operations CloudEvent with automatic metadata population.
+/// Contains all CloudEvents fields with AIO-specific extensions (aiodeviceref, aioassetref).
+/// This class holds all the CloudEvents headers that will be automatically
+/// populated based on connector, device, asset, and dataset configuration.
+/// </summary>
+public class AioCloudEvent
+{
+    /// <summary>
+    /// Generated CloudEvents source URI.
+    /// Formula: ms-aio:<Device-CompoundKey>|<ProtocolSpecificIdentifier>|<Device-externaldeviceId*>|<Device-Name>[/Sub-Source]
+    /// Note: "|" represents OR - first available value in priority order is used.
+    /// </summary>
+    public required Uri Source { get; init; }
+
+    /// <summary>
+    /// Generated CloudEvents type string.
+    /// Formula: ["DataSet"|"Event"]/<typeref-property-value>
+    /// </summary>
+    public required string Type { get; init; }
+
+    /// <summary>
+    /// Generated CloudEvents subject string.
+    /// Formula: <Asset-CompoundKey>|<Asset-ExternalAssetId*>|<AssetName>/<DataSet-Name>|<EventGroup-Name>[/Sub-Subject]
+    /// Note: "|" represents OR - first available value in priority order is used.
+    /// </summary>
+    public required string Subject { get; init; }
+
+    /// <summary>
+    /// Schema registry ID for the payload schema.
+    /// Maps to CloudEvents dataschema attribute.
+    /// </summary>
+    public string? DataSchema { get; init; }
+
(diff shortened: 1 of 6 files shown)


dotnet/src/Azure.Iot.Operations.Connector/Models/DeviceIdentity.cs
```diff
@@ -0,0 +1,48 @@
...
+// Licensed under the MIT License.
+
+namespace Azure.Iot.Operations.Connector.Models;
+
+/// <summary>
+/// Represents device identity information used for CloudEvents correlation.
+/// This class captures all device-level identifiers needed to construct
+/// CloudEvents headers according to the message_correlation.md specification.
+/// </summary>
+public class DeviceIdentity
```
timtay-microsoft: Most of this information is already present in the ADR model type for a device ([here](https://github.com/Azure/iot-operations-sdks/blob/main/dotnet/src/Azure.Iot.Operations.Services/AssetAndDeviceRegistry/Models/Device.cs)).

From what I understood from the email thread, the compound key is the only new addition here, but Alex is proposing it to be modeled in the ADR service.

With that in mind, I'm not sure that we would need this connector-level representation of a device

dotnet/src/Azure.Iot.Operations.Connector/Models/EndpointIdentity.cs
```diff
@@ -0,0 +1,39 @@
...
+    public required string Name { get; init; }
+    
+    /// <summary>
+    /// Protocol-specific identifier used to uniquely identify the endpoint.
+    /// Examples:
+    /// - OPC UA: "opc.tcp://192.168.1.10:4840/UA/Server"
+    /// - REST/SSE: "https://api.example.com/v1/telemetry"
+    /// - Modbus: "modbus.tcp://192.168.1.20:502"
+    /// </summary>
+    public required string ProtocolIdentifier { get; init; }
```
timtay-microsoft: Isn't this just the [endpoint address](https://github.com/Azure/iot-operations-sdks/blob/e77761b4f075504bbe54805f048b017bec0fa933/dotnet/src/Azure.Iot.Operations.Services/AssetAndDeviceRegistry/Models/DeviceOutboundEndpoint.cs#L5)? If so, then, again, this data type is already modeled in the ADR service so we probably don't need to model it again in the connector package

dotnet/src/Azure.Iot.Operations.Connector/ConnectorTelemetryFactory.cs
```diff
@@ -0,0 +1,151 @@
...
+        throw new NotImplementedException();
+    }
+    
+    /// <summary>
+    /// Creates connector telemetry senders for all datasets in an asset configuration.
+    /// </summary>
+    /// <typeparam name="T">The type of telemetry data to send.</typeparam>
+    /// <param name="assetConfig">Asset configuration with all its datasets.</param>
+    /// <returns>A dictionary mapping dataset names to their connector telemetry senders.</returns>
+    public async Task<IReadOnlyDictionary<string, ConnectorTelemetrySender<T>>> CreateSendersForAssetAsync<T>(
```
timtay-microsoft: Assets can be discovered separately, so I don't think we will need a bulk API for this. Just the single CreateSenderForAssetAsync should be sufficient

dotnet/src/Azure.Iot.Operations.Connector/ConnectorTelemetryFactory.cs
```diff
@@ -0,0 +1,151 @@
...
+    
+    /// <summary>
+    /// Creates a connector telemetry sender for a specific asset and dataset.
+    /// </summary>
+    /// <typeparam name="T">The type of telemetry data to send.</typeparam>
+    /// <param name="asset">Asset identity information.</param>
+    /// <param name="dataSet">Dataset or event group definition.</param>
+    /// <param name="topicPattern">Optional MQTT topic pattern override. If null, uses default pattern.</param>
+    /// <returns>A configured connector telemetry sender.</returns>
+    public Task<ConnectorTelemetrySender<T>> CreateSenderAsync<T>(
```
timtay-microsoft: nit: this function probably won't have any async calls in it (no network calls, no semaphores, etc.). You can just make it 

```
public ConnectorTelemetrySender<T> CreateSender<T>(
```

PR #1090, opened 2025-11-17
[MQTT] Connection management initial pass

cartertinney:
* Implemented connection and reconnection logic
* Implemented basic exit logic
* Implemented AIO MQ-specific enhanced auth logic for `Session` initial connect
* Implemented generic enhanced auth logic for `Session` reauthentication
* Implemented generic `AuthPolicy` trait that can be extended to support other auth scenarios

Limitations:
- Generic enhanced auth initial connect is not included here (despite generic reauth logic being included) to avoid overindexing on implementation structure pending additional testing
- Still evaluating `force_exit` using `SessionExitHandle` - will readd alongside testing after exploring edge cases more.
- `state` module contains significant complexity that is no longer in use, and will need to be simplified in a follow-up, once again, alongside testing for validation
- Samples were not updated yet
- Custom `AuthPolicy` injection is not exposed via API despite being generic as it will create issues with `MqttConnectionSettings`

### DIFF
--- a/rust/azure_iot_operations_mqtt/src/lib.rs
+++ b/rust/azure_iot_operations_mqtt/src/lib.rs
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-//#![warn(missing_docs)]
+#![warn(missing_docs)]
 
 //! MQTT version 5.0 client library providing flexibility for decoupled asynchronous applications
 //!


rust/azure_iot_operations_mqtt/src/session/auth_policy.rs
```diff
@@ -0,0 +1,130 @@
...
+
+const SAT_AUTHENTICATION_METHOD: &str = "K8S-SAT";
+
+#[async_trait::async_trait]
+pub trait AuthPolicy: Send + Sync {
+    fn authentication_info(&self) -> AuthenticationInfo;
+
+    fn auth_challenge(&self, auth: &Auth) -> Option<Bytes>;
+
+    async fn reauth_notified(&self) -> Option<Bytes>;
```
vaavva: can you add some doc comment to these describing what they should do?

rust/azure_iot_operations_mqtt/src/session/session.rs
```diff
@@ -101,6 +109,17 @@ impl Session {
             vec![]
         };
 
+        // Create AuthPolicy if SAT file is provided
+        // TODO: This would ideally come directly from the SessionOptions instead of MQTT connection settings
+        let auth_policy = if let Some(sat_file) = options.connection_settings.sat_file.as_ref() {
+            Some(
+                Arc::new(SatAuthFileMonitor::new(std::path::PathBuf::from(sat_file)).unwrap())
```
vaavva: is this unwrap() safe, or should this be a `?`?

rust/azure_iot_operations_mqtt/src/session/session.rs
```diff
@@ -160,501 +182,339 @@ impl Session {
     /// # Errors
     /// Returns a [`SessionError`] if the session encounters a fatal error and ends.
     pub async fn run(mut self) -> Result<(), SessionError> {
-        self.state.transition_running();
```
vaavva: do we still want to call this state transition?

rust/azure_iot_operations_mqtt/src/session/reconnect_policy.rs
```diff
@@ -68,16 +82,19 @@ impl Default for ExponentialBackoffWithJitter {
...
-            Some(reconnect_delay)
+        if self.should_reconnect(prev_attempts, error) {
+            Some(self.calculate_delay(prev_attempts))
         } else {
             None
         }
     }
+
+    fn connection_loss_reconnect_delay(&self, _reason: &ConnectionLoss) -> Option<Duration> {
+        return Some(Duration::from_secs(0));
```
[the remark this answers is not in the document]
cartertinney: Correct, unless we are getting connected and then immediately being kicked repeatedly. Which, admittedly, I've seen happen before, although I suspect it may have been the fault of `rumqttc` so I'm not going to handle that (yet)

rust/azure_iot_operations_mqtt/src/session/session.rs
cartertinney: Correct, this is one of the parts that will get removed

rust/azure_iot_operations_mqtt/src/lib.rs
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-#![warn(missing_docs)]
+//#![warn(missing_docs)]
```
RLeclair: Intentional?

rust/azure_iot_operations_mqtt/src/session/session.rs
```diff
@@ -160,501 +184,339 @@ impl Session {
...
-                        break;
-                    }
-                    prev_reconnect_attempts += 1;
-                }
-                () = self.receive() => {
-                    // do anything here? If this ends, it should mean that the client has been dropped
+            let (connect_handle, disconnected_event) = connection.run_until_disconnect().await;
+            self.connect_handle = Some(connect_handle);
+            *self.disconnect_handle.lock().unwrap() = None;
+            self.reauth_handle = None;
```
RLeclair: If we set this to `None` here would we ever run into the expect call above?

RLeclair (approved):
LGTM! Left a couple of comments for understanding

PR #1092, opened 2025-11-18
[MQTT][Protocol] Flatten `session` module + update `mqtt` samples

cartertinney:
- Flattened the `session` module to not have a nested module of the same name.
- Propagated the import changes to `protocol` + METL tests (not the rest of the tests)
- Updated `azure_iot_operations_mqtt` samples.

### DIFF
--- a/rust/azure_iot_operations_mqtt/src/session.rs
+++ b/rust/azure_iot_operations_mqtt/src/session.rs
@@ -17,14 +17,10 @@
 //! client and broker ends. Note that a MQTT session can span multiple connects and disconnects to
 //! the broker.
 //!
-//! The MQTT session can be ended one of three ways:
-//! 1. The MQTT broker ends the MQTT session
-//! 2. The [`ReconnectPolicy`](crate::session::reconnect_policy::ReconnectPolicy) configured on the
+//! The MQTT session can be ended one of two ways:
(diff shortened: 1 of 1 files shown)


rust/azure_iot_operations_mqtt/src/session.rs
```diff
@@ -42,23 +42,39 @@
...
+        ConnectEnhancedAuthResult, ConnectResult, Connection, DisconnectedEvent, ReauthResult,
+    },
+    packet::{AuthProperties, ConnAck, DisconnectProperties, SessionExpiryInterval},
+};
 use thiserror::Error;
+use tokio::sync::Notify;
 
 use crate::azure_mqtt_adapter as adapter;
 use crate::error::{ClientError, ConnectError};
+pub use crate::session::managed_client::{SessionManagedClient, SessionPubReceiver};
```
vaavva: should a `pub use` be in a different section from the other `use`s?

rust/azure_iot_operations_mqtt/src/session.rs
```diff
@@ -42,23 +42,39 @@
 //! discarded. Thus, in order to guarantee that messages will not be lost, you should create the
 //! [`SessionPubReceiver`] *before* subscribing to the topic filter.
```
vaavva: validate that the above doc is still correct

PR #1103, opened 2025-11-20
[Rust][Protocol] Handle new cancelled ack token + update cache

RLeclair:
This PR:
- Adds handling of dupes to the executor when a command is in progress.
  - Modifies cache entry to handle this scenario by introducing the `InProgress` enum variant.
  - If an entry is marked as `InProgress`, the duplicate request will be ignored and its ack will queue waiting on the request to be completed.
- Minor simplifications to organization.
- Adds duplicate flag for telemetry receiver messages.
  - It is up to the application to handle duplicate telemetry receiver messages due to correlation ID not being enforced making it unviable to properly deduplicate messages.

### DIFF
--- a/rust/azure_iot_operations_protocol/src/rpc_command/executor.rs
+++ b/rust/azure_iot_operations_protocol/src/rpc_command/executor.rs
@@ -368,8 +368,8 @@ impl Cache {
                     processing_cancellation_token,
                 } => {
                     // If an entry is in progress and its processing cancellation token is cancelled
-                    // it means it timed out so it can be safely removed. If it didn't time out it
-                    // would have been converted to a Cached entry.
+                    // it means it timed out or the application dropped it, so it can be safely
+                    // removed. If it didn't time out it would have been converted to a Cached entry.
                     !processing_cancellation_token.is_cancelled()
                 }
             }
@@ -868,8 +868,7 @@ where
                             CacheLookupResult::NotFound
                         ) {
                             // This means there is an entry for this correlation ID so either we
-                            // have a cached response, the request is already being processed or it
-                            // is expired.
+                            // have a cached response or the request is already being processed.
                             break 'process_request;
                         }
 
@@ -1063,13 +1062,11 @@ where
                         }
                     }
 
-                    // Checking that command expiration time was calculated and has not
-                    // expired. If it has, we do not respond to the invoker.
-                    if let Some(command_expiration_time) = command_expiration_time
-                        && !command_expiration_time.elapsed().is_zero()
-                    {
+                    // Checking that command expiration time was calculated, if it has not we do not
+                    // respond to the invoker.
+                    let Some(command_expiration_time) = command_expiration_time else {
                         continue;
-                    }
+                    };
 
                     match response_arguments.cache_lookup_result {
                         CacheLookupResult::Cached {
@@ -1114,30 +1111,34 @@ where
                         }
                         CacheLookupResult::NotFound => {
                             // Indicates the command should be processed as an error
-                            tokio::task::spawn({
-                                let app_hlc_clone = self.application_hlc.clone();
-                                let client_clone = self.mqtt_client.clone();
-                                let cache_clone = self.cache.clone();
-                                let executor_cancellation_token_clone =
-                                    self.executor_cancellation_token.clone();
-                                async move {
-                                    tokio::select! {
-                                        () = executor_cancellation_token_clone.cancelled() => { /* executor dropped */},
-                                        () = Self::process_command(
-                                            app_hlc_clone,
-                                            client_clone,
-                                            pkid,
-                                            response_arguments,
-                                            (None, None),
-                                            cache_clone,
-                                            processing_drop_guard,
-                                        ) => {
-                                            // Finished processing command
-                                            handle_ack(ack_token, executor_cancellation_token_clone, pkid).await;
-                                        },
+
+                            // Check the command has not expired, if it has, we do not respond to the invoker.
+                            if command_expiration_time.elapsed().is_zero() {
+                                tokio::task::spawn({
+                                    let app_hlc_clone = self.application_hlc.clone();
+                                    let client_clone = self.mqtt_client.clone();
+                                    let cache_clone = self.cache.clone();
+                                    let executor_cancellation_token_clone =
+                                        self.executor_cancellation_token.clone();
+                                    async move {
+                                        tokio::select! {
+                                            () = executor_cancellation_token_clone.cancelled() => { /* executor dropped */},
+                                            () = Self::process_command(
+                                                app_hlc_clone,
+                                                client_clone,
+                                                pkid,
+                                                response_arguments,
+                                                (None, None),
+                                                cache_clone,
+                                                processing_drop_guard,
+                                            ) => {
+                                                // Finished processing command
+                                                handle_ack(ack_token, executor_cancellation_token_clone, pkid).await;
+                                            },
+                                        }
                                     }
-                                }
-                            });
+                                });
+                            }
                         }
                     }
 


rust/azure_iot_operations_protocol/src/rpc_command/executor.rs
```diff
@@ -1053,172 +1178,221 @@ where
...
         pkid: u16,
         mut response_arguments: ResponseArguments,
-        response_rx: Option<oneshot::Receiver<Response<TResp>>>,
-        completion_tx: Option<oneshot::Sender<Result<(), AIOProtocolError>>>,
+        application_channels: (
+            Option<oneshot::Receiver<Response<TResp>>>,
+            Option<oneshot::Sender<Result<(), AIOProtocolError>>>,
+        ), // TODO: Once simplified, remove this complex type
         cache: Cache,
+        _processing_drop_guard: DropGuard,
```
vaavva: does the underscore drop the value immediately though?

rust/azure_iot_operations_protocol/src/rpc_command/executor.rs
```diff
@@ -1433,7 +1610,18 @@ async fn handle_ack(
                 Ok(ack_ct) => {
                     match ack_ct.await {
                         Ok(()) => log::info!("[pkid: {pkid}] Acknowledged"),
-                        Err(e) => log::error!("[pkid: {pkid}] Ack error: {e}"),
+                        Err(e) => {
+                            match e {
+                                azure_iot_operations_mqtt::error::CompletionError::Detatched => {
```
vaavva: is this a typo in the underlying error or just here? If underlying, can you fix that in this PR if it's small?

rust/azure_iot_operations_protocol/src/rpc_command/executor.rs
```diff
@@ -261,24 +261,35 @@ struct CacheKey {
...
 /// Command Executor Cache Entry Status enum.
 ///
 /// Used to indicate the status of a cache entry.
-///
-/// Note: It is not possible for a cache entry to be in progress due to the nature of the underlying
-/// session. If a command request is received, the session will drop duplicates while the original
-/// request is being processed.
-#[derive(PartialEq, Debug)]
+#[derive(Debug)]
 enum CacheEntryStatus {
```
RLeclair: CacheLookupResult maybe?

vaavva (commented):
these may or may not all still be relevant after our discussion, but a few things I noted before then

rust/azure_iot_operations_protocol/src/rpc_command/executor.rs
```diff
@@ -316,9 +358,37 @@ impl Cache {
...
+                    // Always retain in progress entries
+                    true
+                }
+            }
+        });
         cache.insert(key, entry);
     }
+
+    /// Remove a cache entry from the cache.
+    fn remove(&self, key: &CacheKey) {
```
vaavva: should there be anything around this or where it's called for this to only be called on InProgress cache entries, or is this fine to happen on Cached entries as well since it would just clean them up slightly sooner?

rust/azure_iot_operations_protocol/src/rpc_command/executor.rs
```diff
@@ -988,30 +1075,73 @@ where
...
                             log::debug!(
                                 "[{}][pkid: {}] Duplicate request has expired",
                                 self.command_name,
                                 pkid
                             );
-                            continue;
                         }
-                        _ => {
+                        CacheLookupResult::NotFound => {
+                            // Indicates the command should be processed as an error
```
vaavva: I suppose this doesn't need to be entered as InProgress because it would evaluate the same every time (no danger of "running" twice) and it'll be put in as Cached soon enough. But is there any weirdness that could happen with a race condition where this tries to get put in the cache as Cached twice with the same key if there's a dupe? (not sure if this is a new scenario or not as well)

rust/azure_iot_operations_protocol/src/rpc_command/executor.rs
```diff
@@ -1053,172 +1178,221 @@ where
...
+                serialized_payload.payload,
+                publish_properties,
+            )
+            .await
+        {
+            Ok(publish_completion_token) => {
+                // Wait and handle puback
+                match publish_completion_token.await {
+                    Ok(puback) => {
+                        match puback.as_result() {
```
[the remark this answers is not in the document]
RLeclair: Good catch! Will change that

rust/azure_iot_operations_protocol/src/rpc_command/executor.rs
RLeclair: This is directly from the mqtt client so the fix would have to be there

rust/azure_iot_operations_protocol/src/rpc_command/executor.rs
```diff
@@ -1433,7 +1610,18 @@ async fn handle_ack(
                 Ok(ack_ct) => {
                     match ack_ct.await {
                         Ok(()) => log::info!("[pkid: {pkid}] Acknowledged"),
-                        Err(e) => log::error!("[pkid: {pkid}] Ack error: {e}"),
+                        Err(e) => {
+                            match e {
+                                azure_iot_operations_mqtt::error::CompletionError::Detatched => {
+                                    log::error!("[pkid: {pkid}] Ack error: {e}");
```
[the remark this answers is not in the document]
RLeclair: I don't think the cancelled one should be logged as error because we can still ack the request as a whole whereas the other one means we can't finish the request at all

rust/azure_iot_operations_protocol/src/rpc_command/executor.rs
```diff
@@ -988,30 +1075,73 @@ where
 
                     // Checking that command expiration time was calculated and has not
                     // expired. If it has, we do not respond to the invoker.
-                    if let Some(command_expiration_time) = command_expiration_time {
-                        if !command_expiration_time.elapsed().is_zero() {
-                            continue;
+                    if let Some(command_expiration_time) = command_expiration_time
+                        && !command_expiration_time.elapsed().is_zero()
+                    {
+                        if let Some(cached_key) = response_arguments.cached_key {
```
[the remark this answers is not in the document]
RLeclair: This is checking the expiration time of the dupe, with the new changes I've removed this remove here because it won't be necessary with the drop guard being held throughout processing

rust/azure_iot_operations_protocol/src/rpc_command/executor.rs
RLeclair: This is protected by the in progress entry, if this turned out as `NotFound` an entry for in progress would be created regardless of if it is an error or not so any dupes wouldn't get to this point. They'd go through the in progress branch

rust/azure_iot_operations_protocol/src/rpc_command/executor.rs
```diff
@@ -796,13 +861,26 @@ where
...
 
-                        // If the cache entry is not found, continue processing the request
-                        if response_arguments.cached_entry_status != CacheEntryStatus::NotFound {
+                        if !matches!(
+                            response_arguments.cache_lookup_result,
+                            CacheLookupResult::NotFound
+                        ) {
+                            // This means there is an entry for this correlation ID so either we
+                            // have a cached response, the request is already being processed or it
+                            // is expired.
```
vaavva:
```suggestion
                            // have a cached response or the request is already being processed.
```

rust/azure_iot_operations_protocol/src/rpc_command/executor.rs
```diff
@@ -1800,30 +2071,141 @@ mod tests {
...
+                assert!(range.contains(&response_message_expiry_interval));
+            }
+            _ => {
+                panic!("Expected cached entry");
+            }
+        }
+    }
+
+    #[test]
+    fn test_cache_in_progress_expired_with_different_key_set() {
```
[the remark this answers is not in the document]
RLeclair: Done!

PR #1110, opened 2025-11-24
[Rust][MQTT] Fixed plenary ack deadlock + added initial receive tests

cartertinney:


rust/azure_iot_operations_mqtt/tests/mock_receive.rs
```diff
@@ -0,0 +1,482 @@
...
+}
+
+#[tokio::test]
+async fn qos0_single_filtered_receiver() {
+    let (session, mock_server) =
+        setup_client_and_mock_server("qos0_single_filtered_receiver_test_client");
+    let managed_client = session.create_managed_client();
+
+    // Start the session run loop
+    tokio::task::spawn(session.run());
```
vaavva: does it matter to not clean up the session at the end of the test?

rust/azure_iot_operations_mqtt/tests/mock_receive.rs
```diff
@@ -0,0 +1,482 @@
...
+    mock_server.expect_connect_and_accept(true).await;
```
vaavva: for mocks, should we use "clean start true" because we aren't emulating any of the broker behavior for if the session is present?
```suggestion
    mock_server.expect_connect_and_accept(false).await;
```

rust/azure_iot_operations_mqtt/tests/mock_receive.rs
```diff
@@ -0,0 +1,482 @@
...
+    assert_pending!(ct6.poll()); // Completion token for PUBACK 6 not ready yet
+    let ct5 = acktoken5.ack().await.unwrap(); // ACK Publish 5
+    ct5.await.unwrap(); // Wait for PUBACK 5 to complete
+    let puback5 = mock_server.expect_puback_and_return().await; // PUBACK for Publish 5
+    assert_eq!(puback5.packet_identifier, 5);
+    ct6.await.unwrap(); // Wait for PUBACK 6 to complete
+    let puback6 = mock_server.expect_puback_and_return().await; // PUBACK for Publish 6
+    assert_eq!(puback6.packet_identifier, 6);
+
+    ///////////// Multiple messages (unorderd acks) using recv_manual_ack() with dropped AckTokens /////////////
```
vaavva:
```suggestion
    ///////////// Multiple messages (unordered acks) using recv_manual_ack() with dropped AckTokens /////////////
```

rust/azure_iot_operations_mqtt/tests/mock_receive.rs
```diff
@@ -0,0 +1,482 @@
...
+    let response3 = receiver3.recv_manual_ack().await.unwrap();
+    assert_eq!(response3.0, expected_publish);
+    assert!(response3.1.is_none()); // No manual ack mechanism for QoS 0
+
+    // No ack expected for QoS 0
+    mock_server.expect_no_packet();
+}
+
+#[tokio::test]
+async fn qos0_multiple_identical_filtered_receivers() {
```
vaavva: could be interesting to do a variant of this one with different wildcards as well

rust/azure_iot_operations_mqtt/tests/mock_receive.rs
```diff
@@ -0,0 +1,482 @@
...
+        ),
+        retain: false,
+        payload: bytes::Bytes::from("Publish 2"),
+        other_properties: mqtt_proto::PublishOtherProperties::default(),
+    };
+    let expected_publish2 = proto_publish2.clone().into();
+    mock_server.send_publish(proto_publish2);
+    let response = receiver.recv_manual_ack().await.unwrap();
+    assert_eq!(response.0, expected_publish2);
+    let acktoken = response.1.expect("Expected ack token for QoS 1");
```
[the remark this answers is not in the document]
cartertinney: When we get into later tests that I've excluded I think you'll see keeping things more concise will be better

PR #1111, opened 2025-11-24
[Rust][MQTT] Clippy fixes + azure_mqtt version

cartertinney:


rust/azure_iot_operations_mqtt/src/session/managed_client.rs
```diff
@@ -43,35 +51,76 @@ impl SessionManagedClient {
...
             .publish_qos0(topic, payload.into(), retain, properties)
             .await
     }
 
+    /// Issue an MQTT `PUBLISH` at Quality of Service 1 ("at least once" delivery).
+    ///
+    /// If connection is unavailable, `PUBLISH` will be queued and delivered when connection is
+    /// re-established. Blocks if at capacity for queueing.
+    ///
+    /// Returns a token that can be awaited to indicate the completion of the `PUBLISH` operation
```
vaavva: doesn't the first await correlate to the publish being sent and the second one to the puback being received? Not sure if there's anything around that that should be clarified in this comment so that folks make sure to .await twice if they want full confirmation

rust/azure_iot_operations_mqtt/src/session/managed_client.rs
```diff
@@ -93,30 +142,49 @@ impl SessionManagedClient {
             .await
     }
 
+    /// Issue an MQTT `UNSUBSCRIBE` to receive `PUBLISH`es on the provided topic filter.
```
vaavva:
```suggestion
    /// Issue an MQTT `UNSUBSCRIBE` to stop receiving `PUBLISH`es on the provided topic filter.
```

rust/azure_iot_operations_mqtt/src/session/reconnect_policy.rs
```diff
@@ -10,9 +10,13 @@ use rand::Rng;
...
 use crate::error::{ConnectError, ProtocolError};
 
-pub enum ConnectionLoss {
+/// Reason for connection loss.
+pub enum ConnectionLossReason {
+    /// Disconnected by server with DISCONNECT packet.
     DisconnectByServer(Disconnect),
+    /// Disconnected due to an I/O error.
     IoError(std::io::Error),
+    /// Disconnected due to a protocol error.
```
vaavva: or something like that, not sure if committed is too aggressive
```suggestion
    /// Disconnected due to a protocol error committed by the server.
```

rust/azure_iot_operations_mqtt/src/session.rs
```diff
@@ -536,63 +531,51 @@ impl Session {
...
-                                // NOTE: This only could really happen if an MQTT disconnect AND a reauth notification
-                                // occurr at the same time, which is extremely unlikely, and also completely harmless.
+                                // NOTE: This only could really happen if an MQTT disconnect happened while
+                                // waiting for the reauth response. Completely harmless.
                                 return;
                             }
+                        } else {
+                            log::warn!("Reauth handle detached. Exiting reauthentication monitor.");
+                            // NOTE: This only could really happen if an MQTT disconnect AND a reauth notification
+                            // occurr at the same time, which is extremely unlikely, and also completely harmless.
```
vaavva:
```suggestion
                            // occur at the same time, which is extremely unlikely, and also completely harmless.
```

vaavva (approved):

RLeclair (approved):

PR #1114, opened 2025-11-26
[Rust][MQTT] Integrated support for HL timeouts + clippy fixes

cartertinney:
- Updated to latest version of `azure_mqtt`
- Updated the use of connection timeouts and keepalive to correspond to new APIs
- Added new timeout error cases
- Fixed clippy warnings

### DIFF
--- a/rust/azure_iot_operations_mqtt/src/azure_mqtt_adapter.rs
+++ b/rust/azure_iot_operations_mqtt/src/azure_mqtt_adapter.rs
@@ -288,7 +288,7 @@ impl MqttConnectionSettings {
 
         let keep_alive = azure_mqtt::client::KeepAliveConfig::Duration {
             ping_after,
-            response_timeout: Duration::from_secs(2),
+            response_timeout: Duration::from_secs(2),   // TODO: Make configurable?
         };
 
         let password = if let Some(password_file) = self.password_file {


rust/azure_iot_operations_mqtt/src/azure_mqtt_adapter.rs
```diff
@@ -256,16 +272,24 @@ impl MqttConnectionSettings {
...
+            })?)
+            .ok_or(ConnectionSettingsAdapterError {
+                msg: "keep_alive must be greater than zero".to_string(),
+                field: ConnectionSettingsField::KeepAlive(self.keep_alive),
+                source: None,
+            })?;
+
+        let keep_alive = azure_mqtt::client::KeepAliveConfig::Duration {
+            ping_after,
+            response_timeout: Duration::from_secs(2),
```
RLeclair: Where does this 2 come from?

RLeclair (approved):

rust/azure_iot_operations_mqtt/src/azure_mqtt_adapter.rs
cartertinney: Arbitrary - we can add a setting in Session Options to pipe it through later if we care, but this was never something we configured before.

PR #1115, opened 2025-12-01
[Rust][Go][Protocols] Enable more METL test cases

vaavva:
This PR:
- fixes a broken metl test case that had the steps out of order (and un-excluded it for rust and go. Dotnet already has it excluded because it doesn't support the feature)
- support concurrent execution in executor test framework, unblocking several dispatcher feature tests
- Update skipped tests/features to enable a few more that are fixed by the new mqtt client/executor cache updates


vaavva:
> The test failures are mildly concerning, but this all looks fine to me.

I believe these come from our conversation around this PR: https://github.com/Azure/iot-operations-sdks/pull/1085

PR #1121, opened 2025-12-03
[Rust][MQTT][Protocols][Services][Connector] GA Logging cleanup

vaavva:
This PR:
- all existing log levels adjusted properly according to https://github.com/Azure/iot-operations-sdks/blob/main/doc/dev/adr/0005-logging.md
- ensured that we aren't logging any full payloads above trace
	- I left any command names, asset/etc names, DSS key names as things that were fine to log above trace as "operational data"
	- I added display impls for one DSS log that was logging the full value that was set to change it to just show whether the operation was a set or delete - realistically this is the useful information and then we don't have to worry about logging giant payloads or anything
- checked all error logs to make sure they weren't doing anything too funky beyond logging the error contents
- did a bit of cleanup around the command logs to make sure that we always logged the command name/etc.
- did a quick clean-up to try and make sure anything logged above debug has some context of what logged the error (since in the connectors, only debug logs and below show the module name)

This PR doesn't:
- add any additional logs that we are currently missing - I only evaluated what currently exists, and I didn't check for redundant/unnecessary logs
- heavily standardize the log contents (I tried to lightly do this around the command names/context, but there may still be discrepancies between the logs in the invoker/executor/telemetry/services/etc - such as no logs in the SR client)
- make sure logs are pretty/readable/useful beyond a sanity check
- make any changes to logs in samples

### DIFF
--- a/rust/azure_iot_operations_mqtt/src/session/auth_policy.rs
+++ b/rust/azure_iot_operations_mqtt/src/session/auth_policy.rs
@@ -84,6 +84,9 @@ impl SatAuthFileMonitor {
                                 Ok(data) => Bytes::from(data),
                                 Err(e) => {
                                     log::warn!("Error reading updated SAT file: {e}");
+                                    log::warn!(
+                                        "SAT file reading will be retried on next change/connection attempt."
+                                    );
                                     return;
                                 }
                             };
@@ -100,6 +103,9 @@ impl SatAuthFileMonitor {
                     }
                     Err(e) => {
                         log::warn!("Error(s) on SAT file directory debounce event: {e:?}");
+                        log::warn!(
+                            "SAT file reading will be retried on next change/connection attempt."
+                        );
                     }
                 }
             },


rust/azure_iot_operations_mqtt/src/session/auth_policy.rs
```diff
@@ -99,7 +99,7 @@ impl SatAuthFileMonitor {
                         }
                     }
                     Err(e) => {
-                        log::error!("Error(s) on SAT file directory debounce event: {e:?}");
+                        log::warn!("Error(s) on SAT file directory debounce event: {e:?}");
```
cartertinney: I do think this is a failure. Perhaps it doesn't meet the standards of our policy, so we can discuss, but intuitively, this feels like a failure, as does the above one.

rust/azure_iot_operations_protocol/src/rpc_command/executor.rs
```diff
@@ -751,8 +763,8 @@ where
                         }
                         rt
                     } else {
-                        log::error!(
-                            "[{}][pkid: {}] Response topic missing",
+                        log::warn!(
```
cartertinney: Isn't this a protocol violation?

rust/azure_iot_operations_protocol/src/rpc_command/executor.rs
```diff
@@ -1190,21 +1203,21 @@ where
                 match publish_completion_token.await {
                     Ok(puback) => {
                         if !puback.is_success() {
-                            log::error!(
-                                "[{command_name}][pkid: {pkid}] Puback error on cached response: {puback:?}"
+                            log::warn!(
+                                "[{command_name}][pkid: {pkid}] Puback error on cached command response: {puback:?}"
```
cartertinney: Maybe "puback reported failure for cached command response"? I'm not sure this is particularly clear.

rust/azure_iot_operations_protocol/src/telemetry/receiver.rs
```diff
@@ -640,7 +640,7 @@ where
                     };
 
                     // Process the received message
-                    log::info!("[pkid: {pkid}] Received message");
+                    log::debug!("[pkid: {pkid}] Received message");
```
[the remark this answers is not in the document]
vaavva: There's no unique identifier, unfortunately

rust/azure_iot_operations_protocol/src/rpc_command/executor.rs
vaavva: yes (aio protocol violation), but so are most of the other things we validate. Since this doesn't stop the executor completely, just ignores one message, this should be warn. The non-error scenario for this is a message on the same topic filter that isn't a command request at all

RLeclair (approved):

rust/azure_iot_operations_protocol/src/rpc_command/invoker.rs
```diff
@@ -1408,7 +1419,7 @@ async fn flatten<T>(
         Ok(Err(e)) => Err(e),
         Err(e) => {
             // tasks can't panic
-            log::error!("Join Error: {e}");
+            log::error!("Invoker Join Error: {e}");
```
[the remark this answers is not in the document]
vaavva: works for me!

PR #1124, opened 2025-12-04
[Rust][MQTT] Reauth cleanup

cartertinney:
Fixed several bugs where reauth task could live longer than it was intended to.

Added tests for these situations.


rust/azure_iot_operations_mqtt/tests/mock_connection_management.rs
```diff
@@ -573,3 +573,169 @@ async fn force_exit_while_disconnected() {
...
+    assert!(run_f.await.unwrap().is_ok());
+
+    // Updating the SAT file after exit does not cause any AUTH packets to be sent
+    mock_sat_file.update_contents();
+    tokio::time::sleep(Duration::from_secs(11)).await;
+    mock_server.expect_no_packet();
+}
+
+/// Thist test validates that multiple SAT file updates does not cause multiple AUTH packets
+/// to be sent if done in rapid succession.
```
vaavva: why is this the case? is it because we wouldn't evaluate new sat file contents until the current auth handshake is complete and then go again with the latest?

PR #1132, opened 2025-12-09
[.NET] Generic CloudEvent handling

maximsemenov80:
This pull request introduces a clear separation between generic and protocol-specific CloudEvent usage in the Azure IoT Operations Protocol SDK. The main change is the introduction and adoption of a new `ProtocolCloudEvent` class for protocol-managed scenarios, while the existing `CloudEvent` class is now intended solely for generic MQTT usage. This transition is reflected across the SDK, with updates to documentation, API usage, and internal logic to ensure type safety and clarify intent. Additionally, several usability improvements and documentation enhancements are included.

**CloudEvent/ProtocolCloudEvent Separation and Adoption:**

* Introduced a new `ProtocolCloudEvent` class for use in protocol-specific scenarios, where the SDK controls the `Type` and `DataContentType` properties, and updated all protocol code to use this new class instead of the generic `CloudEvent`. The generic `CloudEvent` class is now intended for user-controlled, non-protocol MQTT scenarios only. [[1]](diffhunk://#diff-97285af66eaa324b48ebf58ff24cf321de645c9fe0236e46377879de92eedd5cL6-R26) [[2]](diffhunk://#diff-a6ee23397d88983d9aa12ed7283562541e8d6297bd5144d58886a012b1b33dd4L28-R29) [[3]](diffhunk://#diff-6360266bbaff8724433371d23ac76015cc83fc59fdf3e71e664f63414626df18L31-R31) [[4]](diffhunk://#diff-c2e9c79d80fae4dac273bbf60ff9e654a016334d5f27bfc36a450b68a55356b8L17-R19) [[5]](diffhunk://#diff-3387afa51868605a18a6868bce0243b2a1803e533b9fae752749edf2a2f7828fL21-R21)
* Updated the `AioCloudEvent` class to provide a `ToProtocolCloudEvent` method for creating protocol-compliant events, and updated all relevant usages to call this new method instead of `ToCloudEvent`. [[1]](diffhunk://#diff-77b91964885689b4cd8a2cf82b04c8a9998c333a21c6c59eef9f87e48888e297R70-R88) [[2]](diffhunk://#diff-9720dd7edebff7075fd633fd304b74c99123fc25817e7c46d53399754e6dabf5L362-R362) [[3]](diffhunk://#diff-9720dd7edebff7075fd633fd304b74c99123fc25817e7c46d53399754e6dabf5L503-R503)

**Documentation and Developer Guidance:**

* Added a comprehensive [CloudEvent-Architecture.md](https://github.com/Azure/iot-operations-sdks/blob/maxim/generic-cloud-event-02/dotnet/src/Azure.Iot.Operations.Protocol/CloudEvent-Architecture.md) document describing the rationale, usage, and migration guidance for the new dual-class CloudEvent architecture, including diagrams and property control matrices.
* Enhanced XML documentation for the `CloudEvent` class to clarify its intended usage and direct protocol users to the new `ProtocolCloudEvent`.

**API Improvements and Usability:**

* Changed the `DataContentType` property on `CloudEvent` to have a public setter, making it easier for users to set this property in generic scenarios.
* Improved the `MqttApplicationMessage` API by adding helper methods for adding user properties and converting payloads to strings, and updated the `AddMetadata` method to accept nullable metadata and to use the new `SetCloudEvent` method. [[1]](diffhunk://#diff-b554073471038c0a755b5916205eb9d5dd8013190937bc731f160f4ffc33dbdbR174-R206) [[2]](diffhunk://#diff-b554073471038c0a755b5916205eb9d5dd8013190937bc731f160f4ffc33dbdbL199-R219)

**Internal SDK Access:**

* Added an `InternalsVisibleTo` attribute to allow the `Azure.Iot.Operations.Connector` assembly access to internal members of the protocol SDK, supporting the new protocol event handling.

**General Codebase Updates:**

* Updated all relevant sample and implementation files to use `ProtocolCloudEvent` instead of `CloudEvent` where appropriate, and added necessary using statements. [[1]](diffhunk://#diff-a6ee23397d88983d9aa12ed7283562541e8d6297bd5144d58886a012b1b33dd4R4) [[2]](diffhunk://#diff-77b91964885689b4cd8a2cf82b04c8a9998c333a21c6c59eef9f87e48888e297R4)

These changes collectively improve type safety, clarify developer intent, and provide a more robust and understandable eventing model for both protocol and generic MQTT scenarios.

### DIFF
--- a/dotnet/test/Azure.Iot.Operations.Protocol.IntegrationTests/MemMonEnvoyTests.cs
+++ b/dotnet/test/Azure.Iot.Operations.Protocol.IntegrationTests/MemMonEnvoyTests.cs
@@ -165,7 +165,7 @@ public class MemMonEnvoyTests
         var memStatsMD = memmonClient.ReceivedMemoryStatsTelemetryMetadata[0];
         Assert.NotNull(memStatsMD);
         Assert.NotNull(memStatsMD.UserData);
-        Assert.Equal(8, memStatsMD.UserData.Count);
+        Assert.Equal(7, memStatsMD.UserData.Count); // Only custom user data, CloudEvent is separate
         Assert.NotNull(memStatsMD.GetCloudEvent());
         Assert.Equal("1.0", memStatsMD.GetCloudEvent().SpecVersion);
         Assert.Equal("test://mq/", memStatsMD.GetCloudEvent().Source!.ToString());
@@ -180,7 +180,7 @@ public class MemMonEnvoyTests
         var ManagedMemoryMD = memmonClient.ReceivedManagedMemoryTelemetryMetadata[0];
         Assert.NotNull(ManagedMemoryMD);
         Assert.NotNull(ManagedMemoryMD.UserData);
-        Assert.Equal(8, ManagedMemoryMD.UserData.Count);
+        Assert.Equal(7, ManagedMemoryMD.UserData.Count); // Only custom user data, CloudEvent is separate
         Assert.Equal("1.0", ManagedMemoryMD.GetCloudEvent()!.SpecVersion);
         Assert.Equal("test://mq/", ManagedMemoryMD.GetCloudEvent().Source!.ToString());
         Assert.Equal("ms.aio.telemetry", ManagedMemoryMD.GetCloudEvent().Type);


timtay-microsoft (approved):

dotnet/src/Azure.Iot.Operations.Protocol/RPC/CommandResponseMetadata.cs
```diff
@@ -101,6 +113,11 @@ public void MarshalTo(MqttApplicationMessage message)
                 message.AddUserProperty(AkriSystemProperties.Timestamp, Timestamp.EncodeToString());
             }
 
+            if (CloudEvent != null)
+            {
+                message.SetCloudEvent(CloudEvent);
```
vaavva: If there's a data content type set on the cloud event in rpc, this shouldn't override the content type already set on the message by rpc - I think this would happen here since you're using the mqtt application message?

dotnet/src/Azure.Iot.Operations.Protocol/CloudEvent.cs
vaavva: now that this isn't telemetry specific, this default probably shouldn't be here, right?

dotnet/src/Azure.Iot.Operations.Protocol/RPC/CommandResponseMetadata.cs
vaavva: This either means that rpc can't use the generic implementation or there needs to be a variant exposed that rpc can use that doesn't overwrite the content type

dotnet/src/Azure.Iot.Operations.Protocol/CloudEvent.cs
vaavva: Does that default need to get added back in somewhere for telemetry though?

dotnet/src/Azure.Iot.Operations.Protocol/RPC/CommandResponseMetadata.cs
timtay-microsoft: > I thought it is what's required: image haven't I got it right? I need to update method docs though

That adr doc comment is strictly for the generic use case. For the RPC-specific use case, we should still get the content type from the serializer like Valerie pointed out. If you need an example, I believe the telemetry case already does this.

dotnet/test/Azure.Iot.Operations.Protocol.IntegrationTests/MemMonEnvoyTests.cs
```diff
@@ -129,7 +129,7 @@ public async Task Send_ReceiveTelemetryWithMetadataAndCE()
         var ManagedMemoryCorrelationId = Guid.NewGuid();
         var ManagedMemoryUserDataKey = Guid.NewGuid().ToString();
         var ManagedMemoryUserDataValue = Guid.NewGuid().ToString();
-        var ManagedMemoryTelemetryMetadata = new OutgoingTelemetryMetadata() { CloudEvent = new CloudEvent(new Uri("test://mq")) };
+        var ManagedMemoryTelemetryMetadata = new OutgoingTelemetryMetadata() { CloudEvent = new CloudEvent(new Uri("test://mq"),  "ms.aio.telemetry") };
```
vaavva: would the customer have to know this value to set it, or is this defaulted within telemetry somewhere? (might just be rusty in dotnet)

dotnet/src/Azure.Iot.Operations.Protocol/Models/MqttApplicationMessage.cs
```diff
@@ -235,6 +336,149 @@ public void AddCloudEvents(CloudEvent md)
             {
                 AddUserProperty(nameof(md.DataSchema).ToLowerInvariant(), md.DataSchema);
             }
+
+            // Override ContentType if DataContentType is set
+            if (!string.IsNullOrWhiteSpace(md.DataContentType))
```
vaavva: should this be the bit that doesn't happen on the protocol variant of this fn?

dotnet/src/Azure.Iot.Operations.Protocol/RPC/CommandExecutor.cs
```diff
@@ -475,7 +475,12 @@ private async Task<MqttApplicationMessage> GenerateResponseAsync(
             string timestamp = await _applicationContext.ApplicationHlc.UpdateNowAsync();
             message.AddUserProperty(AkriSystemProperties.Timestamp, timestamp);
 
-            EnsureCloudEventTypeAndContentType(metadata, _msAioRpcResponse, payloadContext?.ContentType);
+            if (metadata?.CloudEvent != null)
+            {
+                // Set Type and DataContentType for RPC response CloudEvents
+                metadata.CloudEvent.Type = _msAioRpcResponse;
```
vaavva: this should only be set if the customer didn't set a type (it should be a default, not an override)

dotnet/src/Azure.Iot.Operations.Protocol/RPC/CommandRequestMetadata.cs
```diff
@@ -49,6 +49,14 @@ public class CommandRequestMetadata
         /// </summary>
         public string? Partition { get; }
 
+        /// <summary>
+        /// CloudEvent metadata for this RPC request. The Type will be automatically set to "ms.aio.rpc.request"
```
vaavva:
```suggestion
        /// CloudEvent metadata for this RPC request. The Type will be default to "ms.aio.rpc.request" if not set
```

dotnet/src/Azure.Iot.Operations.Protocol/RPC/CommandResponseMetadata.cs
```diff
@@ -47,11 +47,12 @@ public class CommandResponseMetadata
         public Dictionary<string, string> UserData { get; set; } = new();
 
         /// <summary>
-        /// The CloudEvent metadata attached to the response.
+        /// CloudEvent metadata for this RPC response. The Type will be automatically set to "ms.aio.rpc.response"
```
vaavva: is this struct for the sending and receiving side? If so, might want to have some clarification that this is only for the executor side of things. On the receiving side, the Type would be set from the user properties and the DataContentType would be set from the publish properties as well

dotnet/src/Azure.Iot.Operations.Protocol/CloudEvent-Architecture.md
```diff
@@ -0,0 +1,225 @@
...
+        +ProtocolCloudEvent? GetProtocolCloudEvent() internal
+    }
+    
+    class OutgoingTelemetryMetadata {
+        +ProtocolCloudEvent? CloudEvent
+        +Dictionary~string,string~ UserData
+    }
+    
+    class IncomingTelemetryMetadata {
+        +ProtocolCloudEvent GetCloudEvent()
```
vaavva: Should there be equivalent GetCloudEvent fns on the command metadatas, or is there a difference that makes it not needed?

dotnet/src/Azure.Iot.Operations.Protocol/CloudEvent-Architecture.md
```diff
@@ -0,0 +1,225 @@
...
+    participant Sender as TelemetrySender
+    participant MQTT as MqttApplicationMessage
+    
+    User->>PCE: new ProtocolCloudEvent(source)
+    Note over PCE: Type = "ms.aio.telemetry" (default)
+    User->>PCE: Set optional properties (Subject, DataSchema)
+    User->>Meta: metadata.CloudEvent = pce
+    User->>Sender: SendTelemetryAsync(payload, metadata)
+    Sender->>PCE: Set DataContentType (from serializer)
+    Note over Sender: SDK controls Type and DataContentType
```
vaavva: this is only true for the sending direction. When receiving, this is treated the same as the generic scenario

PR #1143, opened 2025-12-10
[Rust][MQTT] Session error revisions

cartertinney:
* Added handling for error cases which were previously `unwrap()` and `expect()`
* Added a `source` field on `SessionError` to allow for reporting of cert errors during run
* Added `#[non_exhaustive]` for "Kind" enums
* Added a new variant to `SessionError`
* Fixed various documentation issues
* Ensured all errors are exported correctly
* Source errors are now `Send` so they can cross thread boundaries
* Fixed incorrect logs around `SessionExitError` handling
* Removed the now-unnecessary `attempted` boolean (and its corresponding accessor function) from `SessionExitError`

### DIFF
--- a/doc/dev/generated/MetlCases.md
+++ b/doc/dev/generated/MetlCases.md
@@ -28,119 +28,119 @@ See below for categorized tests.
 
 | Normative statement | Expected behavior |
 | --- | --- |
-| CommandExecutor request topic contains a '{modelId}' token but no model ID is specified. | CommandExecutor starts successfully. |
-| CommandExecutor receives request with a protocol version that it does not support. | CommandExecutor sends response with status NotSupportedVersion. |
-| CommandExecutor receives request with no __srcId header. | CommandExecutor sends response with status OK. |
-| CommandExecutor receives basic valid request. | CommandExecutor sends response and acknowledges request. |
-| CommandExecutor receives request with different topic than subscribed. | CommandExecutor ignores request, and MQTT client auto-acknowledges. |
-| CommandExecutor receives valid request but ACK dropped when publishing response. | Connection automatically re-established, publication retried, success. |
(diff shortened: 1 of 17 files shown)


rust/azure_iot_operations_mqtt/src/session.rs
```diff
@@ -115,11 +135,11 @@ impl From<SessionErrorKind> for SessionError {
 #[error(transparent)]
 pub struct SessionConfigError(#[from] adapter::ConnectionSettingsAdapterError);
 
+// NOTE: Retain a struct/kind pattern for `SessionExitError` for future-proofing
```
cartertinney: `.attempted` could become relevant again in the future and be re-added if, for instance, we needed to have a "wait for connect and then end session" API of some kind. This would also potentially add new error cases, thus why the `SessionExitErrorKind` is non-exhaustive. I don't think these are likely scenarios, but it seems safer to not take too strong a position on exit errors until we see how customers use it. Plus, there's a symmetry with all "kinds" being non-exhaustive.

rust/azure_iot_operations_mqtt/src/session.rs
```diff
@@ -705,27 +727,34 @@ impl SessionExitHandle {
...
-            false
+        log::debug!("Attempting to exit session gracefully");
+        match self.try_exit() {
+            Ok(()) => {
+                log::debug!("Session exited gracefully");
+                true
+            }
+            Err(e) => {
+                if e.kind() == SessionExitErrorKind::Detached {
+                    log::debug!("Session already detached, no action needed");
```
vaavva: this case is safe to not notify the force_exit because this would only be the case if the task that this cleans up is already cleaned up, right?

PR #1144, opened 2025-12-10
[Rust][Connector] Update connector cloud event headers

vaavva:
This PR:
- updates the cloud event headers populated by the Connector SDK to match cross-AIO patterns
- adds a new API to allow the customer to provide a protocol specific identifier to use in the `source` header. I added this as an additional function to give us more flexibility in the future if we add more cloud event header customizations (at that point, the base forward_data fn without cloud event header arguments and a much more custom fn might be the two most desired functions and the one with just protocol specific identifier specified might not be used much anymore, so it seemed better to have the base forward_data as an existing API)
- added tests to validate building the headers with the new patterns

### DIFF
--- a/.github/workflows/ci-dotnet.yml
+++ b/.github/workflows/ci-dotnet.yml
@@ -150,7 +150,7 @@ jobs:
       run: dotnet test dotnet/Azure.Iot.Operations.sln -e MQTT_TEST_BROKER_CS="HostName=127.0.0.1;TcpPort=8884;UseTls=true;ClientId=TestClient;CaFile=${{ env.CA_FILE_PATH }};SatAuthFile=${{ env.TOKEN_FILE_PATH }}" -e FAULTABLE_MQTT_TEST_BROKER_CS="HostName=localhost;TcpPort=1884;UseTls=false;ClientId=TestClient" --collect:"XPlat Code Coverage" --logger:"trx;LogFileName=test-results.trx" -- xunit.parallelizeAssembly=true
 
     - name: Test SDK (TLS + x509)
-      run: dotnet test dotnet/Azure.Iot.Operations.sln --filter "FullyQualifiedName~Azure.Iot.Operations.Services.IntegrationTests.LeaderElectionClientIntegrationTests.TestFencing" -e MQTT_TEST_BROKER_CS="HostName=127.0.0.1;TcpPort=8883;UseTls=true;ClientId=TestClient;CaFile=${{ env.CA_FILE_PATH }};KeyFile=${{ env.CLIENT_KEY }};CertFile=${{ env.CLIENT_CERT }}" --logger:"trx;LogFileName=test-results.trx" -- xunit.parallelizeAssembly=true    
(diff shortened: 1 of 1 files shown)


rust/azure_iot_operations_connector/src/destination_endpoint.rs
```diff
@@ -57,6 +65,11 @@ pub enum ErrorKind {
 pub(crate) struct Forwarder {
     message_schema_reference: Option<adr_models::MessageSchemaReference>,
     destination: ForwarderDestination,
+    device_uuid: Option<String>,
```
cartertinney: No change really needed here, but consider `Option<'static &str>` in the future, especially for internal use. It makes certain things significantly smoother if there's no need for mutability.

rust/azure_iot_operations_connector/src/destination_endpoint.rs
```diff
@@ -283,6 +315,202 @@ impl Forwarder {
...
+        asset_ref: &AssetRef,
+        asset_uuid: Option<&String>,
+        asset_external_asset_id: Option<&String>,
+        data_timestamp: Option<HybridLogicalClock>,
+        protocol_specific_identifier: Option<String>,
+    ) -> Result<telemetry::sender::CloudEvent, Error> {
+        // TODO: remove once message schema validation is turned back on
+        #[allow(clippy::manual_map)]
+        let message_schema_uri =
+            if let Some(message_schema_reference) = &self.message_schema_reference {
```
cartertinney: Why not use `map` here?

rust/azure_iot_operations_connector/src/destination_endpoint.rs
```diff
@@ -283,6 +315,202 @@ impl Forwarder {
...
+                Some(format!(
+                    "aio-sr://{}/{}:{}",
+                    message_schema_reference.registry_namespace,
+                    message_schema_reference.name,
+                    message_schema_reference.version
+                ))
+            } else {
+                // TODO: validate this for other destinations as well
```
cartertinney: Remove? Or is this for a follow-up?

rust/azure_iot_operations_connector/src/destination_endpoint.rs
```diff
@@ -283,6 +315,202 @@ impl Forwarder {
...
+        source
+    }
+
+    /// Creates both the (event type, subject) fields for a cloud event header.
+    /// Event type Format: [“DataSet”|”Event”]/<typeref-property-value>
+    /// Subject Format:
+    ///     <Asset-CompoundKey>|<Asset-ExternalAssetId*>|<AssetName>/<DataSet-Name>|<EventGroup-Name>[/Sub-Subject]
+    ///     *Asset-externalAssetId should only be used if different from `AssetUUID`
+    ///     `Asset-CompoundKey` currently doesn't exist, but may in the future
+    ///     `Sub-Subject` name is the event name
```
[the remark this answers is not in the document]
vaavva: Agreed, but I want to keep this as close to how the original requirements are documented, so I'll leave this one as-is for now

rust/azure_iot_operations_connector/src/destination_endpoint.rs
```diff
@@ -283,6 +315,202 @@ impl Forwarder {
...
+    /// Creates the source field for a cloud event header. Format:
+    /// ms-aio:<Device-CompoundKey>|<ProtocolSpecificIdentifier>|<Device-externaldeviceId*>|<Device-Name>|[/Sub-Source] (Sub-Source is the dataSource on the event or dataset)
+    ///     * Device-externalDeviceId should only be used if different from `DeviceUUID`
+    ///     `Device-CompoundKey` currently doesn't exist, but may in the future
+    ///     `Sub-Source` is the dataSource on the event or dataset
+    fn cloud_event_header_source(
+        asset_ref: &AssetRef,
+        protocol_specific_identifier: Option<String>,
+        device_uuid: Option<&String>,
+        device_external_device_id: Option<&String>,
```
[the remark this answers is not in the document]
vaavva: updated! They'll all start originally as strings, but better to optimize as high up as we can

PR #1154, opened 2025-12-12
[.NET] Implement generic cloud event handling in MQTT protocol

maximsemenov80:
Implementation of the "ADR27: Generic usage of cloud event headers"

### DIFF
--- a/dotnet/test/Azure.Iot.Operations.Mqtt.IntegrationTests/OrderedAckMqttClientIntegrationTests.cs
+++ b/dotnet/test/Azure.Iot.Operations.Mqtt.IntegrationTests/OrderedAckMqttClientIntegrationTests.cs
@@ -148,6 +148,8 @@ public class OrderedAckMqttClientIntegrationTests
         await mqttClient.DisconnectAsync();
     }
 
+    // Test sending an MQTT message using our Cloud Event get/set methods. Validate that
+    // the sent cloud event looks identical to the received cloud event.
     [Fact]
     public async Task TestGenericCloudEventUsage()
     {


dotnet/src/Azure.Iot.Operations.Protocol/RPC/CommandInvoker.cs
```diff
@@ -575,6 +576,25 @@ public async Task<ExtendedResponse<TResp>> InvokeCommandAsync(TReq request, Comm
                     requestMessage.ContentType = payloadContext.ContentType;
                 }
 
+                if (metadata?.CloudEvent is not null)
+                {
+                    // if type has default value then user did not set it, so set it to rpc.request
+                    if (metadata.CloudEvent.Type == "ms.aio.telemetry")
+                    {
+                        var newCloudEvent = new CloudEvent(metadata.CloudEvent.Source, "ms.aio.rpc.request", metadata.CloudEvent.SpecVersion)
```
timtay-microsoft: Cloning like this usually results in bugs later if we add more fields to the CloudEvent type (because we likely forget to update lines 586 to 589 to include copying over the new fields as well).

In cases like this, it is fine to just alter the object that the user passed in like

```csharp
metadata.CloudEvent.Type ??= "ms.aio.rpc.request"
```

Same for assigning the default values in the telemetry sender + command executor

dotnet/src/Azure.Iot.Operations.Protocol/RPC/CommandInvoker.cs
```diff
@@ -575,6 +576,25 @@ public async Task<ExtendedResponse<TResp>> InvokeCommandAsync(TReq request, Comm
...
+                            Id = metadata.CloudEvent.Id,
+                            Time = metadata.CloudEvent.Time,
+                            Subject = metadata.CloudEvent.Subject,
+                            DataSchema = metadata.CloudEvent.DataSchema
+                        };
+                        metadata.CloudEvent = newCloudEvent;
+                    }
+
+                    metadata.CloudEvent.Id ??= Guid.NewGuid().ToString();
+                    metadata.CloudEvent.Time ??= DateTime.UtcNow;
```
vaavva: With the requirement that the customer should be able to set all fields, they should also be able to specify that optional fields aren't sent rather than set to the default, but this might fall into a language difference of best practices in .net

dotnet/src/Azure.Iot.Operations.Protocol/ExtendedCloudEvent.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// </summary>
+        public string? DataContentType { get; set; }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ExtendedCloudEvent"/> class.
+        /// </summary>
+        /// <param name="source">The source of the event.</param>
+        /// <param name="type">The type of the event.</param>
+        /// <param name="specversion">The version of the CloudEvents specification.</param>
+        public ExtendedCloudEvent(Uri source, string type = "ms.aio.telemetry", string specversion = "1.0")
```
vaavva: This shouldn't default to telemetry in the generic case

dotnet/src/Azure.Iot.Operations.Protocol/Models/MqttApplicationMessage.cs
```diff
@@ -183,6 +183,72 @@ public void AddUserProperty(string key, string value)
...
+            }
+
+            ExtendedCloudEvent cloudEvent = new(source!, type!, specversion!)
+            {
+                DataContentType = ContentType
+            };
+
+            if (UserProperties.TryGetProperty("id", out string? id))
+            {
+                cloudEvent.Id = id;
```
vaavva: Id is required - this should return null if it isn't present

dotnet/src/Azure.Iot.Operations.Protocol/Models/MqttApplicationMessage.cs
```diff
@@ -183,6 +183,72 @@ public void AddUserProperty(string key, string value)
...
+            }
+
+            if (!UserProperties.TryGetProperty("type", out string? type))
+            {
+                return null;
+            }
+
+            ExtendedCloudEvent cloudEvent = new(source!, type!, specversion!)
+            {
+                DataContentType = ContentType
```
vaavva: I might be missing this, since it's probably somewhere else, but there's validation somewhere that this matches the required regex format if it's present, right?

dotnet/src/Azure.Iot.Operations.Protocol/RPC/CommandInvoker.cs
maximsemenov80: yep, that fancy syntax ??= does that: "set if not set"

dotnet/src/Azure.Iot.Operations.Protocol/RPC/CommandResponseMetadata.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using Azure.Iot.Operations.Protocol.Models;
+using Azure.Iot.Operations.Protocol.Telemetry;
```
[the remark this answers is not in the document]
maximsemenov80: artifact left from the time CE was in the Telemetry namespace.

dotnet/src/Azure.Iot.Operations.Protocol/RPC/CommandInvoker.cs
vaavva: right, but with this, there isn't a way for them to specify that they _don't_ want there to be a time on the cloud event, right?

dotnet/src/Azure.Iot.Operations.Protocol/Telemetry/IncomingTelemetryMetadata.cs
```diff
@@ -109,60 +111,5 @@ internal IncomingTelemetryMetadata(MqttApplicationMessage message, uint packetId
...
-                                p => p.Key.Equals(name.ToLowerInvariant(),
-                                StringComparison.OrdinalIgnoreCase)).Value ?? string.Empty;
-            }
-
-            string specVersion = safeGetUserProperty(nameof(CloudEvent.SpecVersion).ToLowerInvariant());
-
-
-            if (!specVersion.Equals("1.0", StringComparison.Ordinal))
-            {
-                throw new ArgumentException($"Could not parse cloud event from telemetry: Only version 1.0 supported. Version provided: {specVersion}");
```
vaavva: is there any way for the customer to get this detail on why the cloud event wasn't valid anymore? And is all of the validation that was present here still present on the generic one?

dotnet/src/Azure.Iot.Operations.Protocol/RPC/CommandRequestMetadata.cs
```diff
@@ -140,6 +144,17 @@ internal void MarshalTo(MqttApplicationMessage message)
                 message.AddUserProperty("$partition", Partition);
             }
 
+            if (CloudEvent != null)
+            {
+                message.AddCloudEvents(CloudEvent);
+                if (CloudEvent.Type.IsNullOrEmpty())
+                {
+                    CloudEvent.Type = "ms.aio.rpc.request";
```
timtay-microsoft: This value isn't being set on the MQTT message like this. What you want is something like:

```csharp
                    message.UserProperties.Add("<cloud event header name for type>", "ms.aio.rpc.request");
```

maximsemenov80: sorry can't think straight in a rush

dotnet/src/Azure.Iot.Operations.Protocol/Models/MqttApplicationMessage.cs
maximsemenov80: agree to implement, but I don't think it is a blocker now :)

PR #1164, opened 2025-12-16
[Rust][Services] Updating defaultEventsDestination to defaultDestinations for event group

RLeclair:
This PR updates the ADR service dtdl and makes changes to the default destination naming of the event group struct.


rust/azure_iot_operations_services/src/azure_device_registry/models/asset.rs
```diff
@@ -894,7 +894,7 @@ impl From<DiscoveredEventGroup> for base_client_gen::DiscoveredAssetEventGroup {
     fn from(value: DiscoveredEventGroup) -> Self {
         base_client_gen::DiscoveredAssetEventGroup {
             events: value.events.option_vec_into(),
-            default_events_destinations: value.default_events_destinations.option_vec_into(),
+            default_destinations: value.default_events_destinations.option_vec_into(),
```
vaavva: I think there will also need to be an update in managed_azure_device_registry in the connectors crate to update to the new naming

PR #1166, opened 2025-12-17
[.NET] Cloud event fixes

timtay-microsoft:
- Validate data content type satisfies regex
- Validate type is non-null and non-empty if provided, but allow default null value to signal to telemetry/rpc layer to inject their specific default values
- Validate data schema is a uri if not null/whitespace
- Validate spec version is not an empty string
- Remove an incorrect header doc
  - this setCloudEvent function does not touch the content type of a message
- Only parse cloud events once on the receiving end of an RPC message
- Make ID default to a guid in the cloud event constructor. Only allow user to set values that are non-null and non-whitespace
  - ID was set to a guid by default in a few different places, but this change centralizes that

### DIFF
--- a/rust/sample_applications/sample_connector_scaffolding/Cargo.toml
+++ b/rust/sample_applications/sample_connector_scaffolding/Cargo.toml
@@ -3,13 +3,13 @@
 
 [package]
 name = "sample_connector_scaffolding"
-version = "0.2.0"
+version = "0.3.0"
 edition = "2024"
 
 [dependencies]
-azure_iot_operations_connector = { version = "0.5.3", registry = "aio-sdks"  }
-azure_iot_operations_otel = { version = "0.2.1", registry = "aio-sdks" }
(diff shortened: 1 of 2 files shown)


dotnet/src/Azure.Iot.Operations.Protocol/CloudEvent.cs
```diff
@@ -27,21 +27,75 @@ public class CloudEvent(Uri source, string type = "", string specversion = "1.0"
...
+                if (string.IsNullOrWhiteSpace(value))
+                {
+                    throw new ArgumentException("Id cannot be null or whitespace");
+                }
+
+                _id = value;
+            }
+        }
+
+        private string _id = Guid.NewGuid().ToString();
```
vaavva: Now that this default is here like this, does Id still need to be optional, or can it be required for extra safety?

timtay-microsoft: I'd like to make it required, but that would cause a breaking change at this point

PR #1183, opened 2026-01-13
Merge codegen2 branch into main

jrdouceur:


timtay-microsoft:
Apologies for being out of the loop here, but will this new code gen project support ingesting DTDL still? If so, is it in a state where I could try it out on existing DTDL (schema registry, DSS, etc)?

jrdouceur:
> Apologies for being out of the loop here, but will this new code gen project support ingesting DTDL still? If so, is it in a state where I could try it out on existing DTDL (schema registry, DSS, etc)?

Indirectly, yes.  The Dtdl2Wot tool will accept DTDL models and convert them to WoT models, which can then be passed to the WoT ProtocolCompiler.  In the `codegen2/eval/aio` folder, the `conv.sh` script runs Dtdl2Wot on the AIO service DTDL models to yield WoT Thing Models.  You can then run `gen.sh` in the same folder to execute the new WoT ProtocolCompiler.

At that point, you can diff the folders `codegen2/eval/aio/eng/dtdl/FromDtdl` and `codegen2/eval/aio/eng/dtdl/FromWot` to view the differences in generated code.

codegen2/README.md
```diff
@@ -0,0 +1,11 @@
+# codegen2
```
avishekpant: nitpick: should we name it wot-codegen instead?

codegen2/src/Azure.Iot.Operations.EnvoyGenerator/rust/Actions/t4/RustCommandInvoker.tt
```diff
@@ -0,0 +1,262 @@
...
+        self.0.shutdown().await
+    }
+<# if (this.errorResultName != null && this.normalRequiredFields.Any()) { #>
+
+    fn get_err(field_name: &str) -> AIOProtocolError {
+        AIOProtocolError {
+            message: Some(format!("Command response missing field {field_name}")),
+            kind: AIOProtocolErrorKind::PayloadInvalid,
+            is_shallow: false,
+            is_remote: false,
```
vaavva: this used to be slightly different - what is the reason for the change?
```suggestion
            message: Some("Command response has neither normal nor error payload content".to_string() ),
                        kind: AIOProtocolErrorKind::PayloadInvalid,
                        is_shallow: false,
                        is_remote: true,
```

jrdouceur: There are a couple of changes here.

First, command (action) responses are now able to support outputs with multiple return values, each of which is a separate field in the returned object.  Each value may or may not be required, depending on what the WoT model specifies.  Previously, when there was only a single return value, if the model specified that the value was required and if there was no error response, a reasonable message was, "Command response has neither normal nor error payload".  Now that there may be multiple fields, if a required field is not present, we need a more general message, i.e., "Command response missing field", since some other field might be present, but the absence of the required field is still an error.

Second, the change in `is_remote` from `true` to `false` is a bug fix.  This is not a remotely detected error condition that is being relayed, but rather a locally detected error condition caused by an invalid payload sent from the remote end.

jrdouceur:
> But then in the WoT output, the description changes

I haven't looked into this specific change, but I can hazard a guess.  One of WoT's deficiencies (IMO) is that it does not distinguish between a `description` of an object field and a `description` of the type of the field value.  For example, if you write this:

```json
{
  "type": "object",
  "title": "Foobar",
  "properties": {
    "foo": {
      "description": "This is a bit ambiguous",
      "type": "object",
      "title": "Barfoo",
      "properties": {
        "alpha": { "type": "string" },
        "beta": { "type": "string" }
      }
    }
  }
}
```

Does the description "This is a bit ambiguous" apply to field `foo` of object `Foobar`?  Or does it apply to the object type `Barfoo`?  It's not clear.

DTDL offers a `description` property on each object field and on the overall object type.  I was faced with the decision of how to map these to the impoverished `description` properties available in WoT.  I don't think there is a right choice here, but I tried to preserve as much information as I could.  The result is that you see situations like this one.

timtay-microsoft:
When generating code with this new protocol compiler and with a specified namespace, it appears the csharp namespaces are forcibly camelCased even when CamelCase was provided:

```
./Azure.Iot.Operations.ProtocolCompiler.exe `
  --things C:\Users\timtay\source\iot\iot-operations-sdks\eng\wot\schemaregistry\SchemaRegistry.TM.json `
  --lang csharp `
  --outDir C:\Users\timtay\source\iot\iot-operations-sdks\dotnet\src\Azure.Iot.Operations.Services\SchemaRegistry\Generated `
  --noProj `
  --namespace Azure.Iot.Operations.Services.SchemaRegistry.Generated `
  --typeNamer C:\Users\timtay\source\iot\iot-operations-sdks\eng\wot\schemaregistry\SchemaNames.json
```

```csharp
/* Code generated by Azure.Iot.Operations.ProtocolCompilerLib v0.10.0.0; DO NOT EDIT. */

#nullable enable

namespace Generated.Azure.iot.operations.services.schemaRegistry.generated
{
   ...
}

```

jrdouceur:
> it appears the csharp namespaces are forcibly camelCased even when CamelCase was provided:

This is a weakness in the name converter that I can fix.  I was not anticipating that a submitted namespace would contain dots, so the current rule for generating PascalCase gets confused.  The way I have previously seen our namespaces used is like this:

There would be a folder named "Azure.Iot.Operations.Services.SchemaRegistry", which would contain the project.  The top-level namespace for the project would match the folder name, "Azure.Iot.Operations.Services.SchemaRegistry".  Within the project, a single additional level of namespace would be used for generated code, with a name like "Generated", causing the nested namespace to be "Azure.Iot.Operations.Services.SchemaRegistry.Generated", and causing the nested classes to be in a sub-folder named "Generated".

jrdouceur:
> This still seems a bit off though:

Your outdir is "Azure.Iot.Operations.Services\SchemaRegistry".  Try replacing the backslash with a dot.  It is the folder name that indicates the namespace.

timtay-microsoft:
> > This still seems a bit off though:
> 
> Your outdir is "Azure.Iot.Operations.Services\SchemaRegistry". Try replacing the backslash with a dot. It is the folder name that indicates the namespace.

That doesn't work either, though. It outputs files with the base namespace I want ("Azure.Iot.Operations.Services.SchemaRegistry") but it outputs them into the folder

```
"src/Azure.Iot.Operations.Services.SchemaRegistry"
```

I don't believe this matches the .NET convention of having a Nuget package name like "Azure.Iot.Operations.Services" wherein you have folders for each subnamespace like

```
"src/Azure.IotOperations.Services/SchemaRegistry"
"src/Azure.IotOperations.Services/AzureDeviceRegistry"
"src/Azure.IotOperations.Services/StateStore"
```

jrdouceur:
> This would result in a single library instead of 3 libraries, so maybe not what you want.

Actually, when I look in the "dotnet" folder of the repo, this is exactly what I see:  A single Azure.Iot.Operations.Services.csproj file in folder `dotnet\src\Azure.Iot.Operations.Services`, with sub-folders for `SchemaRegistry`, `StateStore`, etc.

jrdouceur:
> Yeah, this is what I'm trying to match to.

So if you just change `--namespace SchemaRegistry.Generated` to `--namespace SchemaRegistry`, does that not get you what you want?

jrdouceur:
> > Yeah, this is what I'm trying to match to.
> 
> So if you just change `--namespace SchemaRegistry.Generated` to `--namespace SchemaRegistry`, does that not get you what you want?

BTW, if this does work, I should probably add a section to the [README.md](https://github.com/Azure/iot-operations-sdks/blob/johndo/codegen2/codegen2/README.md) that explains how to use the folder name and the `--namespace` option to yield this canonical pattern.

timtay-microsoft:
Okay things look better to me now. I can run a command like

```powershell
./Azure.Iot.Operations.ProtocolCompiler.exe `
  --things C:\Users\timtay\source\iot\iot-operations-sdks\eng\wot\schemaregistry\SchemaRegistry.TM.json `
  --lang csharp `
  --outDir C:\Users\timtay\source\iot\iot-operations-sdks\dotnet\src\Azure.Iot.Operations.Services `
  --noProj `
  --namespace GeneratedTypes `
  --common GeneratedCommon `
  --typeNamer C:\Users\timtay\source\iot\iot-operations-sdks\eng\wot\schemaregistry\SchemaNames.json
```

And it outputs a folder structure like

```
Azure.Iot.Operations.Services/
  GeneratedCommon/
  GeneratedTypes/
```

Where files in the GeneratedCommon folder have a namespace of ```Azure.Iot.Operations.Services.GeneratedCommon```

and files in GeneratedTypes have a namespace of ```Azure.Iot.Operations.Services.GeneratedTypes```

timtay-microsoft:
Overall, LGTM on the C# side now

jrdouceur:
@vaavva, do you want to take another look before I merge?  Or does your previous approval stand?

PR #1191, opened 2026-01-16
[Rust][Connector] Adding health event reporting functionality to connector

RLeclair:
This PR:
- Adds new APIs for reporting health events
     - RuntimeHealthEvent - Struct for reporting health
     - report_health_event() - Report health using snapshotted version
     - pause_health_reporting() - Stop background periodic reporting
     - refresh_health_version() - Update version snapshot
     - pause_and_refresh_health_version() - Combined pause + refresh for handling updates
- When a health event is reported it gets re-reported in the background at a user-specified interval
- The health event background task also does deduplication to skip redundant reports for the same event

### DIFF
--- a/rust/azure_iot_operations_connector/examples/base_connector_sample.rs
+++ b/rust/azure_iot_operations_connector/examples/base_connector_sample.rs
@@ -26,7 +26,7 @@ use azure_iot_operations_connector::{
     deployment_artifacts::connector::ConnectorArtifacts,
 };
 use azure_iot_operations_protocol::application::ApplicationContextBuilder;
-use azure_iot_operations_services::azure_device_registry::{self, HealthStatus};
+use azure_iot_operations_services::azure_device_registry;
 
 /// Only reports status on first time (None) and when changing from OK to Error.
 /// Skips reporting when status has already been reported and hasn't changed.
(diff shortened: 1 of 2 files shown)


rust/azure_iot_operations_connector/src/base_connector/health_event.rs
```diff
@@ -0,0 +1,244 @@
...
+            }
+        }
+        // report current or new status
+        if let Some(ref curr_status) = current_status {
+            match component
+                .report_health_event(&connector_context, curr_status.clone())
+                .await
+            {
+                Ok(()) => {
+                    log::debug!("Reported health event: {curr_status:?}");
```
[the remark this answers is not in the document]
RLeclair: Addressed in #1193

rust/azure_iot_operations_connector/examples/base_connector_sample.rs
```diff
@@ -380,6 +398,11 @@ async fn run_dataset(log_identifier: String, mut data_operation_client: DataOper
                     }
                     Err(e) => {
                         log::error!("{log_identifier} Error reporting message schema: {e}");
+                        data_operation_reporter.report_health_event(RuntimeHealthEvent {
+                            message: None,
+                            reason_code: Some("SchemaReportErr".to_string()),
```
vadim-kovalyov: two things here:
- can we write a user friendly message with details?
- we should follow guidelines for reason codes, but since it is a base connector, we can't just prefix it with `ConnectorSchemaReportError`

rust/azure_iot_operations_connector/examples/base_connector_sample.rs
```diff
@@ -390,8 +413,20 @@ async fn run_dataset(log_identifier: String, mut data_operation_client: DataOper
...
+                            status: HealthStatus::Available,
+                        });
                     }
-                    Err(e) => log::error!("{log_identifier} error forwarding data: {e}"),
+                    Err(e) => {
+                        log::error!("{log_identifier} error forwarding data: {e}");
+                        data_operation_reporter.report_health_event(RuntimeHealthEvent {
+                            message: None,
+                            reason_code: Some("DataForwardErr".to_string()),
+                            status: HealthStatus::Unavailable,
```
vadim-kovalyov: same here

PR #1196, opened 2026-01-21
[DO NOT MERGE] New rust codegen output

vaavva:
for reference on what changes are present to the output. Opened this as a PR so we can comment on anything specific
I also included the generated WoT files as reference

### DIFF
--- a/eng/test/schema-samples/Counter.TM.json
+++ b/eng/test/schema-samples/Counter.TM.json
@@ -1,0 +1,102 @@
+{
+  "@context": [
+    "https://www.w3.org/2022/wot/td/v1.1",
+    { "dtv": "http://azure.com/DigitalTwins/dtmi#" }
+  ],
+  "@type": "tm:ThingModel",
+  "title": "Counter",
+  "links": [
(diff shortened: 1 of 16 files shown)


codegen2/src/Azure.Iot.Operations.EnvoyGenerator/rust/Actions/t4/RustCommandExecutor.cs
cartertinney: Why is this commented?

rust/azure_iot_operations_services/src/azure_device_registry/adr_base_gen/adr_base_service/code_schema.rs
```diff
@@ -2,16 +2,12 @@
 
 use serde::{Deserialize, Serialize};
 
+/// The error code that identifies the error.
```
cartertinney: I realize this is generated from a document, but why does it call this a "code"?

RLeclair (dismissed):

PR #1198, opened 2026-01-23
Add ADR 29, WoT Platform Binding for AIO

jrdouceur:


jrdouceur:
Hi @avishekpant, a reminder that I'll need your approval to check in this ADR, which describes the AIO-specific properties we are using in our WoT models.  These properties are accepted by the new ProtocolCompiler, so it would be good to have this document in main now that the new ProtocolCompiler is in main.

timtay-microsoft (approved):

avishekpant (approved):

PR #1203, opened 2026-02-04
Fix missing DataSource serialization in DiscoveredAssetManagementGroup.ToProtocol()

koepalex:
This pull request introduces a minor update to the protocol conversion logic for asset management groups. The main change is the inclusion of the `DataSource` property when converting to the protocol model, ensuring that this information is now correctly transferred.

* Protocol conversion update:
  * Added mapping of the `DataSource` property from the source object to the protocol object in the `ToProtocol` method in `ProtocolConverter.cs`.


timtay-microsoft:
Closing in favor of #1204

PR #1261, opened 2026-03-03
[.NET][Rust] Target newest broker + Akri service bits

timtay-microsoft:
This should allow us to test the new event groups destination fix

Notable changes in Akri behavior from this update:

- When invoking ADR service client APIs, connector client Id is now expected to follow the pattern "```<name>-ss-<pod ordinal>```" (like "mockconnector-ctemplate-00000000-ss-0")
  - This broke a few integration .NET tests because our connector client id used to test the ADR service client didn't adhere to this shape
- When updating a device status, the inbound endpoint name must be present
  - This also broke a .NET integration test as our base connector wasn't doing this for the user

It looks like Rust is hitting similar failures as well, so feel free to contribute the fixes to this branch


rust/azure_iot_operations_services/tests/azure_device_registry_client_tests.rs
timtay-microsoft, line 144: @RLeclair can you help me by updating this test so that this updated status includes the device endpoint like I did for .NET above?

dotnet/src/Azure.Iot.Operations.Connector/PollingTelemetryConnectorWorker.cs
```diff
@@ -30,6 +30,12 @@ await args.DeviceEndpointClient.GetAndUpdateDeviceStatusAsync((currentDeviceStat
                     currentDeviceStatus.Config ??= new();
                     currentDeviceStatus.Config.LastTransitionTime = DateTime.UtcNow;
                     currentDeviceStatus.Config.Error = null;
+                    currentDeviceStatus.Endpoints ??= new();
+                    currentDeviceStatus.Endpoints.Inbound ??= new();
+                    currentDeviceStatus.Endpoints.Inbound[args.InboundEndpointName] = new()
```
timtay-microsoft: @RLeclair these are the equivalent .NET changes, for reference

RLeclair (approved):
LGTM!

maximsemenov80 (approved):

PR #1294, opened 2026-04-10
[Rust][Connector] Structural changes for ConfigMap support

cartertinney:
- Renamed `ProjectedVolumeDebouncer` to `AtomicWriterVolumeDebouncer` as it also supports other volume types
- Moved `ConnectorConfiguration` and related structs to their own module to facilitate future work
- Updated documentation as necessary

### DIFF
--- a/rust/azure_iot_operations_connector/src/deployment_artifacts/connector.rs
+++ b/rust/azure_iot_operations_connector/src/deployment_artifacts/connector.rs
@@ -5,17 +5,19 @@
 
 use std::env::VarError;
 use std::ffi::OsString;
-use std::fs::File;
-use std::io::{BufRead, BufReader};
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
 use std::time::Duration;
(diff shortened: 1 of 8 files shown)


rust/azure_iot_operations_connector/src/deployment_artifacts/connector.rs
```diff
@@ -5,17 +5,20 @@
...
 use std::time::Duration;
 
 use azure_iot_operations_mqtt as aio_mqtt;
-use serde::Deserialize;
 use serde_json;
 use thiserror::Error;
 
 use crate::deployment_artifacts::{FileMount, Secrets};
+// TODO: not sure if all this needs exposing here, revisit.
+// These should probably be exporeted in the root of `deployment_artifacts` instead,
```
vaavva:
```suggestion
// These should probably be exported in the root of `deployment_artifacts` instead,
```

vaavva (approved):

PR #1295, opened 2026-04-15
[Rust][Connector] Watchers for `ConfigMap` volume values

cartertinney:
- Added `Watched` for watching changes to a struct.
- Added monitoring for the various fields of `ConnectorConfiguration` extracted from the ConfigMap volume.
- Improved simulation of `TempAtomicWriterVolume` test utility to more accurately simulate K8S behavior.


rust/azure_iot_operations_connector/src/deployment_artifacts/connector_configuration.rs
```diff
@@ -1,135 +1,248 @@
...
 pub struct ConnectorConfiguration {
     /// MQTT connection details
-    pub mqtt_connection_configuration: MqttConnectionConfiguration,
+    pub mqtt_connection_configuration: Watched<MqttConnectionConfiguration>,
     /// Diagnostics
-    pub diagnostics: Option<Diagnostics>,
-    /// Persistent Volume Mount Path
+    pub diagnostics: Watched<Option<Diagnostics>>,
+    /// Persistent Volume Mount Paths
     pub persistent_volumes: Vec<PathBuf>,
```
cartertinney: The data specified by `persistent_volumes` is not intended to change without a pod restart due to the nature of the data, but theoretically it could.

To be specific, it is a vec of paths to K8S Persistent Volume mounts. If a persistent volume is added or removed, this should trigger a pod restart. Thus, any change to this ConfigMap would be immediately followed by a pod restart.

With this in mind: should this value be `Watched`? It'd be more consistent and future-proof us if anything changes. On the other hand, it's useless now, and probably forever.

rust/azure_iot_operations_connector/src/deployment_artifacts/connector_configuration.rs
```diff
@@ -1,135 +1,248 @@
...
-        let additional_configuration = match Self::extract_additional_configuration(mount_path) {
+        let additional_configuration = match extract_additional_configuration(
+            &mount_path.join(ADDITIONAL_CONNECTOR_CONFIGURATION_FILENAME),
+        ) {
             Ok(ac) => Some(ac),
             Err(DeploymentArtifactErrorRepr::FilePathMissing(_)) => None,
             Err(e) => Err(e)?,
         };
 
+        let (mqtt_tx, mqtt_rx) = tokio::sync::watch::channel(initial_mqtt_config);
```
vaavva: nit: I'd add a little more context in the name of this one since there's always a bunch of mqtt stuff going on in the SDKs
```suggestion
        let (mqtt_conf_tx, mqtt_conf_rx) = tokio::sync::watch::channel(initial_mqtt_config);
```

rust/azure_iot_operations_connector/src/deployment_artifacts/connector_configuration.rs
```diff
@@ -1,135 +1,248 @@
...
+                                                );
+                                            }
+                                        }
+                                    }
+                                    DIAGNOSTICS_FILENAME => {
+                                        match extract_diagnostics(&event.path) {
+                                            Ok(new_diag) => {
+                                                let _ = diag_tx.send(Some(new_diag));
+                                            }
+                                            Err(e) => {
```
vaavva, line 113: should this have handling for this case here to match what happens on new?
> Err(DeploymentArtifactErrorRepr::FilePathMissing(_)) => None,

vaavva: oh, looks like this would be a delete notification instead, but is there any desire to still handle this for safety?

rust/azure_iot_operations_connector/src/deployment_artifacts/connector_configuration.rs
```diff
@@ -1,135 +1,248 @@
...
+                                            }
+                                        }
+                                    }
+                                    PERSISTENT_VOLUME_MOUNT_PATHS_FILENAME => {
+                                        log::warn!(
+                                            "Changes to {PERSISTENT_VOLUME_MOUNT_PATHS_FILENAME} are not currently supported and will be ignored"
+                                        );
+                                    }
+                                    _ => {
+                                        log::warn!(
```
vaavva, line 142: does this happen regularly because of the mount weirdness, or is this really unexpected? Only asking because if it could happen semi-regularly, then it might be better to not be a `warn`

rust/azure_iot_operations_connector/src/deployment_artifacts/watched.rs
```diff
@@ -0,0 +1,82 @@
...
+        }
+    }
+
+    /// Returns a reference guard to the current value.
+    ///
+    /// The returned [`WatchedRef`] dereferences to `T` and holds a read lock on the
+    /// inner value. Drop it promptly to avoid blocking updates.
+    #[must_use]
+    pub fn borrow(&self) -> WatchedRef<'_, T> {
+        WatchedRef(self.rx.borrow())
```
vaavva, line 56: what's the reason to use borrow instead of borrow_and_update? I'm assuming it's because you don't want this to mark the value as seen, but the pros/cons here (https://docs.rs/tokio/latest/tokio/sync/watch/index.html#borrow_and_update-versus-borrow) are interesting if we know that this might be used mostly in the loop-type scenario

vaavva (approved):

rust/azure_iot_operations_connector/src/deployment_artifacts/connector_configuration.rs
vaavva: I'd leave it not as `Watched` for now since there would be some background work happening to monitor this that wouldn't be useful, and customers might think then that they need to handle it as if it would update. Since it'll currently cause a pod restart, if we need to have it be `Watched` later on, it's probably a significant enough behavior change to warrant the API breaking change at that point.

cartertinney: I think most of the arguments you make are compelling, but just to be clear, there would not be any additional monitoring overhead, it would be included in the same thread that monitors the other files in the ConfigMap

RLeclair (approved):

rust/azure_iot_operations_connector/src/deployment_artifacts/connector_configuration.rs
cartertinney: This should not ever happen.

rust/azure_iot_operations_connector/src/deployment_artifacts/connector_configuration.rs
cartertinney: No. Technically I could introduce yet another internal error type if you'd prefer, but it's simpler this way.

I could leave a comment if that helps. Or explicitly log unexpected errors, but it seems overkill.

vaavva: seems fine to me. Comment couldn't hurt, but I'm not sure it's necessary since it's clear from the next block of code why it wouldn't happen here

rust/azure_iot_operations_connector/src/deployment_artifacts/watched.rs
cartertinney: Added a comment indicating that this race condition is not rcurrently a concern due to latency in K8S, but we could expose `borrow_and_update` in the future as necessary.

Issue #1296, opened 2026-04-17
[BUG] CommandExecutor operations block if underlying SessionClient is not connected


erhardeibl:
### Product

.NET SDK

### Azure Iot Operations version

N/A (standalone SDK issue, reproducible without AIO deployment)
Reproduced in latest release:
https://github.com/Azure/iot-operations-sdks/releases/tag/net/mqtt/1.0.1
https://github.com/Azure/iot-operations-sdks/releases/tag/net%2Fprotocol%2F1.0.1

### What happened?

The methods `CommandExecutor.StopAsync()` and `DisposeAsync()` hang indefinitely when the underlying `MqttSessionClient` is disconnected.

`CommandExecutor.StopAsync()` calls `_mqttClient.UnsubscribeAsync()`. `MqttSessionClient.UnsubscribeAsync()` enqueues a `QueuedUnsubscribeRequest` with a `TaskCompletionSource` and awaits `tcs.Task`. The only thing that completes the TCS is the worker thread in `ExecuteQueuedItemsAsync`, but `DisconnectAsync()` already cancelled it via `StopPublishingSubscribingAndUnsubscribing()`. The cancellation token registered in `UnsubscribeAsync` could cancel the TCS, but `DisposeAsyncCore` calls `StopAsync()` with no argument.

**Call chain:**

**`StopAsync` path:**
1. `CommandExecutor.StopAsync(CancellationToken.None)` -> `MqttSessionClient.UnsubscribeAsync(options, CancellationToken.None)` -> enqueues request, awaits `tcs.Task`
2. No worker thread, no cancellation -> TCS never completes -> **hangs forever**

**`DisposeAsync` path:**
1. `CommandExecutor.DisposeAsync()` -> `DisposeAsyncCore` -> `StopAsync()` with no argument (`default` = `CancellationToken.None`)
2. Same hang as above, and `DisposeAsync()` has no parameter to accept a cancellation token

Calling `StopAsync(ct)` with a real token could work around it, but `DisposeAsync()` does not accept one.

### List the steps to reproduce

```cs
using Azure.Iot.Operations.Mqtt.Session;
using Azure.Iot.Operations.Protocol;
using Azure.Iot.Operations.Protocol.Connection;
using Azure.Iot.Operations.Protocol.Models;
using Azure.Iot.Operations.Protocol.RPC;
using MQTTnet.Server;
using System.Buffers;
using System.Net;
using System.Net.Sockets;
using System.Text.Json;

[...]

PR #1297, opened 2026-04-18
[.NET][MQTT/Protocol] Rework StopAsync/DisposeAsync methods

timtay-microsoft:
Also add opt-in setting at session client level to immediately throw if a user attempts to publish/subscribe/unsubscribe on a client that is neither connected nor trying to reconnect.

This addresses the issue raised in https://github.com/Azure/iot-operations-sdks/issues/1296 in two ways:
1) If the user opts into the setting mentioned above, command executor tear down will immediately give up on trying to unsubscribe from the RPC topic
2) If the user does not opt into the setting above, the command executor disposal can be canceled by passing in a cancellation token

Also edit codegen such that it outputs RPC/Telemetry clients' disposeAsync methods with cancellation tokens in the overload

Also fix the "run codegen" pipeline now that the wot codegen lives in a new folder 

----------------------

edit: This PR now also includes deliberately not calling ```StopAsync``` in ```DisposeAsync``` methods. This allows users to have finer grain control over how long to attempt to cleanup network state vs how long to attempt to cleanup local resources as discussed with @erhardeibl

As a consequence of this, I have introduced a new ```StopAsync``` method to the command invoker so that it can follow this pattern. This is a mild breaking change.


dotnet/src/Azure.Iot.Operations.Mqtt/Session/MqttSessionClientOptions.cs
```diff
@@ -62,6 +62,12 @@ public class MqttSessionClientOptions : OrderedAckMqttClientOptions
         /// </remarks>
         public TimeSpan ConnectionAttemptTimeout { get; set; } = TimeSpan.FromSeconds(2);
 
+        /// <summary>
+        /// If true, any attempt to publish, subscribe, or unsubscribe while the session client is not either connected or reconnecting will immediately throw
+        /// a <see cref="SessionClosedException"/>. If false, the publish/subscribe/unsubscribe will remain enqueued and can be sent when the session starts again.
+        /// </summary>
+        public bool ThrowIfUsedWhenSessionInactive { get; set; } = false;
```
timtay-microsoft, line 69: This default matches the current behavior of the SDK from before this setting. I am tempted to make the default "True" here instead since the use case for "False" seems limited to me

jrdouceur (approved):

codegen/src/Azure.Iot.Operations.ProtocolCompilerLib/T4/communication/dotnet/Service/t4/DotNetService.tt
```diff
@@ -559,6 +559,12 @@ namespace <#=this.projectName#>.<#=this.genNamespace.GetTypeName(TargetLanguage.
             public async Task StopAsync(CancellationToken cancellationToken = default)
             {
                 await Task.WhenAll(
+<# foreach (var cmdEnvoyInfo in this.cmdEnvoyInfos) { #>
+                    this.<#=cmdEnvoyInfo.Name.GetVariableName(TargetLanguage.CSharp, "command", "invoker")#>.StopAsync(cancellationToken)<#=this.IsLast(cmdEnvoyInfo) ? ").ConfigureAwait(false);" : ","#>
+<# } #>
+
+                await Task.WhenAll(
+
```
jrdouceur: nit: spurious blank line

jrdouceur (approved):

maximsemenov80 (approved):

PR #1298, opened 2026-04-20
[Rust] CI matrix + CI refactor + workspace cleanup

cartertinney:
- Refactored CI infrastructure and scripts to remove outdated remnants of old infrastructure, improve documentation, and overall clarity
- Moved `statestore-cli`, `stub-services` and `rust-connector-scaffolding` into their own respective CI jobs.
- Added a matrix for testing the Rust SDK, including MSRV (1.88) and the latest stable compiler.
- Added caching for builds to improve performance.
- Added tests against variant features
- Enforced use of toolchains for standardizing `make check`
- Minor fixes to `tools/azure_iot_operations_stub_service` as it had static analysis bugs from not being tested properly in the past.
- Disabled the ability to release `tools/azure_iot_operations_stub_service` and `tool/statestore-cli` as library crates.

Improves coverage while also reducing CI runtime by ~10 minutes.

There is one flaky test that will be improved in a follow-up PR. Unrelated to any CI changes other than perhaps adding the stable (currently 1.95) Rust compiler.


.github/workflows/_rust-tool-check.yml
```diff
@@ -0,0 +1,82 @@
...
+      run:
+        shell: bash
+        working-directory: ${{ inputs.path }}
+    env:
+      # Forwarded to cargo only when `aio-sdks-registry: true`; harmless
+      # to set unconditionally (unused by crates.io-only builds).
+      CARGO_REGISTRIES_AIO_SDKS_TOKEN: ${{ secrets.aio-sdks-token }}
+    steps:
+      - uses: actions/checkout@v4
+
```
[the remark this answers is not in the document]
cartertinney: No, we want the flexibility for some tools to depend on the registry, and other to depend on crates.io

vadim-kovalyov (approved):

.github/workflows/ci-rust-connector-scaffolding.yml
```diff
@@ -0,0 +1,32 @@
+name: CI-rust-connector-scaffolding
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - '.github/workflows/ci-rust-connector-scaffolding.yml'
+      - '.github/workflows/_rust-tool-check.yml'
+      - 'rust/sample_applications/sample_connector_scaffolding/**'
```
vaavva, line 9: should this also have rust/ci?

.github/workflows/ci-rust.yml
```diff
@@ -43,82 +39,167 @@ env:
...
+        with:
+          tool: cargo-machete
+
+      - name: Run checks (fmt, clippy, copyright, machete, external crates)
+        run: make check
+
+      - name: Regenerate codegen output and verify no drift
+        working-directory: rust/azure_iot_operations_services
+        run: |
+          sh ./gen.sh
```
vaavva, line 90: so that this properly exits as a failure if one of the steps fails (unless you've validated that it does that without the `set -e`)
```suggestion
          set -e
          sh ./gen.sh
```

.github/workflows/ci-rust.yml
```diff
@@ -43,82 +39,167 @@ env:
...
+  tests:
+    name: Tests (rust ${{ matrix.rust }})
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+
+    strategy:
+      fail-fast: false
+      matrix:
+        # `1.88` is the MSRV declared in rust/Cargo.toml's `rust-version`.
+        # Keep in sync with that field when bumping MSRV.
```
vaavva, line 112: can you add a comment over there to note to bump the msrv here as well if there isn't a way to dynamically link them?

.github/workflows/ci-rust.yml
```diff
@@ -43,82 +39,167 @@ env:
...
+        rust: ["1.88", "stable"]
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install AIO
+        uses: ./.github/actions/configure-aio
+        with:
+          wait-for-broker: 'true'
+          # configure-aio installs the pinned toolchain; we add the
```
vaavva, line 122: wouldn't we want to do the Install Rust step first then for this to correctly pick up the matrix toolchain?

.github/workflows/ci-rust.yml
vaavva: there are a few other `run: ` spots where the `set -e` has been removed from the old file as well

.github/workflows/ci-rust.yml
```diff
@@ -43,82 +39,167 @@ env:
...
-        cargo clippy --all-features
-        cargo fmt --verbose --all
-
-    - name: Verify Stub Service
-      working-directory: tools/azure_iot_operations_stub_services
-      run: | # Checks both with default features and without since the stub service can be used without an output folder
-        source ../../rust/ci/build-deps.sh
-        cargo check 
-        cargo check --no-default-features
-        # cargo clippy --all-features # codegen must be regenerated and the stub service code updated for this to be re-enabled. In the meantime, we'll validate cargo build
```
vaavva, line 100: is this now passing because an older toolchain for clippy is pinned within the stub service crate now?

.github/workflows/ci-rust-connector-scaffolding.yml
```diff
@@ -0,0 +1,32 @@
...
+  push:
+    branches:
+      - main
+  schedule:
+    - cron: '0 11 * * *'  # Nightly (04:00 PT)
```
RLeclair, line 15: I don't recall but did we have a nightly for the scaffolding before?

.github/workflows/_rust-tool-check.yml
```diff
@@ -0,0 +1,114 @@
...
+          shared-key: tool-${{ inputs.name }}
+
+      - name: Format check
+        run: cargo fmt --verbose --all --check
+
+      - name: Clippy
+        run: cargo clippy --all-targets -- --deny=warnings
+
+      - name: Build
+        run: cargo build --verbose
```
vaavva: all features might need to be passed to this and cargo clippy for some of the runs (services, and also stub services should be run with and without all features to validate both work)

rust/scripts/check.sh
```diff
@@ -0,0 +1,72 @@
...
+        echo "ERROR: ${FILE} is missing [lints] workspace = true" >&2
+        exit 1
+    fi
+done
+
+# Require `rustfmt` pass.
+cargo fmt --verbose --all --check
+
+# Require clippy pass.
+cargo clippy --all --all-features --tests --examples -- --deny=warnings
```
vaavva, line 70: for things that have features, we probably want to make sure clippy passes with and without them

.github/workflows/ci-rust-connector-scaffolding.yml
vaavva: or rather, rust/scripts now

.github/workflows/ci-rust.yml
```diff
@@ -43,82 +39,167 @@ env:
...
+      fail-fast: false
+      matrix:
+        # `1.88` is the MSRV declared in rust/Cargo.toml's `rust-version`.
+        # Keep in sync with that field when bumping MSRV.
+        rust: ["1.88", "stable"]
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install AIO
```
RLeclair, line 118: Why do you need to install AIO for each leg of the matrix? I can see this only being needed once

Edit: I assume this is because of the ADR resources that get deployed and altered during the tests?

.github/workflows/ci-rust-connector-scaffolding.yml
cartertinney: Good call. Added to all three tools CIs.

rust/scripts/cargo-tools.sh
```diff
@@ -0,0 +1,32 @@
...
+
+set -euo pipefail
+
+if ! command -v cargo >/dev/null 2>&1; then
+    echo "error: cargo not found on PATH; install the Rust toolchain first" >&2
+    exit 1
+fi
+
+if ! command -v cargo-llvm-cov >/dev/null 2>&1; then
+    cargo install --version '^0.5' --locked cargo-llvm-cov
```
RLeclair, line 27: A separate discussion but I believe some of our newer components (connectors, sr) are using cargo  tarpaulin

Not sure if there is an advantage to it over cargo-llvm-cov

.github/workflows/ci-rust.yml
cartertinney: This is basically saying that we deploy AIO with a pinned toolchain, but then we install a variant version of Rust depending on what's under test for the matrix, i.e. all legs of the matrix deploy AIO with the same Rust, but the SDK is compiled with different versions.

RLeclair (approved):

.github/workflows/ci-rust.yml
cartertinney: Not sure exactly what you mean here, but stub services, as well as all the other "tools" (including connector scaffolding), now get their own dedicated CI.

.github/workflows/ci-rust-connector-scaffolding.yml
cartertinney: It was part of the SDK CI before, which didn't make sense.

.github/workflows/ci-rust.yml
cartertinney: Parallel tests - I don't want to run into client ID collisions or topic collisions, etc.

If they ran sequentially they could share, but it would be slower, and also we wouldn't get the nice ability to separate out the compiler versions in ADO

rust/scripts/cargo-tools.sh
cartertinney: I have no opinion on this. I'm just modifying what was already here.

rust/scripts/check.sh
cartertinney: Added cargo hack to address this for clippy.

.github/workflows/ci-rust.yml
cartertinney: Technically not necessary due to GH actions bash defaults + the content of the gen.sh script, but I'll add `set -euo pipefail` defensively.

Not sure what you mean about there being other `run:` spots where this is necessary - this only matters for multiline blocks AFAIK

.github/workflows/_rust-tool-check.yml
cartertinney: Addressed.

.github/workflows/ci-rust.yml
vaavva: okay, if this only affects multi-line ones, then this looks good now!

.github/workflows/ci-rust.yml
vaavva: This was previously there as a comment for the stub service tests
> # cargo clippy --all-features # codegen must be regenerated and the stub service code updated for this to be re-enabled. In the meantime, we'll validate cargo build
I don't think we've updated the stub services, so I'm kinda surprised that clippy is now passing with your new testing format for the stub services

vaavva (approved):

.github/workflows/ci-rust.yml
cartertinney: Works fine locally and in CI. Not sure what that comment ever meant. I also improved it even further to use cargo hack to get even more coverage of feature combinations.

.github/workflows/ci-rust-connector-scaffolding.yml
RLeclair: Yeah I am not sure why, maybe worth removing in a separate PR

PR #1310, opened 2026-04-23
[.NET] Management Actions 

maximsemenov80:
This pull request introduces a new sample connector, `ManagementActionConnector`, and integrates it into the CI/CD pipeline and solution. It also adds the implementation for the connector, including device simulation, action handlers, and shared contracts.


dotnet/samples/Connectors/ManagementActionConnector/ManagementActionConnectorWorker.cs
```diff
@@ -0,0 +1,350 @@
...
+        {
+            _logger.LogInformation(
+                "Asset {AssetName} on device {DeviceName} is available.",
+                args.AssetName,
+                args.DeviceName);
+
+            // Spawn one handler task per management action. They run concurrently for the
+            // lifetime of the asset; cancelling 'cancellationToken' tears them all down.
+            var actionTasks = new List<Task>();
+            foreach (var group in args.Asset.ManagementGroups ?? Enumerable.Empty<AssetManagementGroup>())
```
timtay-microsoft: Rather than ask the user to manage all these threads, would it be possible for the base connector to do that for them? I'm imagining that our base connector just sends the application layer a callback each time a management action is invoked. Something like

```dotnet
public class ConnectorWorker
{
    public Func<ManagementActionInvokedEventArgs, CancellationToken, Task<ManagementActionResponse>>? OnManagementActionInvoked;
}
```

where ```ManagementActionInvokedEventArgs``` contains details on what management action the request is related to

dotnet/src/Azure.Iot.Operations.Connector/ManagementActionRequest.cs
```diff
@@ -0,0 +1,70 @@
...
+    /// </summary>
+    public sealed class ManagementActionRequest : IAsyncDisposable
+    {
+        internal ManagementActionRequest()
+        {
+            // Populated by ManagementActionExecutor from the underlying ExtendedRequest.
+        }
+
+        /// <summary>Raw request payload bytes as delivered by the invoker.</summary>
+        public byte[] Payload { get; internal set; } = Array.Empty<byte>();
```
timtay-microsoft: nit: Some of our users strongly prefer we use ```ReadOnlySequence<byte>``` to represent data types like these. Supposedly, there are some performance concerns with just using ```byte[]```

dotnet/src/Azure.Iot.Operations.Connector/ManagementActionRequest.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// </summary>
+        public IReadOnlyDictionary<string, string> TopicTokens { get; internal set; }
+            = new Dictionary<string, string>();
+
+        /// <summary>
+        /// True if the executor has been cancelled (action deleted, connector shutting
+        /// down, etc.) after this request was received. A completion call will still be
+        /// attempted but may not reach the invoker.
+        /// </summary>
+        public bool IsCancelled => throw new NotImplementedException();
```
timtay-microsoft: If we do find a way to just provide a callback like I mentioned in a previous comment, this concept would be replaced by us just giving the user a cancellation token in that callback

dotnet/src/Azure.Iot.Operations.Connector/ManagementActionExecutor.cs
```diff
@@ -0,0 +1,35 @@
...
+        internal ManagementActionExecutor()
+        {
+            // Constructed by AssetClient / ConnectorWorker; wraps the underlying CommandExecutor.
+        }
+
+        /// <summary>
+        /// Await the next incoming management action invocation for this executor.
+        /// Returns <c>null</c> if the executor has been shut down (no more requests will
+        /// be delivered).
+        /// </summary>
```
timtay-microsoft, line 97: It would probably be worthwhile to note that the provided cancellation token can be used to signal that the request is no longer applicable (management action was deleted, connector is shutting down, etc.)

We want to make sure a user is aware of why they should check this token when processing a management action

dotnet/src/Azure.Iot.Operations.Connector/ManagementActionExecutor.cs
```diff
@@ -0,0 +1,35 @@
...
+        public Task<ManagementActionRequest?> RecvRequestAsync(CancellationToken cancellationToken = default)
+            => throw new NotImplementedException();
+
+        /// <summary>
+        /// Stop the underlying <c>CommandExecutor</c>, unsubscribe from the action
+        /// request topic, and release resources. Any still-queued requests are
+        /// auto-completed with an error response.
+        /// </summary>
+        public ValueTask DisposeAsync() => throw new NotImplementedException();
```
timtay-microsoft: In #1297, I'm reworking the StopAsync/DisposeAsync patterns a bit and it should be reflected here as well. Generally, StopAsync should exist to tear down MQTT subscriptions. DisposeAsync should exist to dispose local resources (semaphores, callback registrations, etc) and optionally dispose the underlying MQTT client. 

For this class in particular, I don't think there is a use case for disposing the underlying MQTT client, so you may just want to have something like

```dotnet 
        // unsubscribe from MQTT topics
        public ValueTask StopAsync(CancellationToken cancellationToken = default) => throw new NotImplementedException();

        // dispose local resources
        public ValueTask DisposeAsync(CancellationToken cancellationToken = default) => throw new NotImplementedException();
```

dotnet/samples/Connectors/ManagementActionConnector/ManagementActionConnectorWorker.cs
timtay-microsoft: Or maybe the callback func lives in the asset client

```dotnet
public class AssetClient
{
    public Func<ManagementActionInvokedEventArgs, CancellationToken, Task<ManagementActionResponse>>? OnManagementActionInvoked;
}
```

timtay-microsoft: This approach probably requires that we delineate between "calls", "reads", and "writes", though. So maybe more like 

```dotnet
public class AssetClient
{
    // base connector forwards the response back to the broker
    public Func<ManagementActionCallInvokedEventArgs, CancellationToken, Task<ManagementActionCallResponse>>? OnManagementActionInvoked;

    // base connector forwards the read value back to the broker
    public Func<ManagementActionReadInvokedEventArgs, CancellationToken, Task<ManagementActionReadResponse>>? OnManagementActionInvoked;

    // Response object has fewer fields because write calls maybe don't return anything?
    public Func<ManagementActionWriteInvokedEventArgs, CancellationToken, Task<ManagementActionWriteResponse>>? OnManagementActionInvoked;
}
```

dotnet/samples/Connectors/ManagementActionConnector/KubernetesResources/mgmt-action-asset-definition.yaml
```diff
@@ -0,0 +1,41 @@
...
+      # Fallback topic used by any action in this group that does not
+      # declare its own `topic`. Tokens in braces are resolved at
+      # runtime against the device / endpoint / asset context.
+      defaultTopic: mgmt/{deviceName}/{assetName}/{managementGroupName}/{managementActionName}
+      defaultTimeoutInSeconds: 30
+      managementGroupConfiguration: '{}'
+      actions:
+        - name: reboot
+          # ActionType values: Call | Read | Write (see AssetManagementGroupActionType)
+          actionType: Call
```
timtay-microsoft, line 31: I definitely think it would help your development of the connector APIs if we had a sample that demonstrates all three types of management actions. That way, we can see how a developer would feel when writing support for each kind of management action.

I wouldn't worry about developing the actual service that this connector "connects" with yet (like the HTTP server in the sample rest connector, the TCP server in the sample event-driven connector), but I would consider coming up with a scenario that allows you to show all three management action types.

Maybe something like another rest server that allows you to read and write one particular value (the "Read" management action and the "Write" management action), and allows you to update/reboot the server (the "Call" management action)

timtay-microsoft: I don't think most connectors will actually use all three types, but this sample connector will be how we test this feature in our integration tests, and it is useful even just as a sample we can point users towards

doc/dev/tmp/management-action-design-onepager.md
```diff
@@ -0,0 +1,239 @@
...
+| **ManagementActionRequest** | Incoming invocation — exposes payload, metadata; completed via `CompleteAsync(response)` |
+| **ManagementActionResponse** | `record` with `required` Payload, ContentType, CloudEvent; optional ApplicationError |
+| **ManagementActionApplicationError** | `record` with ErrorCode + ErrorPayload |
+| **ManagementActionNotification** | Abstract `record` base with 4 derived types: Updated, UpdatedWithNewExecutor, AssetUpdated, Deleted |
+
+### Modified Types
+
+| Type | Change |
+|------|--------|
+| **AssetClient** | New methods: `GetManagementActionExecutorAsync`, `RecvManagementActionNotificationAsync`, `ReportManagementAction{Request/Response}MessageSchemaAsync`. New internal state: per-action notification channels. New dependency: `SchemaRegistryClient` |
```
vaavva: shouldn't SchemaRegistryClient already be a part of this for dataset/etc message schema reporting?

vaavva: reading further down how the dataset/event schemas are reported as part of forwarding, I'm still a bit confused about the existing behavior - does the application need to attach the full schema each time data is forwarded, even if they know it hasn't changed?

doc/dev/tmp/management-action-design-onepager.md
```diff
@@ -0,0 +1,239 @@
...
+    CancellationToken ct)
+{
+    // Register schemas
+    await assetClient.ReportManagementActionRequestMessageSchemaAsync(
+        groupName, actionName, requestSchema, ct);
+    await assetClient.ReportManagementActionResponseMessageSchemaAsync(
+        groupName, actionName, responseSchema, ct);
+
+    // Get initial executor
+    var executor = await assetClient.GetManagementActionExecutorAsync(
```
vaavva: I'm not super up to date on my .net, but note that there might not always be a valid executor depending on the definition, if there need to be any null checks/handling here for that

dotnet/samples/Connectors/ManagementActionConnector/KubernetesResources/mgmt-action-asset-definition.yaml
```diff
@@ -0,0 +1,41 @@
...
+  #   group  = "device-control"
+  #   action = "reboot"
+  # keyed internally as "device-control::reboot".
+  # -----------------------------------------------------------------
+  managementGroups:
+    - name: device-control
+      # Fallback topic used by any action in this group that does not
+      # declare its own `topic`. Tokens in braces are resolved at
+      # runtime against the device / endpoint / asset context.
+      defaultTopic: mgmt/{deviceName}/{assetName}/{managementGroupName}/{managementActionName}
```
vaavva: for now, this field won't be supported because of ambiguity and these tokens not being supported yet, so I wouldn't include this field in your definition example

dotnet/samples/Connectors/ManagementActionConnector/KubernetesResources/mgmt-action-asset-definition.yaml
```diff
@@ -0,0 +1,41 @@
...
+      defaultTimeoutInSeconds: 30
+      managementGroupConfiguration: '{}'
+      actions:
+        - name: reboot
+          # ActionType values: Call | Read | Write (see AssetManagementGroupActionType)
+          actionType: Call
+          targetUri: device://reboot
+          # Optional explicit topic. Omit to inherit `defaultTopic`.
+          topic: mgmt/{deviceName}/{assetName}/device-control/reboot
```
vaavva: similar comment here - don't have these tokens in your sample because the cloud side of management actions won't be able to support them

dotnet/samples/Connectors/ManagementActionConnector/ManagementActionConnectorWorker.cs
```diff
@@ -0,0 +1,350 @@
...
+            }
+
+            await executor.DisposeAsync();
+            _logger.LogInformation("Handler for {Group}::{Action} exited.", groupName, actionName);
+        }
+
+        private async Task HandleRequestAsync(
+            ManagementActionRequest request,
+            string groupName,
+            string actionName,
```
vaavva: for a more full example, you'd probably need to know the target uri/any additional configuration. I don't think you need to plumb it through in this sample necessarily, as long as it is easily possible since that will be the standard flow

dotnet/samples/Connectors/ManagementActionConnector/ManagementActionConnectorWorker.cs
```diff
@@ -0,0 +1,350 @@
...
+        {
+            switch (notification)
+            {
+                case ManagementActionUpdated updated:
+                    _logger.LogInformation(
+                        "{Group}::{Action} definition updated (same topic). error={Error}",
+                        groupName, actionName, updated.Error);
+                    await assetClient.PauseManagementActionRuntimeHealthReportingAsync(groupName, actionName, cancellationToken);
+                    // Real connectors would re-validate the new definition here. For this sample
+                    // we assume it's fine and re-report Available, which also resumes periodic
```
vaavva: you'd probably also report config status here

dotnet/samples/Connectors/ManagementActionConnector/ManagementActionConnectorWorker.cs
```diff
@@ -0,0 +1,350 @@
...
+            ManagementActionNotification notification,
+            ManagementActionExecutor currentExecutor,
+            Action<ManagementActionExecutor> updateExecutor,
+            CancellationToken cancellationToken)
+        {
+            switch (notification)
+            {
+                case ManagementActionUpdated updated:
+                    _logger.LogInformation(
+                        "{Group}::{Action} definition updated (same topic). error={Error}",
```
vaavva: if there's an error here, should that prevent the health from being reported as available?

dotnet/samples/Connectors/ManagementActionConnector/ManagementActionConnectorWorker.cs
```diff
@@ -0,0 +1,350 @@
...
+                    return false;
+
+                case ManagementActionAssetUpdated assetUpdated:
+                    _logger.LogInformation(
+                        "{Group}::{Action}: parent asset updated. error={Error}",
+                        groupName, actionName, assetUpdated.Error);
+                    return false;
+
+                case ManagementActionDeleted:
+                    _logger.LogInformation("{Group}::{Action} deleted.", groupName, actionName);
```
vaavva: executor should be drained on this branch as well

dotnet/samples/Connectors/ManagementActionConnector/README.md
```diff
@@ -0,0 +1,904 @@
...
+       .GetManagementActionExecutorAsync("device-control", "reboot", ct);
+   ```
+   Every `ManagementActionRequest` that comes out of *that* executor's
+   `RecvRequestAsync` is, by construction, an invocation of
+   `device-control::reboot`. So if you already decided that `reboot`'s
+   input type is `MyRebootArgs`, there is no ambiguity — the `(group,
+   action)` tuple is your dispatch key.
+2. **You published the request's schema to the Schema Registry** (via
+   `AssetClient.ReportManagementActionRequestMessageSchemaAsync(...)`).
+   That schema is the public contract: invokers see it through the asset's
```
vaavva: The invoker doesn't (currently) necessarily do any validation that it must match this schema, so it being reported isn't necessarily a guarantee

dotnet/samples/Connectors/ManagementActionConnector/README.md
```diff
@@ -0,0 +1,904 @@
...
+    return;
+}
+
+// (b) Optional: version check via a custom header you defined in your schema
+if (request.CustomUserData.TryGetValue("x-schema-version", out var v) && v != "1")
+{
+    /* reject with SchemaVersionMismatch */
+}
+
+// (c) Now it is safe to deserialize to the type you chose for this action
```
vaavva: note as well that the connector doesn't necessarily need to deserialize the payload. If it has enough information to just forward the payload to the target uri and then forward the response back to the broker, this is acceptable for some scenarios (like a generic REST connector, whereas a more specific connector might want to do the deserialization for validation)

dotnet/src/Azure.Iot.Operations.Connector/AssetClient.cs
```diff
@@ -396,6 +396,52 @@ private async Task<AssetStatus> GetAssetStatusAsync(
                 cancellationToken);
         }
 
+        // ============================================================
+        // Management Action API
+        // ============================================================
+
+        /// <summary>
+        /// Get the <see cref="ManagementActionExecutor"/> for <paramref name="managementGroupName"/> /
+        /// <paramref name="managementActionName"/>. Returns the currently-valid executor bound to
```
vaavva: called this out somewhere else, but I'll call it out here as well - there might not always be a valid executor, but that's not a use-case that should cause any panicking, it just requires waiting for a new action definition

dotnet/src/Azure.Iot.Operations.Connector/AssetClient.cs
```diff
@@ -396,6 +396,52 @@ private async Task<AssetStatus> GetAssetStatusAsync(
...
+            string managementActionName,
+            CancellationToken cancellationToken = default)
+            => throw new NotImplementedException();
+
+        /// <summary>
+        /// Pause periodic runtime-health reporting for a specific management action. Used on
+        /// definition changes so the next health event reflects the re-validated definition rather
+        /// than the previous one. Matches Rust's <c>pause_and_refresh_health_version</c> pattern.
+        /// </summary>
+        public Task PauseManagementActionRuntimeHealthReportingAsync(
```
vaavva: from the other docs in this PR, I thought this already existed?

dotnet/src/Azure.Iot.Operations.Connector/ManagementActionExecutor.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// <summary>
+        /// Await the next incoming management action invocation for this executor.
+        /// Returns <c>null</c> if the executor has been shut down (no more requests will
+        /// be delivered).
+        /// </summary>
+        public Task<ManagementActionRequest?> RecvRequestAsync(CancellationToken cancellationToken = default)
+            => throw new NotImplementedException();
+
+        /// <summary>
+        /// Stop the underlying <c>CommandExecutor</c>, unsubscribe from the action
```
vaavva: I think this should probably be done by the Connector SDK when an executor is known to be no longer active. When the application is draining the executor, the unsubscribe should have already occurred so that no new requests come in once we don't want to use the definition anymore

dotnet/src/Azure.Iot.Operations.Connector/ManagementActionNotification.cs
```diff
@@ -0,0 +1,57 @@
...
+    public sealed record ManagementActionUpdatedWithNewExecutor(
+        ManagementActionExecutor? NewExecutor,
+        ConfigError? Error) : ManagementActionNotification;
+
+    /// <summary>
+    /// The parent asset was updated but this specific management action's definition
+    /// is unchanged. The connector may need to re-evaluate surrounding state
+    /// (device context, other actions on the asset).
+    /// </summary>
+    public sealed record ManagementActionAssetUpdated(ConfigError? Error) : ManagementActionNotification;
```
vaavva: I'm obviously a fan of this pattern since we did it for rust, but if this pattern isn't similar for asset update notifications for datasets/events/streams in .net, then it could be worth considering how they should be aligned

dotnet/src/Azure.Iot.Operations.Connector/ManagementActionRequest.cs
vaavva: +1, if this isn't a function on the base .net executor request already, then this is probably a rust-ism that shouldn't be ported over. For us, it just calls the internal fn by the same name

dotnet/src/Azure.Iot.Operations.Connector/AssetClient.cs
maximsemenov80: it is implemented in the AssetRuntimeHealthReporter here we'll call it using private instance of that.

doc/dev/tmp/management-action-design-onepager.md
maximsemenov80: you right that "new dependency" wording is misleading. SchemaRegistryClient is already used in the Connector layer today instantiate one on demand and call PutAsync for dataset/event schema registration. 

On your second question — no, the schema is not re-attached on every forward. The first time a (device, endpoint, asset, dataset|event) tuple is forwarded, the SDK calls MessageSchemaProvider.GetMessageSchemaAsync(...), uploads via SchemaRegistryClient.PutAsync(...), and caches the returned Schema (_registeredDatasetMessageSchemas / _registeredEventMessageSchemas). Every subsequent forward hits the cache and only attaches a MessageSchemaReference.

dotnet/src/Azure.Iot.Operations.Connector/IManagementActionHandlerFactory.cs
```diff
@@ -0,0 +1,46 @@
...
+
+namespace Azure.Iot.Operations.Connector
+{
+    /// <summary>
+    /// Factory for creating <see cref="IManagementActionHandler"/> instances, one per management
+    /// action defined on an asset. Called by <see cref="ManagementActionConnectorWorker"/> when
+    /// an asset becomes available.
+    /// </summary>
+    /// <remarks>
+    /// Follows the same factory pattern as <see cref="IDatasetSamplerFactory"/>: per-action
```
timtay-microsoft, line 15: This was a helpful note to help me contextualize this interface, thanks!

dotnet/samples/Connectors/ManagementActionConnector/ManagementActionConnectorWorkerSample.cs
```diff
@@ -0,0 +1,54 @@
...
+{
+    /// <summary>
+    /// Thin <see cref="BackgroundService"/> wrapper that owns a
+    /// <see cref="Azure.Iot.Operations.Connector.ManagementActionConnectorWorker"/>
+    /// and runs its <see cref="ConnectorWorker.RunConnectorAsync"/> loop. All the
+    /// interesting business logic lives in the per-action handlers under
+    /// <c>Handlers/</c>; the SDK base class handles executor lifecycle, notification
+    /// processing, drain, and health/config reporting on our behalf.
+    /// </summary>
+    public sealed class ManagementActionConnectorWorkerSample : BackgroundService
```
timtay-microsoft: nit: don't declare a class with a name different than the file name. This class can just be called "ManagementActionConnectorWorker"

dotnet/samples/Connectors/ManagementActionConnector/KubernetesResources/mgmt-action-asset-definition.yaml
```diff
@@ -0,0 +1,48 @@
...
+metadata:
+  name: my-mgmt-action-asset
+  namespace: azure-iot-operations
+spec:
+  uuid: 3456-7890-1234-5678
+  attributes:
+    assetId: my-mgmt-action-asset
+    assetType: mgmt-action-sample
+  # -----------------------------------------------------------------
+  # Single management group with a single action, intended for
```
timtay-microsoft: nit: out of date documentation. This asset now supports all three types of actions

dotnet/src/Azure.Iot.Operations.Connector/IManagementActionHandler.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        /// <summary>
+        /// Handle a "read" action (minimal/no request payload, value payload out).
+        /// </summary>
+        /// <param name="args">Details about the invocation including payload and metadata.</param>
+        /// <param name="cancellationToken">
+        /// Signaled when the request is no longer applicable.
+        /// </param>
+        /// <returns>The response to send back to the invoker containing the read value.</returns>
+        Task<ManagementActionResponse> HandleReadAsync(ManagementActionInvokedEventArgs args, CancellationToken cancellationToken);
```
timtay-microsoft: It seems a bit incorrect to use the same parameters and return types for all of HandleRead/HandleCall/HandleWrite here. 

For example, a read management action will never have a request payload (going off of [this](https://learn.microsoft.com/en-us/azure/iot-operations/discover-manage-assets/howto-use-management-actions#action-types)). Similarly, a write management action will never have a response payload.

So we maybe want something like 

```csharp 
class IManagementActionHandler
{
    Task<ManagementActionReadResponse> HandleReadAsync(ManagementActionReadInvokedEventArgs args, CancellationToken cancellationToken);
    Task<ManagementActionWriteResponse> HandleWriteAsync(ManagementActionWriteInvokedEventArgs args, CancellationToken cancellationToken);
    Task<ManagementActionCallResponse> HandleCallAsync(ManagementActionCallInvokedEventArgs args, CancellationToken cancellationToken);
}

...
public abstract class ManagementActionResponse
{
    // remove this payload field
    //public required ReadOnlySequence<byte> Payload { get; init; }
    public required string ContentType { get; init; }
    public required CloudEvent? CloudEvent { get; init; }
    public MqttPayloadFormatIndicator FormatIndicator { get; init; }
    public Dictionary<string, string>? CustomUserData { get; init; }
    public ManagementActionApplicationError? ApplicationError { get; init; }
}

public class ManagementActionReadResponse : ManagementActionResponse
{
    // Read management actions do have a response payload
    public required ReadOnlySequence<byte> Payload { get; init; }
}

public class ManagementActionWriteResponse : ManagementActionResponse
{
    // Write management actions do not have a response payload
}

public class ManagementActionCallResponse : ManagementActionResponse
{
    // Call management actions do have a response payload
    public required ReadOnlySequence<byte> Payload { get; init; }
}
```

timtay-microsoft: And a similar split for the parameter ManagementActionInvokedEventArgs

dotnet/src/Azure.Iot.Operations.Connector/ManagementActionInvokedEventArgs.cs
```diff
@@ -0,0 +1,59 @@
...
+    public class ManagementActionInvokedEventArgs : EventArgs
+    {
+        /// <summary>The management group name this action belongs to.</summary>
+        public required string GroupName { get; init; }
+
+        /// <summary>The management action name.</summary>
+        public required string ActionName { get; init; }
+
+        /// <summary>The type of the management action (Call, Read, or Write).</summary>
+        public required AssetManagementGroupActionType ActionType { get; init; }
```
timtay-microsoft, line 34: This field probably isn't necessary to include since the callbacks we expose to the user already explicitly say "Handle**Read**Async"

dotnet/src/Azure.Iot.Operations.Connector/ManagementActionInvokedEventArgs.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// Values extracted from wildcard segments of the action request topic pattern.
+        /// </summary>
+        public IReadOnlyDictionary<string, string> TopicTokens { get; init; }
+            = new Dictionary<string, string>();
+
+        /// <summary>The name of the asset that this action belongs to.</summary>
+        public required string AssetName { get; init; }
+
+        /// <summary>The name of the device that the asset belongs to.</summary>
+        public required string DeviceName { get; init; }
```
timtay-microsoft, line 65: It is a bit redundant to include asset name, device name, group name, and action name in this args bundle since all of this information is already provided to the IManagementActionHandler when it is constructed in the IManagementActionHandlerFactory

timtay-microsoft: This may mean you can just remove these fields from this args bundle, but it also may mean that they are more appropriate to provide in this args bundle than they are in the factory method. I'd consider trying out both approaches and seeing which makes the sample writing feel easier.

I'm guessing the ```HandleCallAsync```/```HandleReadAsync```/```HandleWriteAsync``` methods will want to access this information so that the handler can read/call/write the correct asset on the correct device, so it may make more sense to keep these fields in this args bundle and remove them from the factory method.

dotnet/samples/Connectors/ManagementActionConnector/Handlers/ReadTemperatureHandler.cs
```diff
@@ -0,0 +1,70 @@
...
+        public ReadTemperatureHandler(ILogger logger, FakeDevice device, string deviceName, string assetName)
+        {
+            _logger = logger;
+            _device = device;
+            _deviceName = deviceName;
+            _assetName = assetName;
+        }
+
+        public Task<ManagementActionResponse> HandleCallAsync(ManagementActionInvokedEventArgs args, CancellationToken cancellationToken)
+            => throw new InvalidOperationException("ReadTemperatureHandler is wired to a Read action; HandleCallAsync should never be called.");
```
timtay-microsoft: Since it will probably be common that, like in this sample, users will want a way to express "this handler doesn't handle that type of management action", I'd like to see you declare a formal way for users to express this. 

For example, you could declare some singleton that extends the ManagementActionResponse type like

```csharp
public class UnsupportedManagementActionInvokedResponse : ManagementActionResponse
{
    // singleton definition etc.
}
``` 

This would allow users to do something like

```csharp
        public Task<ManagementActionResponse> HandleCallAsync(ManagementActionInvokedEventArgs args, CancellationToken cancellationToken)
            => return Task.FromResult(UnsupportedManagementActionInvokedResponse.Instance);
```

And that would allow our base connector to handle that case like

```csharp
public class ConnectorWorker
{
    public void ManagementActionInvoked
    {
        IManagementActionHandler handler = factory.getHandler();
        ManagementActionResponse userResponse = await handler.Invoke(...);
        if (userResponse == UnsupportedManagementActionInvokedResponse.Instance)
        {
             _logger.log("An unsupported management action was invoked etc. etc.")
        }
    }
}
```

timtay-microsoft: You could also just define an exception type like

```csharp
public class UnsupportedManagementActionInvokedException : Exception
{
}
```

And our base connector would know to catch that specific exception and log it accordingly

dotnet/src/Azure.Iot.Operations.Connector/ManagementActionExecutor.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// Returns <c>null</c> if the executor has been shut down (no more requests will
+        /// be delivered).
+        /// </summary>
+        /// <param name="cancellationToken">
+        /// Cancels the wait, and should also be observed while processing the returned
+        /// request: it is signalled when the request is no longer applicable (action
+        /// deleted or replaced, asset unavailable, connector shutting down) so the
+        /// handler can abort device I/O instead of producing an undeliverable response.
+        /// </param>
+        public Task<ManagementActionRequest?> RecvRequestAsync(CancellationToken cancellationToken = default)
```
timtay-microsoft: This style of callback vs await may not line up easily with how our RPC layer works. 

If look at the CommandExecutor, it does not offer a method like "RecvCommandAsync" that asynchronously waits for the next command to come in. Instead, it offers a callback that users subscribe to:

```csharp
        public required Func<ExtendedRequest<TReq>, CancellationToken, Task<ExtendedResponse<TResp>>> OnCommandReceived { get; set; }
``` 

As a result, you would need some conversion logic to convert the RPC callback into this async method which might be a bit clunky. I'd consider trying to match the callback style of the command executor since this class is a pretty thin wrapper of the command executor.

timtay-microsoft: Note that this suggested change would also affect the ```HandleReadAsync```/```HandleWriteAsync```/```HandleCallAsync``` methods that your IManagementActionHandler currently defines. They would also benefit from becoming more callback-like

dotnet/src/Azure.Iot.Operations.Connector/IManagementActionHandler.cs
timtay-microsoft: I'm basing all of this on the assumption that read actions have no request payload and write actions have no response payload, so please correct me if I'm wrong on that one.

dotnet/samples/Connectors/ManagementActionConnector/ManagementActionConnectorWorkerSample.cs
maximsemenov80: it wraps "base" ManagementActionConnectorWorker, hence the Sample prefix to avoid a confusion, I'll rename the file.

dotnet/src/Azure.Iot.Operations.Connector/IManagementActionHandler.cs
maximsemenov80: @vaavva could you please confirm: the Rust SDK treats request and response payloads symmetrically across all three ActionType variants.
Potentially, Read could legitimately take a filter/query in the request, and Write could echo the new state in the response - so this seems more like a connector author concern.

dotnet/src/Azure.Iot.Operations.Connector/ManagementActionInvokedEventArgs.cs
maximsemenov80: The latter feels more natural.

dotnet/samples/Connectors/ManagementActionConnector/Handlers/ReadTemperatureHandler.cs
maximsemenov80: I'll go with Exception

dotnet/src/Azure.Iot.Operations.Connector/ManagementActionExecutor.cs
maximsemenov80: Addressed the first part of the comment, but I’m not quite seeing how it impacts IManagementActionHandler

dotnet/src/Azure.Iot.Operations.Connector/IManagementActionHandler.cs
vaavva: oh yeah, I don't think we should be opinionated at all about the form of these different types of management actions. The mslearn page is interesting, but ultimately just guidance since there's nothing stopping someone from assigning arbitrary meaning to the different types as long as they know what it is. Any handling of the differences should be by the connector author.

Additionally, the type (read/write/call) is on the definition of the management action, so the type won't change per request, if that makes any difference to the mental model of how the code would handle this

doc/dev/tmp/management-action-design-onepager.md
```diff
@@ -0,0 +1,231 @@
...
+    style ManagementActionUpdatedWithNewExecutor fill:#d4edda,stroke:#28a745
+    style ManagementActionAssetUpdated fill:#d4edda,stroke:#28a745
+    style ManagementActionDeleted fill:#d4edda,stroke:#28a745
+```
+**Legend:** Blue = user-facing types (interface + base class). Green = internal types. Yellow = modified existing types.
+---
+## Key Design Decisions
+### 1. IManagementActionHandler interface with three methods (Call/Read/Write)
+Per review feedback, the base connector should manage all per-action threading and lifecycle internally. Users implement `IManagementActionHandler` with three methods matching the `AssetManagementGroupActionType` enum:
+- `HandleCallAsync` — general RPC (payload in, payload out)
```
vaavva: What is the benefit of having separate callbacks per type rather than having the user do that split themselves? I imagine some connectors might have very similar implementations for all three (ex. if they just need to change one parameter on what they send to the endpoint), and a single action will only ever be able to receive one type of request at a time since it's defined in the definition

doc/dev/tmp/management-action-design-onepager.md
```diff
@@ -0,0 +1,231 @@
...
+### 3. ManagementActionConnectorWorker base class
+Mirrors `PollingTelemetryConnectorWorker`. Extends `ConnectorWorker` and internalizes:
+- Per-action task spawning and cancellation
+- Executor acquisition, swap (drain old + adopt new)
+- All 4 notification types (Updated, UpdatedWithNewExecutor, AssetUpdated, Deleted)
+- Health + config-status reporting (respects `Error` field — reports Unavailable on error)
+- Exception-to-error-response translation (unhandled handler exceptions ? `InternalError`)
+- Drain-and-dispose with timeout budget
+### 4. Notification model (internal, not user-facing)
+`ManagementActionNotification` with its 4 variants (Updated, UpdatedWithNewExecutor, AssetUpdated, Deleted) is internal to `ManagementActionConnectorWorker`. Users never interact with notifications directly — the base class handles them and calls the user's handler methods only when actual invocations arrive. CancellationToken signals lifecycle events to the handler.
```
vaavva: by
> calls the user's handler methods only when actual invocations arrive

do you mean when the notification arrives, or that the customer only gets change notifications when they receive the next request?

doc/dev/tmp/management-action-implementation-design.md
```diff
@@ -0,0 +1,1842 @@
...
+
+**Internalized responsibilities:**
+- Sets `WhileAssetIsAvailable` to iterate `asset.ManagementGroups` and spawn per-action tasks
+- Creates `IManagementActionHandler` via factory for each action, passing full context
+- Acquires executor via `AssetClient.GetManagementActionExecutorAsync()`
+- Runs per-action `WhenAny` loop (request vs notification)
+- Dispatches to `HandleCallAsync` / `HandleReadAsync` / `HandleWriteAsync` based on `action.ActionType`
+- On handler exception: catches, logs, sends `ManagementActionApplicationError` with code `InternalError` back to invoker
+- On `ManagementActionUpdated`: pauses health, reports config status (using `Error` field), reports Available/Unavailable
+- On `ManagementActionUpdatedWithNewExecutor`: drains old executor, swaps to new, reports status
```
vaavva: If this is being implemented by the SDK, I'm not sure that's the best idea - there are multiple valid ways that an executor could be drained, and the connector application should be making that decision

doc/dev/tmp/management-action-implementation-design.md
```diff
@@ -0,0 +1,1842 @@
...
+3. When ADR reports that the asset was updated/deleted, `AssetUnavailableAsync` calls `cts.Cancel()` on *that* asset's token. The per-action loops observe it and unwind through their drain logic; the asset task completes; the SDK may then re-invoke `WhileAssetIsAvailable` with the new `Asset` snapshot.
+
+The **per-action** loop inside `ManagementActionConnectorWorker.RunActionLoopAsync` — which user code never sees — looks like this in shape:
+
+```csharp
+ManagementActionExecutor? executor = await assetClient.GetManagementActionExecutorAsync(group, action, ct);
+while (!ct.IsCancellationRequested)
+{
+    var recvRequestTask = executor.RecvRequestAsync(ct);
+    var recvNotificationTask = assetClient.RecvManagementActionNotificationAsync(group, action, ct);
```
vaavva: is there a way to bias this notification over receiving a request? If many requests are waiting to be handled, the notification should be handled first so that we can switch to the new definition. Otherwise, update notifications could be starved out

doc/dev/tmp/management-action-implementation-design.md
```diff
@@ -0,0 +1,1842 @@
...
+```
+
+That's the SDK's loop. The **user's** loop is conceptually just `HandleCallAsync(args, ct) => ...`.
+
+#### B3.2 Where `ManagementActionExecutor` actually lives
+
+`ManagementActionExecutor` wraps `CommandExecutor<byte[], byte[]>` from `Azure.Iot.Operations.Protocol` (with the existing `PassthroughSerializer`):
+
+- **One `CommandExecutor` per management action**, created by `AssetClient.GetManagementActionExecutorAsync` at asset-available time.
+- Its `RequestTopicPattern` is taken from `AssetManagementGroupAction.Topic` (falling back to `AssetManagementGroup.DefaultTopic`), with topic tokens resolved against the device / endpoint / asset context.
```
vaavva: for now, no topic tokens will be supported, and the defaultTopic shouldn't be used since it can't be supported properly without topic token support

doc/dev/tmp/management-action-implementation-design.md
```diff
@@ -0,0 +1,1842 @@
...
+          ──────────▶  robot-connector:2.0  (handles v7 + v8)
+Robot v8  ──────────▶  robot-connector:2.0
+```
+
+Operational consequences:
+
+1. **One binary can support multiple shapes.** Your factory's `(group, action) → handler` dispatch just needs entries for every `(shape, action)` combination you care about. Unknown pairs → return a sentinel handler that replies with `ManagementActionApplicationError { ErrorCode = "UnknownAction" }`.
+2. **Multiple binary versions can coexist.** Deploy `:2.0` alongside `:1.3` and migrate asset instances one at a time by editing each device's connector reference — standard blue/green, not AIO-specific.
+
+#### B6.4 The fence: what you cannot do without a code change
```
vaavva: Arguable - generic connectors generally don't care about the contents of the request/response payload, just send them to the endpoint and put the endpoint's response on the response. It's possible to know more or be more locked into specific payloads, but I wouldn't say it's the only option.

doc/dev/tmp/management-action-implementation-design.md
```diff
@@ -0,0 +1,1842 @@
...
+
+2. **Topic extraction:** How to derive the MQTT request topic from `AssetManagementGroupAction.Topic` / `AssetManagementGroup.DefaultTopic`? The Rust side has `try_executor_topic_from_management_topics()`. Need to understand the topic format and how it maps to `CommandExecutor`'s `RequestTopicPattern`.
+   - **RESOLVED: Dynamic topic configuration at runtime.** The topic comes from the asset definition: `AssetManagementGroupAction.Topic` with fallback to `AssetManagementGroup.DefaultTopic` (fallback not yet implemented in Rust, but should be in .NET). The `CommandExecutor` supports setting `RequestTopicPattern` at construction time without the `[CommandTopic]` attribute — so `ManagementActionExecutor` creates a `CommandExecutor<byte[], byte[]>` and configures topic properties dynamically. `TopicTokenMap` is populated with known values (device name, endpoint, etc.), and unresolved tokens become `+` (MQTT wildcard). The executor subscribes to `$share/{ServiceGroupId}/{resolved topic}`.
```
vaavva: fallback default topic should not be implemented by dotnet as it isn't supported on the invoker side of management actions. Topic tokens are also not supported by the invoker currently, so supporting them in the dotnet executor would be a no-op

dotnet/src/Azure.Iot.Operations.Connector/ManagementActionConnectorWorker.cs
```diff
@@ -0,0 +1,385 @@
...
+            CancellationToken cancellationToken)
+        {
+            switch (notification)
+            {
+                case ManagementActionUpdated updated:
+                    _logger.LogInformation(
+                        "{Group}::{Action} definition updated (same topic). error={Error}",
+                        groupName, actionName, updated.Error);
+                    await assetClient.PauseManagementActionRuntimeHealthReportingAsync(groupName, actionName, cancellationToken);
+                    await ReportActionConfigStatusAsync(assetClient, groupName, actionName, validationError: updated.Error, cancellationToken);
```
vaavva: this logic can't be fully contained in the SDK - the connector application may have additional config validations that they need to do, such as whether the endpoint target uri is valid or not

dotnet/src/Azure.Iot.Operations.Connector/ManagementActionConnectorWorker.cs
```diff
@@ -0,0 +1,385 @@
...
+                        await ReportActionAvailableAsync(assetClient, groupName, actionName, cancellationToken);
+                    }
+                    else
+                    {
+                        await ReportActionUnavailableAsync(assetClient, groupName, actionName, updatedWithNew.Error.Message, cancellationToken);
+                    }
+                    return false;
+
+                case ManagementActionAssetUpdated assetUpdated:
+                    _logger.LogInformation(
```
vaavva: if the asset is updated, statuses need to be re-reported as well as they will have gotten cleared (but again, this should be managed by the application, not the SDK)

dotnet/src/Azure.Iot.Operations.Connector/ManagementActionConnectorWorker.cs
```diff
@@ -0,0 +1,385 @@
...
+                cancellationToken: cancellationToken);
+
+        private static Task ReportActionUnavailableAsync(
+            AssetClient assetClient, string groupName, string actionName, string? message, CancellationToken cancellationToken)
+            => assetClient.ReportManagementActionRuntimeHealthAsync(
+                groupName, actionName,
+                new ConnectorRuntimeHealth
+                {
+                    Status = HealthStatus.Unavailable,
+                    ReasonCode = "ConfigError",
```
vaavva: config errors shouldn't directly be reported as unhealthy events - they should cause all health reporting to stop so that the health status goes to Unknown. Reasoning is that we don't know for certain there is any health issue, we just can't attempt to find out because of the config error

dotnet/src/Azure.Iot.Operations.Connector/ManagementActionNotification.cs
```diff
@@ -0,0 +1,68 @@
...
+    /// Non-null if the updated definition failed validation.
+    /// </param>
+    public sealed record ManagementActionUpdatedWithNewExecutor(
+        ManagementActionExecutor? NewExecutor,
+        ConfigError? Error) : ManagementActionNotification;
+
+    /// <summary>
+    /// The parent asset was updated but this specific management action's definition
+    /// is unchanged. The connector may need to re-evaluate surrounding state
+    /// (device context, other actions on the asset).
```
vaavva: this notification wouldn't be sent on device changes. Other actions probably shouldn't interfere with each other either
```suggestion
    /// (asset context, asset defaults).
```

dotnet/src/Azure.Iot.Operations.Connector/ManagementActionNotSupportedException.cs
```diff
@@ -0,0 +1,56 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Azure.Iot.Operations.Connector
+{
+    /// <summary>
+    /// Thrown by an <see cref="IManagementActionHandler"/> method to signal that the handler
+    /// is not wired up to service this action type (for example, a handler implementing only
```
vaavva: this scenario where an action type isn't supported should be reported on the config when the definition is received (returning it on requests still could make sense in addition though)

dotnet/src/Azure.Iot.Operations.Connector/ManagementActionResponse.cs
```diff
@@ -0,0 +1,43 @@
...
+    public record ManagementActionResponse
+    {
+        /// <summary>Serialized response payload.</summary>
+        public required ReadOnlySequence<byte> Payload { get; init; }
+
+        /// <summary>MIME type of <see cref="Payload"/> (e.g. <c>application/json</c>).</summary>
+        public required string ContentType { get; init; }
+
+        /// <summary>Optional CloudEvent metadata to attach to the response.</summary>
+        public required CloudEvent? CloudEvent { get; init; }
```
vaavva: not sure if this needs to be required, it can default to null

dotnet/test/Azure.Iot.Operations.Connector.IntegrationTests/ManagementActionConnectorTests.cs
```diff
@@ -0,0 +1,311 @@
...
+
+            Assert.Fail($"ReadTemperature never returned unit '{expectedUnit}' after WriteConfiguration.");
+        }
+
+        /// <summary>
+        /// Performs an MQTT 5 RPC call: subscribes to a unique response topic,
+        /// publishes the request with <c>ResponseTopic</c> + <c>CorrelationData</c>,
+        /// and returns the matching response message.
+        /// </summary>
+        private static async Task<MqttApplicationMessage> InvokeAsync(
```
vaavva: why not use an invoker for this test instead of doing the invoker logic manually here? Better to test with a more real scenario

dotnet/src/Azure.Iot.Operations.Connector/ManagementActionExecutor.cs
timtay-microsoft:
```csharp
interface IManagementHandler
{
    // What you currently have
    Task<ManagementActionResponse> HandleCallAsync(ManagementActionInvokedEventArgs args, CancellationToken cancellationToken);
}
```

vs 

```csharp
interface IManagementHandler
{
    // What may be easier to define since it just wraps the ManagementActionHandler class which uses funcs like this
    Func<ManagementActionInvokedEventArgs, CancellationToken, Task<ManagementActionResponse>>? OnCallRequestReceived { get; set; }
}
```

timtay-microsoft: Looking at this again now, though, I'm less convinced of my position when I started this thread. Feel free to take whichever approach you feel is easiest to implement

dotnet/src/Azure.Iot.Operations.Connector/IManagementActionHandler.cs
timtay-microsoft: Gotcha, sounds like I assumed too much here, then. Go ahead and disregard the original comment here.

That said, if we want to treat call/read/write management action invocations the same (but for the "call/read/write" type), then it probably makes sense to collapse these three callbacks (HandleReadAsync/HandleWriteAsync/HandleCallAsync) into one and just have the type be a part of the EventArgs param.

doc/dev/tmp/management-action-design-onepager.md
maximsemenov80: In another thread we agreed that we will not have different call backs since all signatures will be identical.

dotnet/src/Azure.Iot.Operations.Connector/ManagementActionConnectorWorker.cs
```diff
@@ -0,0 +1,372 @@
...
+    /// internalize all per-action lifecycle management (executor acquisition, notification
+    /// handling, drain-and-dispose, health/config reporting). User code only needs to provide
+    /// an <see cref="IManagementActionHandlerFactory"/> whose handlers implement the actual
+    /// device communication.
+    /// </summary>
+    /// <remarks>
+    /// Mirrors the pattern of <see cref="PollingTelemetryConnectorWorker"/> which internalizes
+    /// polling logic and delegates sampling to <see cref="IDatasetSampler"/>.
+    /// </remarks>
+    public class ManagementActionConnectorWorker : ConnectorWorker
```
timtay-microsoft: Is there any reason that this has a dedicated class? Could this management action logic just be added to the ConnectorWorker class?

dotnet/src/Azure.Iot.Operations.Connector/ManagementActionConnectorWorker.cs
```diff
@@ -0,0 +1,372 @@
...
+        }
+
+        private async Task WhileAssetAvailableAsync(AssetAvailableEventArgs args, CancellationToken cancellationToken)
+        {
+            _logger.LogInformation(
+                "Asset {AssetName} on device {DeviceName} is available.",
+                args.AssetName,
+                args.DeviceName);
+
+            var adrClient = _adrClient ?? throw new InvalidOperationException("ADR client not initialized.");
```
timtay-microsoft: The "WhileAssetAvailable" callback here can't be executed until the ADR client has been instantiated, so feel free to just 

```csharp
var adrClient = _adrClient!;
```

dotnet/src/Azure.Iot.Operations.Connector/ManagementActionConnectorWorker.cs
```diff
@@ -0,0 +1,372 @@
...
+        /// callback so incoming requests are dispatched to the handler in <paramref name="ctx"/>
+        /// via <see cref="InvokeHandlerAsync"/>. No-op if <paramref name="executor"/> is null
+        /// (the action's current definition is invalid; the worker is waiting for the next
+        /// notification to swap it in).
+        /// </summary>
+        private void WireCallback(ManagementActionExecutor? executor, ActionContext ctx)
+        {
+            if (executor is null) return;
+
+            executor.OnRequestReceived = (args, ct) =>
```
timtay-microsoft, line 344: You'll need to tear this callback down when tearing down this connector. Typically, you'd do something like

```csharp
public void WireCallback(...)
{
    executor.OnRequestReceived += (args, ct) => SomeHandlerFunction
}

public async DisposeAsync()
{
    executor.OnRequestReceived -= (args, ct) => SomeHandlerFunction
}
```

dotnet/src/Azure.Iot.Operations.Connector/ManagementActionConnectorWorker.cs
timtay-microsoft: I'm guessing this class only does something special when it detects a management group w/ a management action, so it should be safe to add to the base class without breaking existing behavior

dotnet/samples/Connectors/ManagementActionConnector/Handlers/SampleManagementActionHandlerFactory.cs
```diff
@@ -0,0 +1,101 @@
...
+            logger.LogInformation(
+                "Creating handler for action {Action} (type={ActionType}, targetUri={Target})",
+                action.Name, action.ActionType, action.TargetUri);
+
+            return action.Name switch
+            {
+                "reboot" => new RebootHandler(logger, _device, statusReporter),
+                "read-temperature" => new ReadTemperatureHandler(logger, _device),
+                "write-configuration" => new WriteConfigurationHandler(logger, _device),
+                _ => new UnknownActionHandler(logger, action.Name),
```
timtay-microsoft: Weren't you going to replace this with some dedicated exception type? So that this method would throw, and the calling method in our connector package would understand that no handler is present for that action?

timtay-microsoft: It would be nice if we didn't ask users to define ```UnknownActionHandler``` like this

dotnet/src/Azure.Iot.Operations.Connector/IManagementActionHandlerFactory.cs
```diff
@@ -0,0 +1,93 @@
...
+        /// </remarks>
+        ValueTask<ConfigError?> ValidateConfigurationAsync(
+            Device device,
+            string inboundEndpointName,
+            Asset asset,
+            AssetManagementGroupAction action,
+            CancellationToken cancellationToken)
+            => ValueTask.FromResult<ConfigError?>(null);
+
+        bool SupportsAction(AssetManagementGroupAction action);
```
timtay-microsoft, line 95: This method could use an xml doc explaining it

dotnet/src/Azure.Iot.Operations.Connector/ConnectorWorker.cs
```diff
@@ -734,28 +744,79 @@ private void AssetAvailable(string deviceName, string inboundEndpointName, Asset
...
+                    // management-action loop. Disposed once when both branches return.
+                    await using AssetAvailableEventArgs args = new(deviceName, device, inboundEndpointName, assetName, asset, _leaderElectionClient, _adrClient!, this);
+
+                    var assetBranches = new List<Task>(2);
+
+                    // Built-in: management-action handling. Runs whenever a factory was supplied,
+                    // independent of whatever the user assigned to WhileAssetIsAvailable. This is
+                    // intentionally not routed through the public field so a `= WhileAssetIsAvailable`
+                    // assignment from user code can't disable management actions.
+                    if (_managementActionOrchestrator != null)
```
timtay-microsoft: Does it make sense to use the management action orchestrator here even if there are no management actions defined in the asset definition?

timtay-microsoft: Or should this check be more like 

```csharp
if (_managementActionOrchestrator != null && asset.ContainsAtLeastOneManagementAction)

```

dotnet/src/Azure.Iot.Operations.Connector/ConnectorWorker.cs
```diff
@@ -734,28 +744,79 @@ private void AssetAvailable(string deviceName, string inboundEndpointName, Asset
...
+                            "ManagementActionHandling",
+                            ct => _managementActionOrchestrator.ServeActionsWhileAssetIsAvailableAsync(args, ct),
+                            assetTaskCancellationTokenSource.Token,
+                            assetName, deviceName, inboundEndpointName));
                     }
-                    catch (OperationCanceledException)
+
+                    // User-supplied per-asset callback. Captured into a local to avoid races with
+                    // a concurrent reassignment of the field.
+                    var whileAssetIsAvailable = WhileAssetIsAvailable;
```
timtay-microsoft: nit: You probably don't need to assign a local variable with this value

```csharp
var whileAssetIsAvailable = WhileAssetIsAvailable;
if (whileAssetIsAvailable != null)
{
    assetBranches.Add(SafeInvokeAssetBranchAsync(
    "UserWhileAssetIsAvailableCallback",
    ct => whileAssetIsAvailable.Invoke(args, ct),
    assetTaskCancellationTokenSource.Token,
    assetName, deviceName, inboundEndpointName));
}
```

```csharp
if (WhileAssetIsAvailable != null)
{
    assetBranches.Add(SafeInvokeAssetBranchAsync(
    "UserWhileAssetIsAvailableCallback",
    ct => WhileAssetIsAvailable.Invoke(args, ct),
    assetTaskCancellationTokenSource.Token,
    assetName, deviceName, inboundEndpointName));
}
```

dotnet/src/Azure.Iot.Operations.Connector/ManagementActionStatusReporter.cs
```diff
@@ -0,0 +1,68 @@
...
+
+namespace Azure.Iot.Operations.Connector
+{
+    /// <summary>
+    /// Default <see cref="IManagementActionStatusReporter"/> implementation that delegates
+    /// to an <see cref="AssetClient"/>. Constructed by <see cref="ConnectorWorker"/>
+    /// once per action and shared between the worker's notification loop and the user's
+    /// handler (via the factory).
+    /// </summary>
+    internal sealed class ManagementActionStatusReporter : IManagementActionStatusReporter
```
timtay-microsoft: I'd like to avoid creating status reporter clients (and their interface) for sub-asset level resources (management actions, events, datasets, etc) if possible. That was the pattern used in Rust, but this SDK more closely follows the service APIs in that it only allows you to report status for the device and/or asset as a whole. Adding sub-asset clients increases the complexity of this SDK and adds redundant APIs since users already have access to the assetClient action runtime health APIs used in this client below. 

I'd consider removing this class and interface entirely. Any conveniences added by this client can be moved to the asset client runtime health surface instead.

dotnet/src/Azure.Iot.Operations.Connector/AssetClient.cs
```diff
@@ -396,6 +396,58 @@ private async Task<AssetStatus> GetAssetStatusAsync(
...
+            CancellationToken cancellationToken = default)
+            => throw new NotImplementedException();
+
+        /// <summary>
+        /// Await the next lifecycle notification for <paramref name="managementGroupName"/> /
+        /// <paramref name="managementActionName"/>. Exits (returns
+        /// <see cref="ManagementActionDeleted"/>) when the action is removed from the asset
+        /// definition or the asset itself is deleted.
+        /// </summary>
+        public Task<ManagementActionNotification> RecvManagementActionNotificationAsync(
```
timtay-microsoft: I'm a bit confused on this method and ```GetManagementActionExecutorAsync``` being public. As far as I can tell, they are used only internally by the orchestrator. Could these be made internal?

dotnet/src/Azure.Iot.Operations.Connector/ManagementActionApplicationError.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// Set on <see cref="ManagementActionResponse.ApplicationError"/> when the action was
+    /// executed but the outcome is a domain-level failure (vs. a transport error).
+    /// </summary>
+    public record ManagementActionApplicationError
+    {
+        /// <summary>Caller-facing error code (connector-defined).</summary>
+        public required string ErrorCode { get; set; }
+
+        /// <summary>Human-readable payload describing the error. May be empty.</summary>
+        public string ErrorPayload { get; set; } = string.Empty;
```
timtay-microsoft: I'd consider just making this nullable to signal that it is optional rather than using empty string
```suggestion
        public string? ErrorPayload { get; set; }
```

timtay-microsoft: And

```csharp
        /// <summary>Human-readable payload describing the error. May be null.</summary>
```

dotnet/src/Azure.Iot.Operations.Connector/ManagementActionExecutor.cs
```diff
@@ -0,0 +1,89 @@
...
+    /// and the callback will be invoked once per incoming management action request.
+    /// The returned <see cref="ManagementActionResponse"/> is sent back to the invoker.
+    /// </para>
+    /// <para>
+    /// Exceptions thrown by the callback are surfaced to the invoker as
+    /// <see cref="ManagementActionApplicationError"/> responses by the connector worker; see
+    /// <see cref="ConnectorWorker"/>.
+    /// </para>
+    /// </remarks>
+    public sealed class ManagementActionExecutor : IAsyncDisposable
```
timtay-microsoft: I don't see any references to this class outside of the connector package, so could it be internal?

dotnet/src/Azure.Iot.Operations.Connector/ManagementActionNotification.cs
```diff
@@ -0,0 +1,68 @@
...
+    /// <see cref="ManagementActionAssetUpdated"/>, <see cref="ManagementActionDeleted"/>)
+    /// mirror the Rust SDK's <c>DataOperationNotification</c> (Updated / AssetUpdated / Deleted).
+    /// <see cref="ManagementActionUpdatedWithNewExecutor"/> is management-action-specific
+    /// (datasets/events/streams don't have executors).
+    /// <para/>
+    /// TODO: When per-component notifications are added for datasets, events, and streams,
+    /// consider extracting the shared variants into a common base or generic type (e.g.
+    /// <c>ComponentNotification&lt;T&gt;</c>) so all component types share a consistent shape.
+    /// </remarks>
+    public abstract record ManagementActionNotification;
```
timtay-microsoft: Could this be internal as well?

dotnet/src/Azure.Iot.Operations.Connector/ManagementActionOrchestrator.cs
```diff
@@ -0,0 +1,477 @@
...
+
+        /// <summary>
+        /// Entry point — invoked once per <see cref="AssetAvailableEventArgs"/>. Runs every
+        /// management-action loop declared on the asset until the asset goes away (the token
+        /// is cancelled) or every action has been deleted.
+        /// </summary>
+        public async Task ServeActionsWhileAssetIsAvailableAsync(AssetAvailableEventArgs args, CancellationToken cancellationToken)
+        {
+            _logger.LogInformation(
+                "Asset {AssetName} on device {DeviceName} is available.",
```
timtay-microsoft: nit: This log is probably redundant since the ConnectorWorker class should log something similar

dotnet/src/Azure.Iot.Operations.Connector/ManagementActionOrchestrator.cs
```diff
@@ -0,0 +1,477 @@
...
+            ConfigError BuildError() => new()
+            {
+                Code = "UnsupportedAction",
+                Message = $"No handler available for action '{groupName}::{actionName}'.",
+            };
+
+            async Task ReReportAsync()
+            {
+                _logger.LogDebug(
+                    "Unsupported action {Group}::{Action} updated; re-reporting UnsupportedAction config error.",
```
timtay-microsoft: nit: I'd include the device name and asset name in these kinds of logs in this class. 

Users may have hundreds of devices, so adding that context could save them some grief in finding which device was misconfigured

dotnet/src/Azure.Iot.Operations.Connector/ManagementActionOrchestrator.cs
```diff
@@ -0,0 +1,477 @@
...
+
+                        case ManagementActionDeleted:
+                            _logger.LogInformation(
+                                "Unsupported action {Group}::{Action} deleted; ending unsupported-action loop.",
+                                groupName, actionName);
+                            return;
+
+                        default:
+                            _logger.LogWarning(
+                                "Unsupported action {Group}::{Action} received unknown notification type {Type}",
```
timtay-microsoft: nit: include device and asset name here as well

dotnet/samples/Connectors/ManagementActionConnector/connector-metadata.json
```diff
@@ -0,0 +1,72 @@
+{
```
timtay-microsoft, line 1: You'll want to add this new connector metadata doc as part of the ci-connector-schema.yml file which validates that all connector metadata files in our repo satisfy the schema:

https://github.com/Azure/iot-operations-sdks/blob/85f639007581990dd9459b7e68b784b2d55921a5/.github/workflows/ci-connector-schema.yml#L41

dotnet/samples/Connectors/ManagementActionConnector/connector-metadata.json
```diff
@@ -0,0 +1,72 @@
...
+  "version": "1.0.0",
+  "maintainer": "Azure IoT Operations, for sample use only",
+  "imageConfigurationSettings": {
+    "imageName": "azureiotoperations/sample-akri-connectors/management-action",
+    "tag": "1.0.0"
+  },
+  "supportedArchitectures": [
+    "linux/amd64"
+  ],
+  "sourceCode": {
```
timtay-microsoft: Feel free to just omit this "sourceCode" block entirely. The schema defines this field mostly so that, if a user files a bug on a shipped connector, we know what package versions they were using. In this case, we don't intend to ship this connector.

dotnet/samples/Connectors/ManagementActionConnector/connector-metadata.json
timtay-microsoft: Whoops, just noticed you already did this. Nevermind.

dotnet/samples/Connectors/ManagementActionConnector/KubernetesResources/mgmt-action-asset-definition.yaml
```diff
@@ -0,0 +1,53 @@
...
+  name: my-mgmt-action-asset
+  namespace: azure-iot-operations
+spec:
+  uuid: 3456-7890-1234-5678
+  attributes:
+    assetId: my-mgmt-action-asset
+    assetType: mgmt-action-sample
+  # -----------------------------------------------------------------
+  # Single management group with three actions (one of each ActionType:
+  # Call, Read, Write), intended for exercising the upcoming .NET
```
timtay-microsoft: nit: Remove "upcoming" since we will release this about as soon as you merge this PR

timtay-microsoft (commented):
Looks good to me so far! Feel free to continue on to implementing based on the API you have here right now.

dotnet/test/Azure.Iot.Operations.Connector.IntegrationTests/ManagementActionConnectorTests.cs
```diff
@@ -0,0 +1,536 @@
...
+                    requestPayload: Array.Empty<byte>());
+
+                MgmtTemperatureReading body = DeserializeBody<MgmtTemperatureReading>(responseBytes);
+                _output.WriteLine($"[ReadTemperature_Read] parsed body: Value={body.Value}, Unit={body.Unit}");
+                Assert.True(body.Value > -100 && body.Value < 200, $"unexpected temperature {body.Value}");
+                Assert.True(body.Unit is "C" or "F");
+            }
+            catch (AkriMqttException ex) when (IsApplicationError(ex))
+            {
+                // The sample's reboot Call action puts the FakeDevice into a
```
timtay-microsoft: This seems like a race condition we don't want to happen in our integration tests. Would running these tests in serial solve this? That way the test device only reboots during the specific tests that run the reboot management action? 

Alternatively, should we replace the "reboot" management action with an action that wouldn't interfere with running parallel management actions?

dotnet/src/Azure.Iot.Operations.Services/AssetAndDeviceRegistry/Models/DeviceStatus.cs
```diff
@@ -61,4 +61,11 @@ public bool EqualTo(DeviceStatus other)
 
         return true;
     }
+
+    /// <summary>
+    /// Create a deep copy of this device status, including all nested objects.
+    /// </summary>
+    /// <returns>A new <see cref="DeviceStatus"/> instance whose nested objects are independent from this instance.</returns>
+    public DeviceStatus DeepClone()
```
timtay-microsoft: It looks like the only use of this is internal, so could this method also be internal?

dotnet/src/Azure.Iot.Operations.Services/AssetAndDeviceRegistry/Models/AssetStatus.cs
```diff
@@ -306,4 +316,11 @@ public bool EqualTo(AssetStatus other)
 
         return true;
     }
+
+    /// <summary>
+    /// Create a deep copy of this asset status, including all nested collections and objects.
+    /// </summary>
+    /// <returns>A new <see cref="AssetStatus"/> instance whose nested objects are independent from this instance.</returns>
+    public AssetStatus DeepClone()
```
timtay-microsoft: Same question here

dotnet/src/Azure.Iot.Operations.Connector/ConnectorWorker.cs
```diff
@@ -27,21 +26,37 @@ public class ConnectorWorker : ConnectorBackgroundService
...
         protected readonly ILogger<ConnectorWorker> _logger;
         private readonly IMqttClient _mqttClient;
-        private readonly ApplicationContext _applicationContext;
         private readonly IAzureDeviceRegistryClientWrapperProvider _adrClientWrapperFactory;
         protected IAzureDeviceRegistryClientWrapper? _adrClient;
         private readonly IMessageSchemaProvider _messageSchemaProviderFactory;
         private LeaderElectionClient? _leaderElectionClient;
         private readonly ConcurrentDictionary<string, DeviceContext> _devices = new();
+
+        // Asset-available notifications can arrive before the device-available notification that populates
```
timtay-microsoft: If you are seeing this behavior, that may signal a bug in the AssetFileMonitor which is the source of these [asset-available](https://github.com/Azure/iot-operations-sdks/blob/85f639007581990dd9459b7e68b784b2d55921a5/dotnet/src/Azure.Iot.Operations.Connector/Files/AssetFileMonitor.cs#L39) and [device-available](https://github.com/Azure/iot-operations-sdks/blob/85f639007581990dd9459b7e68b784b2d55921a5/dotnet/src/Azure.Iot.Operations.Connector/Files/AssetFileMonitor.cs#L42) events.

I'd rather we try to fix this behavior from the AssetFileMonitor level rather than this level. That way, any user who tries to build their own base connector-like class would get that bug fix

timtay-microsoft (approved):
Some minor comments, but otherwise LGTM

timtay-microsoft (approved):

Issue #1320, opened 2026-04-29
[BUG] Asset update causes duplicate callbacks and task leak


dagassmann:
### Product

.NET SDK

### Azure Iot Operations version

1.3.38

### What happened?

When an asset is updated (ChangeType.Updated), ConnectorWorker does not cancel or await the previous 
WhileAssetIsAvailable task before starting a new one.
 
 AssetUnavailableAsync(isUpdating: true) skips all cleanup — the old task keeps running. Then AssetAvailable calls 
_assetTasks.TryAdd() which fails silently because the old entry was never removed.
 
 Result: both old and new WhileAssetIsAvailable callbacks run concurrently for the same asset, causing duplicate 
message processing. Additionally, the new task's CancellationTokenSource is lost (never stored), creating a task leak 
that prevents clean shutdown.

### List the steps to reproduce

1. Register a WhileAssetIsAvailable callback 
2. Create an asset (ChangeType.Created) — the callback starts
3. Update the asset (ChangeType.Updated) — e.g. change any asset property
4. Observe that the old callback is still running (never cancelled) and a second callback starts
5. Both callbacks are now active

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

timtay-microsoft:
Thanks for filing the bug! #1323 should fix this for you

timtay-microsoft:
This issue is now fixed as of connector package version 1.1.1, so I'll close this thread

PR #1323, opened 2026-04-30
[.NET] Fix bug where updating an asset would not cancel previously invoked 'WhileAssetIsAvailable' callback

timtay-microsoft:
#1320


dotnet/src/Azure.Iot.Operations.Connector/ConnectorWorker.cs
```diff
@@ -756,27 +756,28 @@ private void AssetAvailable(string deviceName, string inboundEndpointName, Asset
             }
         }
 
-        private async Task AssetUnavailableAsync(string deviceName, string inboundEndpointName, string assetName, bool isUpdating)
+        private async Task AssetUnavailableAsync(string deviceName, string inboundEndpointName, string assetName)
```
timtay-microsoft, line 759: Previously, the 'isUpdating' flag would signal to not cancel any previously running callbacks tied to this asset. This was not the correct behavior

maximsemenov80 (approved):

maximsemenov80 (approved):

PR #1325, opened 2026-05-01
Codegen when action has external output schema and error response

jrdouceur:


vaavva (approved):
Approving with addition of get_err fn getting generated for these fixed invokers

vaavva (approved):

timtay-microsoft (approved):

PR #1328, opened 2026-05-06
[Codegen + .NET] Fix issue where wot codegen'd telemetry sender did not wrap non-custom topic tokens with "ex:..."

timtay-microsoft:
Rust [does this](https://github.com/Azure/iot-operations-sdks/blob/8993ab15004cbd13d164e5290d57949b78a0406c/wot-codegen/src/Azure.Iot.Operations.EnvoyGenerator/rust/Events/t4/RustTelemetrySender.tt#L51), but .NET did not

As a consequence, telemetry sender APIs in codegen'd services like ADR didn't actually work correctly because the telemetry sender base class expected to see a topic token key like "ex:connectorClientId" but the codegen layer would pass down a topic token key like "connectorClientId"

### DIFF
--- a/dotnet/samples/Protocol/TestThing/Counter/EventCollectionSender.g.cs
+++ b/dotnet/samples/Protocol/TestThing/Counter/EventCollectionSender.g.cs
@@ -24,6 +24,15 @@ namespace TestThing.Counter
             public EventCollectionSender(ApplicationContext applicationContext, IMqttPubSubClient mqttClient)
                 : base(applicationContext, mqttClient, new Utf8JsonSerializer())
             {
+                Dictionary<string, string> TopicTokenMapWithExPrefix = new();
+                foreach (string key in TopicTokenMap.Keys)
+                {
+                    string keyWithPrefix = $"ex:{{key}}";
+                    TopicTokenMapWithExPrefix[keyWithPrefix] = TopicTokenMap[key];
+                }
+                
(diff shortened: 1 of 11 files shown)


tools/azure_iot_operations_stub_services/src/schema_registry/schema_registry_gen/schema_registry/get_command_executor.rs
```diff
@@ -2,7 +2,7 @@
 
 use std::collections::HashMap;
 
-use azure_iot_operations_mqtt::interface::ManagedClient;
+use azure_iot_operations_mqtt::session::SessionManagedClient;
```
vaavva: can you revert the updates to these files - this is a stable tool that we haven't had time to update to the latest SDKs yet, but updating just the codegen breaks it without the corresponding updates in the non-generated code

timtay-microsoft: I will, yeah

timtay-microsoft: The genAll.sh script is just picking this up when I try to run codegen for the whole repo

timtay-microsoft:
Actually, it looks like codegen is doing the right thing here already. The handwritten layer on top is just not calling the correct method in the codegen layer. Closing this to rework the fix here

PR #1330, opened 2026-05-11
Disable kTLS

gordonwang0:
The kTLS client implementation does not properly handle TLS session tickets sent by the server. Disable kTLS until we have a fix.


cartertinney:
Closing because CI can't run PR's from forks. Reimplemented in #1333

PR #1332, opened 2026-05-12
[ADR] mRPC Back Pressure flag

maximsemenov80:
This pull request adds a new Architecture Decision Record (ADR 31) describing how the SDK will support the MQ broker's new high-priority backpressure-bypass mechanism for control-plane (ex.: State Store) traffic.


doc/dev/adr/0031-backpressure-bypass.md
```diff
@@ -0,0 +1,102 @@
...
+rejects the property, or the CRD kill switch is on), the request
+travels without the property and the executor has nothing to mirror
+&mdash; both legs fall back to normal-priority backpressure with no
+SDK changes.
+
+## Decision
+
+### Wire
+
+- The example name `$high_priority` is broker-owned and sits outside
```
timtay-microsoft: Just for my understanding, is this the field that broker folks settled on? Or are you proposing it here?

doc/dev/adr/0031-backpressure-bypass.md
```diff
@@ -0,0 +1,102 @@
...
+- No protocol-version bump. Brokers that don't recognize the property
+  (or have the kill switch on) treat it as opaque &mdash; safe fallback
+  to normal-priority backpressure.
+- Existing SDK consumers will see their mRPC traffic marked
+  `$high_priority` after upgrading. This is intentional and aligned
+  with the MQ ADR. The broker's authz policy and CRD kill switch are
+  the operator-side controls if a deployment needs to claw the
+  capability back.
+
+[mq-adr]: https://msazure.visualstudio.com/One/_git/Azure-MQ?path=/docs-dev/adr/dmqtt/0093-backpressure-bypass.md&_a=preview
```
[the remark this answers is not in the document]
timtay-microsoft: +1 to this copilot comment. We can't leave links to internal repos in public docs like this.

doc/dev/adr/0031-backpressure-bypass.md
```diff
@@ -0,0 +1,102 @@
...
+**user property on each PUBLISH** (broker owns the exact name, e.g.
+`$high_priority`); the broker also gets a CRD kill switch and an authz
+policy gating who may set the flag.
+
+The flag is set by the *publisher* of each PUBLISH &mdash; the broker
+does not infer it. The MQ ADR notes *"we expect the mRPC code generator
+to set the property in requests and responses"*. We follow that
+guidance: every mRPC PUBLISH the SDK produces carries the flag by
+default, and the SDK exposes an option to turn it off. The motivation
+is that mRPC is the SDK's control-plane RPC layer (and is used heavily
```
timtay-microsoft: The wording in this block is a bit confusing. "mRpc is the SDK's control-plane RPC layer"?

doc/dev/adr/0031-backpressure-bypass.md
```diff
@@ -0,0 +1,102 @@
...
+## Context
+
+The MQ broker [is adding][mq-adr] a high-priority backpressure-bypass
+mechanism so control-plane traffic (mRPC, State Store) is not starved when
+data-plane traffic fills the broker's buffer pool. The mark is an MQTT 5
+**user property on each PUBLISH** (broker owns the exact name, e.g.
+`$high_priority`); the broker also gets a CRD kill switch and an authz
+policy gating who may set the flag.
+
+The flag is set by the *publisher* of each PUBLISH &mdash; the broker
```
timtay-microsoft: This kind of detail probably belongs in a "Proposal" section, not a "context" section.

doc/dev/adr/0031-backpressure-bypass.md
```diff
@@ -0,0 +1,102 @@
...
+```mermaid
+sequenceDiagram
+    autonumber
+    participant Inv as Command Invoker<br/>(SDK)
+    participant Brk as MQ Broker
+    participant Exe as Command Executor<br/>(SDK)
+
+    Note over Inv: bypass option = true (default)
+    Inv->>Brk: PUBLISH request<br/>user-property: $high_priority
+    Note over Brk: authz check on $high_priority<br/>+ higher BP threshold applied
```
timtay-microsoft: I'm not sure I would bother enumerating broker behavior in this doc. We run the risk of it falling out of date or being incorrect in the first place

doc/dev/adr/0031-backpressure-bypass.md
timtay-microsoft: If the broker folks haven't settled on a user property name yet, I would prefer if we did include the "__" prefix to signal that it is reserved

doc/dev/adr/0031-backpressure-bypass.md
```diff
@@ -0,0 +1,102 @@
...
+to set the property in requests and responses"*. We follow that
+guidance: every mRPC PUBLISH the SDK produces carries the flag by
+default, and the SDK exposes an option to turn it off. The motivation
+is that mRPC is the SDK's control-plane RPC layer (and is used heavily
+by first-party services, including AI scenarios); having a single
+customer-tunable escape hatch is preferred over an opt-in-everywhere
+design that risks leaving important callers behind under load.
+
+This ADR specifies how the SDKs expose and set the flag. It does not
+change broker semantics.
```
timtay-microsoft: I have no idea what this is trying to say

doc/dev/adr/0031-backpressure-bypass.md
```diff
@@ -0,0 +1,102 @@
...
+    Exe->>Brk: PUBLISH response<br/>user-property: $high_priority
+    Note over Brk: same authz + threshold<br/>applied to response
+    Brk->>Inv: deliver response
+```
+
+If the customer turns the invoker's option off (or the broker's authz
+rejects the property, or the CRD kill switch is on), the request
+travels without the property and the executor has nothing to mirror
+&mdash; both legs fall back to normal-priority backpressure with no
+SDK changes.
```
timtay-microsoft: One detail missing here is what the MQTT user property value where the name is "$high_priority". Is it "True"/"False", empty string?

doc/dev/adr/0031-backpressure-bypass.md
timtay-microsoft: It should be fine for us to broadly say "MQTT publishes with this flag set will be given higher priority than other MQTT publishes", but I wouldn't go further than that.

doc/dev/adr/0031-backpressure-bypass.md
```diff
@@ -0,0 +1,102 @@
+# ADR 31: MQ Backpressure Bypass for SDK Traffic
+
+> **Status:** Proposed.
+
+## Context
+
+The MQ broker [is adding][mq-adr] a high-priority backpressure-bypass
+mechanism so control-plane traffic (mRPC, State Store) is not starved when
```
vaavva: nit that state store is mRPC traffic

doc/dev/adr/0031-backpressure-bypass.md
```diff
@@ -0,0 +1,102 @@
...
+data-plane traffic fills the broker's buffer pool. The mark is an MQTT 5
+**user property on each PUBLISH** (broker owns the exact name, e.g.
+`$high_priority`); the broker also gets a CRD kill switch and an authz
+policy gating who may set the flag.
+
+The flag is set by the *publisher* of each PUBLISH &mdash; the broker
+does not infer it. The MQ ADR notes *"we expect the mRPC code generator
```
vaavva: From previous discussion, I believe that the desire is for this to be set by default for all RPC requests/responses, so I would leave codegen out of having an opinion about this. The RPC invoker/executor should have the overridable default, and codegen'd code should just expose allowing overriding the default, maintaining consistent behavior between codegen and non-codegen uses

doc/dev/adr/0031-backpressure-bypass.md
```diff
@@ -0,0 +1,102 @@
...
+to set the property in requests and responses"*. We follow that
+guidance: every mRPC PUBLISH the SDK produces carries the flag by
+default, and the SDK exposes an option to turn it off. The motivation
+is that mRPC is the SDK's control-plane RPC layer (and is used heavily
+by first-party services, including AI scenarios); having a single
+customer-tunable escape hatch is preferred over an opt-in-everywhere
+design that risks leaving important callers behind under load.
```
vaavva: just to keep only relevant things here, I might revise this paragraph more like

```
The flag is set by the publisher of each PUBLISH — the broker does not infer it. Every mRPC PUBLISH the SDK produces carries the flag by default, and the SDK exposes an option to turn it off per message. The motivation is that having a single customer-tunable escape hatch is preferred over an opt-in-everywhere design that risks leaving important callers behind under load. Note that this only applies to RPC Publishes, telemetry and raw MQTT publishes do not send this flag by default.
```

doc/dev/adr/0031-backpressure-bypass.md
```diff
@@ -0,0 +1,102 @@
...
+    participant Inv as Command Invoker<br/>(SDK)
+    participant Brk as MQ Broker
+    participant Exe as Command Executor<br/>(SDK)
+
+    Note over Inv: bypass option = true (default)
+    Inv->>Brk: PUBLISH request<br/>user-property: $high_priority
+    Note over Brk: authz check on $high_priority<br/>+ higher BP threshold applied
+    Brk->>Exe: deliver request<br/>(property preserved)
+
+    Note over Exe: mirror: copy $high_priority<br/>from request onto response
```
vaavva: For simplicity of implementation and consistency about when the user opts out, I would probably not have this informed by whether the flag was on for the request or not and have the same pattern of high priority on by default and the user can opt out if they desire. That way, if they always do nothing, it's always present rather than only being added if it was on the request. While we probably do want the flag to always be set on the response if it's on the request, that seems like it would complicate the logic of the user override, and will likely happen almost all of the time without our intervention.
```suggestion
    Note over Exe: bypass option = true (default)
```

vaavva: following up on this comment after reading your explanation of this below the diagram, I don't think there's any harm in the executor setting the bypass flag even if the invoker couldn't (and in fact, I think that would be desirable to have the response have priority even if the invoker didn't have permission to request priority)

doc/dev/adr/0031-backpressure-bypass.md
vaavva: +1 (I think) - Although I'm not opposed to us not validating this as a reserved property, because customers should be able to set it on telemetry messages if they want. So for telemetry, either we shouldn't validate for this property, or we'd want to add SDK support for having us add this to telemetry messages (even though it wouldn't be on by default there). For RPC, it's more interesting because we wouldn't want the customer to specify this property and then we overwrite it after they've given it to us, so we probably have to validate against this property in RPC.

doc/dev/adr/0031-backpressure-bypass.md
```diff
@@ -0,0 +1,102 @@
...
+  cache behavior are all unaffected.
+
+### mRPC
+
+- **Invoker: default ON, customer-tunable OFF.** A single boolean on
+  the invoker options surface, set once at construction, defaulting to
+  `true`. Customers turn it off when they need their mRPC traffic to
+  follow normal-priority backpressure (e.g., for fairness testing or
+  when the broker's authz policy denies them the flag and they want to
+  avoid the rejected-PUBLISH path). No per-invocation toggle.
```
vaavva: Will the broker reject the publish if the flag can't be used, or will it just treat the publish as if the flag weren't there (I would hope for the latter, otherwise the on by default pattern gets a bit more annoying)

doc/dev/adr/0031-backpressure-bypass.md
```diff
@@ -0,0 +1,102 @@
...
+- **Executor mirrors, no option.** The executor copies the bypass user
+  property from the incoming request onto the response. There is no
+  corresponding executor option. This matches the intent of the MQ
```
vaavva: I've added other comments about this already, but is this intended in the MQ design? I could definitely see advantages to having as much as possible being set to high priority even if both request and response can't be. Especially since the response would often be sent by the internal service, which would be more likely to be allowed to set this flag

doc/dev/adr/0031-backpressure-bypass.md
```diff
@@ -0,0 +1,102 @@
...
+  ADR's chosen design and keeps response priority bound to request
+  priority.
+- **SDK-shipped service clients inherit the default.** State Store,
+  Lease Lock, Schema Registry, ADR, and the connector framework use
+  the same default-ON invoker; they re-expose the toggle on their
+  public options so customers can turn it off there too.
```
vaavva: should our services expose turning this off? I could see us being opinionated that any State Store/SR/etc operation should always fall into the high priority category (that said, I can't make a great argument that this could never be the case, although it does make the API surface for Rust a bit more annoying)

doc/dev/adr/0031-backpressure-bypass.md
vaavva: made it to the end, and this aligns with what's in your doc, but maybe worth re-wording/removing this up here to avoid confusion

doc/dev/adr/0031-backpressure-bypass.md
```diff
@@ -0,0 +1,102 @@
...
+> **Status:** Proposed.
+
+## Context
+
+The MQ broker [is adding][mq-adr] a high-priority backpressure-bypass
+mechanism so control-plane traffic (mRPC, State Store) is not starved when
+data-plane traffic fills the broker's buffer pool. The mark is an MQTT 5
+**user property on each PUBLISH** (broker owns the exact name, e.g.
+`$high_priority`); the broker also gets a CRD kill switch and an authz
+policy gating who may set the flag.
```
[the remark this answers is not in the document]
maximsemenov80: @copilot apply changes based on [this feedback](https://github.com/Azure/iot-operations-sdks/pull/1332#discussion_r3230345780) - defining the abbreviations on first use.

doc/dev/adr/0031-backpressure-bypass.md
maximsemenov80: it is in their ADR, I'll chat with them about it.

doc/dev/adr/0031-backpressure-bypass.md
maximsemenov80: This detail is from MQ's ADR, this what MQ is expecting - to have flag set on each PUBLISH (setting granularity of control for each message instead of whole session or connection)

doc/dev/adr/0031-backpressure-bypass.md
maximsemenov80: Intent of mirroring comes from their rejection of the alternative approach of using topic filters (all pubs to a topic matching certain filter treated as high priority) that would not allow to predict response topic hence mark response as high priority. But I agree giving the ability to set a response priority on its own make sense.

doc/dev/adr/0031-backpressure-bypass.md
vaavva: I think we had aligned to per invocation/response rather than on the entire invoker/executor. It's not necessarily my API surface preference, but I agree that it leaves us the most flexible in the future to avoid having to change this again if the needs change.

doc/dev/adr/0031-backpressure-bypass.md
maximsemenov80: check with MQ

doc/dev/adr/0031-backpressure-bypass.md
maximsemenov80: decided No for now

doc/dev/adr/0031-backpressure-bypass.md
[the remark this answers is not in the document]
maximsemenov80: @copilot apply changes based on [this feedback](https://github.com/Azure/iot-operations-sdks/pull/1332#discussion_r3230345780) - defining the abbreviations on first use.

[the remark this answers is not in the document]
maximsemenov80: @copilot apply changes based on [this feedback](https://github.com/Azure/iot-operations-sdks/pull/1332#discussion_r3230345780) - defining the abbreviations on first use.

timtay-microsoft (approved):

doc/dev/adr/0031-backpressure-bypass.md
```diff
@@ -0,0 +1,45 @@
...
+- **Invoker and Executor set the flag independently.**
+  Each side decides per its own logic whether to attach `$high_priority`
+  to the PUBLISH it produces. Request and response priorities are decoupled.
+- **Granularity is per-PUBLISH.**
+  Default behavior is to mark all PUBLISH with the flag, but this can be overridden per PUBLISH.
+
+### Codegen
+
+- No DTDL annotation. Bypass is a property of the caller, not the
+  contract.
```
vaavva:
```suggestion
  contract. Generated code must surface the underlying option to override the default behavior, aligning with the invoker/executor surface.
```

doc/dev/adr/0031-backpressure-bypass.md
```diff
@@ -0,0 +1,45 @@
...
+## Decision
+
+### mRPC
+
+- **Invoker and Executor set the flag independently.**
+  Each side decides per its own logic whether to attach `$high_priority`
+  to the PUBLISH it produces. Request and response priorities are decoupled.
+- **Granularity is per-PUBLISH.**
+  Default behavior is to mark all PUBLISH with the flag, but this can be overridden per PUBLISH.
+
```
vaavva, line 48: I think this note that you had is still good to call out here
```suggestion
The user property `$high_priority` is broker-owned and sits outside
  the SDK-reserved `__` prefix from
  [ADR 4](./0004-reserved-user-properties.md). SDKs must not validate
  against or reject `$`-prefixed user properties. If a user sets `$high_priority` themselves, since the contents don't matter to the broker, there isn't a concern about unexpected behavior.
```
Should we return an error though if the user sets `$high_priority`, but also specifies in the options to override the default of including this? Or are we saying that if they use this property, they should understand the ramifications even if they're explicitly setting high priority on and off?

vaavva, line 48: I think this is still good to call out as well

```suggestion
**SDK-shipped service clients use the default.** State Store,
  Lease Lock, Schema Registry, ADR, and the connector framework use
  default-ON invokers. At this time, the option to override this will not be exposed.
```

doc/dev/adr/0031-backpressure-bypass.md
```diff
@@ -0,0 +1,45 @@
...
+## Decision
+
+### mRPC
+
+- **Invoker and Executor set the flag independently.**
+  Each side decides per its own logic whether to attach `$high_priority`
+  to the PUBLISH it produces. Request and response priorities are decoupled.
+- **Granularity is per-PUBLISH.**
+  Default behavior is to mark all PUBLISH with the flag, but this can be overridden per PUBLISH.
```
vaavva: To align the SDKs, any naming we want to align on for the option? We could do high_priority: bool, but I feel like backpressure_bypass: bool might be easier to understand what's actually being set?

vaavva (approved):

doc/dev/adr/0031-backpressure-bypass.md
```diff
@@ -0,0 +1,48 @@
...
+policy gating who may set the flag. No other MQTT semantics change:
+QoS, expiry, topic, correlation, and cache behavior are all unaffected.
+
+## Decision
+
+### mRPC
+
+- **Invoker and Executor set the flag independently.**
+  Each side decides per its own logic whether to attach `$high_priority`
+  to the PUBLISH it produces. Request and response priorities are decoupled.
```
cartertinney: I think that in implementation this should be abstracted so as not to leak MQTT concepts through our higher-level protocol

doc/dev/adr/0031-backpressure-bypass.md
cartertinney: I like `backpressure_bypass` personally

doc/dev/adr/0031-backpressure-bypass.md
```diff
@@ -0,0 +1,48 @@
...
+  [ADR 4](./0004-reserved-user-properties.md). SDKs must not validate
+  against or reject `$`-prefixed user properties. If a user sets `$high_priority` themselves, since the contents don't matter to the broker, there isn't a concern about unexpected behavior.
+### Codegen
+
+- No DTDL annotation. Bypass is a property of the caller, not the
+  contract. Generated code must surface the underlying option to override the default behavior, aligning with the invoker/executor surface.
+
+### Compatibility
+
+- No protocol-version bump. Brokers that don't recognize the property
```
cartertinney, line 58: Probably worth clarifying that this is not truly part of the AIO protocol, as it's using MQTT protocol to achieve this functionality

cartertinney (approved):

doc/dev/adr/0031-backpressure-bypass.md
varunpuranik: Do we need to expose it on the SDK? Why not add it on all mRPC calls by default?

doc/dev/adr/0031-backpressure-bypass.md
varunpuranik: @arsing / @huguesBouvier - any comments?

arsing: We had discussed this already at the time and had agreement to use `$high_priority`

doc/dev/adr/0031-backpressure-bypass.md
maximsemenov80: discussed offline - agreed not to expose the option.

cartertinney (approved):

varunpuranik (approved):

PR #1335, opened 2026-05-18
Change prefixes dtv: and aov: to dov: and provide legacy support

jrdouceur:


wot-codegen/src/Dtdl2Wot/Command/t4/CommandAffordance.cs
```diff
@@ -143,13 +143,13 @@ public virtual string TransformText()
             this.Write("\"\r\n            }\r\n          ],\r\n");
  } 
  if (this.infoSchemaName != null) { 
-            this.Write("          \"dtv:headerInfo\": [\r\n            {\r\n              \"contentType\": \"appli" +
+            this.Write("          \"dov:headerInfo\": [\r\n            {\r\n              \"contentType\": \"appli" +
```
vaavva, line 146: Will this change require re-converting our service WoT contracts? And then their codegen output as well?

vipeller (approved):

wot-codegen/src/Dtdl2Wot/Command/t4/CommandAffordance.cs
jrdouceur: The prefixes affect only the models, not the on-wire communication format, and the legacy support in this PR should (modulo bugs) ensure that the extant service models will continue to be accepted by the ProtocolCompiler and will produce the same generated code.

vaavva: For any new service WoT contracts, should we use the new prefix going forward?

vaavva:
any readmes or the schemas in the wot-codegen/schema folder that need to be updated with this change?

vaavva (approved):
approved if nothing else needed for other open comment

wot-codegen/src/Dtdl2Wot/Command/t4/CommandAffordance.cs
jrdouceur: Yes, that would be recommended.  Especially for anything that is publicly visible, we want to use the new prefix.

jrdouceur:
> any readmes or the schemas in the wot-codegen/schema folder that need to be updated with this change?

Yes, but (a) the schemas are slightly out-of-date in another way as well, (b) there will be a bunch of additional properties getting added soon, and (c) this PR is already very large due to changes in converted models.  So, my plan had been to update the schemas in another PR.

PR #1337, opened 2026-05-19
[EdgeRegistry] Add Edge Registry WoTs

vaavva:


eng/wot/edge-registry/schema-extension/CreateSchemaAttributes.schema.json
```diff
@@ -0,0 +1,18 @@
...
+  "description": "Request payload for creating a schema with an initial version.",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [ "schemaMetaAttributes", "createSchemaVersionAttributes" ],
+  "properties": {
+    "schemaMetaAttributes": {
+      "description": "Initial meta attributes for the schema.",
+      "$ref": "../core-xregistry/ResourceMetaAttributes.schema.json"
+    },
+    "createSchemaVersionAttributes": {
```
varunpuranik: What is the difference between this and CreateScheamVersionAttributes?

vaavva: `CreateSchemaAttributes` has the `CreateSchemaVersionAttributes` as a component. It also has any attributes for the resource (schema) level meta entity.

eng/wot/edge-registry/ThingDescriptionExtensions.TM.json
```diff
@@ -0,0 +1,180 @@
...
+            "type": "string",
+            "const": "thingdescriptiongroups",
+            "description": "Group type used in xId for xRegistry Thing Description groups."
+        },
+        "ThingDescriptionResourceType": {
+            "type": "string",
+            "const": "thingdescriptions",
+            "description": "Resource type used in xId for xRegistry Thing Description resources."
+        },
+        "ThingDescriptionExtensionError": {
```
RLeclair, line 263: This, `SchemaExtensionError` and  `EdgeRegistryError` are all the same field-wise

Do you see them changing at any point, could they all be the same error?

RLeclair (approved):

eng/wot/edge-registry/ThingDescriptionExtensions.TM.json
vaavva: yeah, I wish I could share them - it's a limitation of WoT that I can't have the return error type be a schema reference, and you can't share WoT schemas across WoTs

eng/wot/edge-registry/SchemaExtensions.TM.json
```diff
@@ -0,0 +1,194 @@
...
+                    }
+                },
+                "required": [ "schemaId"]
+            },
+            "output": {
+                "dtv:ref": "./schema-extension/SchemaVersion.schema.json"
+            },
+            "forms": [
+                {
+                    "dtv:topic": "adr/dtmi:ms:adr:EdgeRegistry;1/get/schema/version",
```
varunpuranik: This is the MQTT Topic, right? why do we have this prefix? 
adr / dtmi?? Why not something like aio/registry/schemaGroups/{groupId}/schemas/{schemaId}/versions/{versionId} in this? 

Note that we need the group Id, schema Id and version Id in the topic so that we can do authz in the MQTT Broker

varunpuranik: Same for other resources

varunpuranik (commented):
As mentioned below, we need to update topics to include the resource being accessed, so that we can do authz in the MQ broker

eng/wot/edge-registry/SchemaExtensions.TM.json
vaavva: updated to `aio/registry/` namespace and included all possible Ids/types in the topics to give us authz flexibility. Couldn't always include groupId or versionId to allow them to be missing so that we can use the default

RLeclair (approved):

PR #1343, opened 2026-06-03
Add ADR 31 WoT Thing Model New Properties

jrdouceur:


doc/dev/adr/0031-wot-new-properties.md
```diff
@@ -0,0 +1,355 @@
...
+| [ADR 29](./0029-wot-platform-binding.md): RDF term prefix `aov:` for the AIO Platform Binding | Superseded by `dov:` for new terms; `aov:` retained as legacy for the previously defined Platform Binding terms enumerated above. |
+| [ADR 29](./0029-wot-platform-binding.md): `aov:memberOf` applies only to action affordances | Extended: `dov:memberOf` (and legacy `aov:memberOf`) applies to action, property, and event affordances; the referenced group must appear in the corresponding `dov:actionGroups`, `dov:propertyGroups`, or `dov:eventGroups` array. |
+| [ADR 29](./0029-wot-platform-binding.md): `aov:withUnit` value of the form `properties/<Name>` | The preferred form is now a 2-level JSON Pointer (`/properties/<Name>`); the legacy form without a leading `/` is still accepted. |
+
+## Consequences
+
+### Positive
+
+* A single Digital Operations Vocabulary prefix (`dov:`) simplifies authoring and reduces the number of `@context` declarations required in a typical Thing Model.
+* Named affordance groups, with cross-array uniqueness of titles, enable downstream tooling to organize affordances into logical groupings without ambiguity.
```
vaavva, line 346: are these groups expected to be used for SDK codegen scenarios, or is this more for other uses of WoTs? If the former, does this change how generated envoys are grouped together?

jrdouceur: At present, the groups are used on the authoring side, and they have no effect on generated code.  But it could perhaps eventually have an effect such as what you're alluding to.

vaavva: Sounds good - if that isn't the plan right now, then it probably makes sense to detail the ramifications for generation at the point where it becomes relevant

vaavva (approved):
I don't have much experience/context on the authoring side of things, but as this isn't affecting SDK codegen at this point, I don't have any concerns from that perspective

RLeclair (approved):

marcschier (approved):

varunpuranik (approved):

PR #1345, opened 2026-06-04
[Edge Registry] Add TM, per type versions, missing creates, deletes, lists

RLeclair:


eng/wot/edge-registry/core-xregistry/CreateResourceRequest.schema.json
```diff
@@ -0,0 +1,34 @@
...
+  "description": "Request payload for creating a generic xRegistry Resource entity.",
+  "type": "object",
+  "additionalProperties": false,
+  "required": ["labels"],
+  "properties": {
+    "groupId": {
+      "description": "Group identifier. Uses the default if not specified.",
+      "type": "string"
+    },
+    "name": {
```
vaavva: surprisingly, Resource doesn't have a `name` field

eng/wot/edge-registry/core-xregistry/CreateResourceRequest.schema.json
```diff
@@ -0,0 +1,34 @@
...
+      "description": "Human-readable name.",
+      "type": "string"
+    },
+    "description": {
```
vaavva: or description/documentation/labels. The ResourceMeta does have labels though

eng/wot/edge-registry/core-xregistry/CreateResourceRequest.schema.json
```diff
@@ -0,0 +1,34 @@
+{
+  "$schema": "https://json-schema.org/draft-07/schema",
+  "title": "CreateResourceRequest",
```
vaavva, line 3: For consistency with Group Attributes and Resource Meta Attributes, I'd probably call this ResourceAttributes - that way if we ever support Update, we could use the same schema

vaavva: Speaking of though, the createResource API should also take in ResourceMetaAttributes

eng/wot/edge-registry/EdgeRegistry.TM.json
```diff
@@ -91,6 +115,30 @@
                 }
             ]
         },
+        "createVersion": {
+            "description": "Create a new xRegistry Version entity under the specified Resource. The parent Resource is implicitly created if it doesn't already exist. Create is strictly idempotent: if every field matches the latest Version of the Resource (including labels), the latest Version is returned unchanged; otherwise a new Version is created.",
```
vaavva: For Core xRegistry, this will need to be slightly different - either it would be something like `Create is strictly idempotent: if the specified Version Id already exists and every field matches (including labels), the existing Version is returned unchanged; otherwise an error is returned. If the Version Id doesn't exist yet, it is created` if versionId is required and in the topic, or if versionId is optional (which we'd probably want to do to support scenarios like the schema extension where we can automatically determine the version id in addition to scenarios where the versionId needs to be specified), then we'd need to explain both of the behavior variants

eng/wot/edge-registry/core-xregistry/CreateVersionRequest.schema.json
```diff
@@ -0,0 +1,59 @@
+{
+  "$schema": "https://json-schema.org/draft-07/schema",
+  "title": "CreateVersionRequest",
+  "description": "Request payload for creating a generic xRegistry Version entity. Create is strictly idempotent: if every field matches the latest Version of the parent Resource (including labels), the latest Version is returned; otherwise a new Version is created.",
+  "type": "object",
+  "additionalProperties": false,
+  "required": ["labels", "document", "resourceLabels"],
```
vaavva:
```suggestion
  "required": ["labels", "resourceLabels"],
```
document won't always be required for core xReg

vaavva: but you'll want to add `extensions`

eng/wot/edge-registry/EdgeRegistry.TM.json
```diff
@@ -310,6 +358,133 @@
...
+                            "description": "Version identifier.",
+                            "type": "string"
+                        }
+                    }
+                },
+                "required": ["ids"]
+            },
+            "forms": [
+                {
+                    "dtv:topic": "aio/registry/listWithLabel/{ex:groupType}/{ex:resourceType}/{ex:resourceId}/versions",
```
vaavva: One thing here, to match `listResourcesWithLabel`, we'll probably want all of these fields to be optional to allow for less restrictive querying as well

eng/wot/edge-registry/EdgeRegistry.TM.json
```diff
@@ -310,6 +358,133 @@
...
+                    "additionalResponses": [
+                        {
+                            "success": false,
+                            "schema": "EdgeRegistryError"
+                        }
+                    ]
+                }
+            ]
+        },
+        "deleteGroup": {
```
vaavva: small organization nit - my preference would be to group APIs by entity, so all the Group APIs at the top, then Resource APIs, then Version APIs

eng/wot/edge-registry/EdgeRegistry.TM.json
```diff
@@ -310,6 +358,133 @@
...
+                        {
+                            "success": false,
+                            "schema": "EdgeRegistryError"
+                        }
+                    ]
+                }
+            ]
+        },
+        "deleteVersion": {
+            "description": "Delete an xRegistry Version entity. [ASK: behavior when deleting the default Version — promote next-most-recent, or error?]",
```
vaavva: yep - next most recent will become the new default for our extensions. I wouldn't necessarily specify that here though, because in a complete xRegistry implementation, there are a few possibilities for how this could get handled

eng/wot/edge-registry/ThingModelExtensions.TM.json
```diff
@@ -0,0 +1,268 @@
...
+            "const": {
+                "WotTM11": "WoT-TM/1.1"
+            },
+            "description": "Non-exhaustive supported WoT TM formats",
+            "properties": {
+                "WotTM11": {
+                    "description": "WoT TM 1.1 format",
+                    "type": "string"
+                }
+            }
```
vaavva, line 219: small update - this is the better format for us to use
```suggestion
            "const": {
                "JsonLD11": "JSON-LD/1.1"
            },
            "description": "Non-exhaustive supported WoT TM formats",
            "properties": {
                "JsonLD11": {
                    "description": "JSON-LD 1.1 format",
                    "type": "string"
                }
            }
```

eng/wot/edge-registry/core-xregistry/CreateVersionRequest.schema.json
```diff
@@ -0,0 +1,59 @@
...
+    "contentType": {
+      "description": "Content type of the Version document.",
+      "type": "string"
+    },
+    "format": {
+      "description": "Format identifier of the Version document (resource-type-specific, e.g. `JsonSchema/draft-07`, `JSON-LD/1.1`, `WoT-TM/1.1`).",
+      "type": "string"
+    },
+    "document": {
+      "description": "Base64-encoded document content for the Version. The interpretation (schema, thing description, thing model, etc.) is determined by the Resource type in the topic.",
```
vaavva:
```suggestion
      "description": "Base64-encoded document content for the Version. The interpretation (schema, thing description, thing model, etc.) is determined by the Resource type.",
```
the fact that it's specified in the topic will probably be abstracted to the end user

eng/wot/edge-registry/core-xregistry/CreateVersionRequest.schema.json
```diff
@@ -0,0 +1,59 @@
...
+    },
+    "contentType": {
+      "description": "Content type of the Version document.",
+      "type": "string"
+    },
+    "format": {
+      "description": "Format identifier of the Version document (resource-type-specific, e.g. `JsonSchema/draft-07`, `JSON-LD/1.1`, `WoT-TM/1.1`).",
+      "type": "string"
+    },
+    "document": {
```
vaavva: in xRegistry internally, we'll allow specifying a raw document or a url to a document, but I think only allowing a raw document in the SDK makes sense for now

eng/wot/edge-registry/core-xregistry/Version.schema.json
```diff
@@ -4,7 +4,7 @@
   "description": "A specific Version of a Resource.",
   "type": "object",
   "additionalProperties": false,
-  "required": [ "resourceId", "versionId", "xid", "epoch", "createdAt", "modifiedAt", "ancestor", "extensions", "labels" ],
+  "required": [ "resourceId", "versionId", "xid", "epoch", "createdAt", "modifiedAt", "ancestor", "extensions", "labels", "document", "documentHash" ],
```
vaavva:
```suggestion
  "required": [ "resourceId", "versionId", "xid", "epoch", "createdAt", "modifiedAt", "ancestor", "extensions", "labels" ],
```

eng/wot/edge-registry/schema-extension/CreateSchemaVersionAttributes.schema.json
```diff
@@ -42,18 +46,18 @@
...
-        },
         "schemaLabels": {
             "description": "Queryable Key Value pairs to be added to the parent Schema. Labels can be used to store additional metadata about the Schema.",
             "type": "array",
             "items": {
                 "type": "object",
                 "$ref": "../core-xregistry/Label.schema.json"
             }
+        },
+        "schemaDocument": {
```
vaavva: since this is now a core xReg field, I'd align the naming
```suggestion
        "document": {
```

eng/wot/edge-registry/schema-extension/SchemaVersion.schema.json
```diff
@@ -1,18 +1,97 @@
...
+    "resourceId",
+    "versionId",
+    "xid",
+    "epoch",
+    "createdAt",
+    "modifiedAt",
+    "ancestor",
+    "extensions",
+    "labels",
+    "schemaDocument",
```
vaavva: add format as required for the extensions as well
```suggestion
    "schemaDocument",
    "format",
```

eng/wot/edge-registry/schema-extension/CreateSchemaVersionAttributes.schema.json
```diff
@@ -4,7 +4,11 @@
     "description": "Request payload for creating a Schema Version.",
     "type": "object",
     "additionalProperties": false,
-    "required": [ "labels", "schemaDocument", "schemaLabels"],
+    "required": [
+        "labels",
+        "schemaDocument",
+        "schemaLabels"
```
vaavva, line 12: format here too

eng/wot/edge-registry/core-xregistry/CreateResourceRequest.schema.json
RLeclair: Removed `CreateResourceRequest` so this goes away

eng/wot/edge-registry/core-xregistry/CreateResourceRequest.schema.json
RLeclair: Removed with removal of `CreateResourceRequest`

eng/wot/edge-registry/core-xregistry/CreateResourceRequest.schema.json
RLeclair: Merged this into `ResourceMetaAttributes` and removed this file

eng/wot/edge-registry/core-xregistry/CreateVersionRequest.schema.json
RLeclair: Removed requirement

RLeclair: Added!

eng/wot/edge-registry/core-xregistry/CreateVersionRequest.schema.json
RLeclair: Stayed at only base64

eng/wot/edge-registry/EdgeRegistry.TM.json
RLeclair: Made versionID optional and rewrote the description to cover both variants. 

If version id, we do exact-match-or-error
If omitted version ID, server assigns one and we apply idempotence against the latest version.

I also moved version ID from being specified in the topic to being part of the `CreateVersionRequest.schema.json`

eng/wot/edge-registry/EdgeRegistry.TM.json
RLeclair: Good call, I've changed it to be the same as ListResourcesWithLabel

eng/wot/edge-registry/EdgeRegistry.TM.json
RLeclair: Done!

eng/wot/edge-registry/EdgeRegistry.TM.json
RLeclair: Got it, removed the ask!

eng/wot/edge-registry/core-xregistry/Version.schema.json
RLeclair: Removed document and document hash and added isDefault as required

Also made it (Version) match with the one from your ADO PR

eng/wot/edge-registry/schema-extension/CreateSchemaVersionAttributes.schema.json
RLeclair: Added format

eng/wot/edge-registry/schema-extension/CreateSchemaVersionAttributes.schema.json
RLeclair: Switched to `document`

eng/wot/edge-registry/schema-extension/SchemaVersion.schema.json
RLeclair: Added!

eng/wot/edge-registry/core-xregistry/Validated.schema.json
```diff
@@ -0,0 +1,18 @@
...
+  "type": "object",
+  "additionalProperties": false,
+  "required": [ "validated" ],
+  "properties": {
+    "validated": {
+      "description": "True if validation was performed and the entity adheres to the rules; false if validation was not performed.",
+      "type": "boolean"
+    },
+    "reason": {
+      "description": "If validation was not performed, the reason why not.",
```
vaavva:
```suggestion
      "description": "If validation was not performed, the reason why not. MUST be present if validated is false. MUST NOT be present if validated is true.",
```

eng/wot/edge-registry/schema-extension/CreateSchemaVersionAttributes.schema.json
```diff
@@ -4,10 +4,14 @@
     "description": "Request payload for creating a Schema Version.",
     "type": "object",
     "additionalProperties": false,
-    "required": [ "labels", "schemaDocument", "schemaLabels"],
+    "required": [
+        "labels",
+        "format",
+        "document"
```
vaavva: otherwise we end up with Option<HashMap>, which just makes things confusing
```suggestion
        "document",
        "extensions"
```

eng/wot/edge-registry/thing-description-extension/CreateThingDescriptionVersionAttributes.schema.json
```diff
@@ -4,10 +4,14 @@
     "description": "Request payload for creating a Thing Description Version.",
     "type": "object",
     "additionalProperties": false,
-    "required": ["labels", "thingDescriptionDocument", "thingDescriptionLabels"],
+    "required": [
+        "labels",
+        "format",
+        "document"
```
vaavva:
```suggestion
        "document",
        "extensions"
```

eng/wot/edge-registry/schema-extension/CreateSchemaVersionAttributes.schema.json
```diff
@@ -35,24 +43,24 @@
...
             "type": "string"
         },
-        "schemaDocument": {
-            "type": "string",
+        "document": {
             "description": "Base64-encoded schema document content.",
+            "type": "string",
             "contentEncoding": "base64"
         },
-        "schemaLabels": {
```
vaavva, line 50: these we still want (here and for the other extensions) - it's kind of a weird SDK surface thing we want to support, but we want to be able to add labels to the overall schema in addition to the specific version for our querying needs.

eng/wot/edge-registry/EdgeRegistry.TM.json
```diff
@@ -270,6 +290,86 @@
...
+                },
+                "required": [ ]
+            },
+            "output": {
+                "description": "The Version entity.",
+                "dtv:ref": "./core-xregistry/Version.schema.json"
+            },
+            "forms": [
+                {
+                    "dtv:topic": "aio/registry/get/{ex:groupType}/{ex:resourceType}/{ex:resourceId}/versions/{ex:versionId}",
```
vaavva: I think versionId should be optional here to return the default version if an Id isn't specified

eng/wot/edge-registry/EdgeRegistry.TM.json
```diff
@@ -91,25 +92,45 @@
...
+                            "schema": "EdgeRegistryError"
+                        }
+                    ]
+                }
+            ]
+        },
+        "createResource": {
+            "description": "Create a new xRegistry Resource entity. Returns the created Resource with epoch=1.",
+            "input": {
+                "dtv:ref": "./core-xregistry/ResourceMetaAttributes.schema.json",
```
vaavva: are there any other fields needed to create a resource other than those on the resource meta?

eng/wot/edge-registry/EdgeRegistry.TM.json
```diff
@@ -310,6 +410,117 @@
...
+            "description": "List xRegistry Versions that have the specified label.",
+            "input": {
+                "type": "object",
+                "properties": {
+                    "groupType": {
+                        "description": "Group type. Lists Versions across all Group types if not specified.",
+                        "type": "string"
+                    },
+                    "groupId": {
+                        "description": "Group identifier. Lists Versions across all Groups if not specified.",
```
vaavva: ugh, we probably need some way to distinguish between listing across all Groups and being able to list for the default group

eng/wot/edge-registry/schema-extension/CreateSchemaVersionAttributes.schema.json
RLeclair: Added extensions

eng/wot/edge-registry/core-xregistry/Validated.schema.json
RLeclair: Added!

eng/wot/edge-registry/schema-extension/CreateSchemaVersionAttributes.schema.json
RLeclair: Re-added here and in the other extensions! Good catch

eng/wot/edge-registry/thing-description-extension/CreateThingDescriptionVersionAttributes.schema.json
RLeclair: Added extensions again

eng/wot/edge-registry/EdgeRegistry.TM.json
RLeclair: Done, reversed it to be optional

eng/wot/edge-registry/EdgeRegistry.TM.json
RLeclair: I don't think so!

RLeclair: Updated per our conversation!

eng/wot/edge-registry/EdgeRegistry.TM.json
RLeclair: Added an allGroups boolean to both listResourcesWithLabel and listVersionsWithLabel. Omitting groupId now means the default group, a set groupId targets that specific group, and allGroups: true lists across all groups

eng/wot/edge-registry/core-xregistry/CreateResourceRequest.schema.json
```diff
@@ -0,0 +1,22 @@
...
+    "groupId": {
+      "description": "Group identifier. Uses the default if not specified.",
+      "type": "string"
+    },
+    "meta": {
+      "description": "The attributes of the Resource (its `meta` sub-entity).",
+      "$ref": "ResourceMetaAttributes.schema.json"
+    },
+    "defaultVersion": {
+      "description": "The attributes of the Resource's default Version.",
```
vaavva: something like this to indicate the extra creation happening
```suggestion
      "description": "The attributes of the Resource's default Version, which will be created along side this Resource.",
```

eng/wot/edge-registry/core-xregistry/CreateResourceRequest.schema.json
```diff
@@ -0,0 +1,22 @@
...
+  "type": "object",
+  "additionalProperties": false,
+  "required": [ "meta", "defaultVersion" ],
+  "properties": {
+    "groupId": {
+      "description": "Group identifier. Uses the default if not specified.",
+      "type": "string"
+    },
+    "meta": {
+      "description": "The attributes of the Resource (its `meta` sub-entity).",
```
vaavva:
```suggestion
      "description": "The attributes needed to create the Resource Meta (its `meta` sub-entity).",
```

eng/wot/edge-registry/core-xregistry/VersionAttributes.schema.json
```diff
@@ -0,0 +1,60 @@
+{
+  "$schema": "https://json-schema.org/draft-07/schema",
+  "title": "VersionAttributes",
+  "description": "Attributes needed to create a Version. Does not include addressing (groupId) or parent-Resource fields.",
```
vaavva: I'd just leave it at this
```suggestion
  "description": "Attributes needed to create a Version.",
```

vaavva (approved):

eng/wot/edge-registry/core-xregistry/CreateResourceRequest.schema.json
RLeclair: Addressed!

PR #1346, opened 2026-06-04
add CLI option --target to enable non-envoy generation

jrdouceur:
The actual functional change in this PR is enabling the ProtocolCompiler to generate schema definitions and schema code without generating envoy binder code.  But the CLI is structured to be broader than this narrow case.  The new `--target` CLI option allows the user to specify which SDK to target:

- a value of `aio` maintains the current behavior
- a value of `none` generates no envoy binder code

This design allows for future expansion to target other SDKs beyond the AIO SDK.  In particular, we may add an option to target a minimal envoy that will be hosted in a WASM container.

### DIFF
--- a/wot-codegen/src/Azure.Iot.Operations.CodeGeneration/TargetSdk.cs
+++ b/wot-codegen/src/Azure.Iot.Operations.CodeGeneration/TargetSdk.cs
@@ -1,0 +1,11 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License
+
+namespace Azure.Iot.Operations.CodeGeneration
+{
+    public enum TargetSdk
+    {
+        Aio,
+        None,
+    }
+}
--- a/wot-codegen/src/Azure.Iot.Operations.ProtocolCompiler/Program.cs
+++ b/wot-codegen/src/Azure.Iot.Operations.ProtocolCompiler/Program.cs
@@ -75,6 +75,13 @@ namespace Azure.Iot.Operations.ProtocolCompiler
                 HelpName = "NAMESPACE",
             };
 
+            var targetOption = new Option<string>("--target")
+            {
+                DefaultValueFactory = (_) => "aio",
+                Description = $"SDK that will be targeted by generated code",
+                HelpName = string.Join('|', CommandHandler.SupportedSdkTargets),
+            };
+
             var sdkPathOption = new Option<string?>("--sdkPath")
             {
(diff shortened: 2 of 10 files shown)


wot-codegen/src/Azure.Iot.Operations.ProtocolCompiler/Program.cs
```diff
@@ -75,6 +75,13 @@ static void Main(string[] args)
                 HelpName = "NAMESPACE",
             };
 
+            var targetOption = new Option<string>("--target")
```
vaavva, line 78: any docs that should be updated with this additional CLI option?

jrdouceur: The only one I'm aware of is the [README](https://github.com/Azure/iot-operations-sdks/tree/main/wot-codegen#readme).  On one hand, it seems like this should be updated the same time as the code; however, this is the only doc for the published .NET tool, so on the other hand, we maybe shouldn't describe an option that's not in the published tool.

wot-codegen/src/Azure.Iot.Operations.CodeGeneration/TargetSdk.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License
+
+namespace Azure.Iot.Operations.CodeGeneration
+{
+    public enum TargetSdk
+    {
+        Aio,
```
vaavva, line 8: it seems a bit weird for something in the AIO SDK repo to generate code for a different SDK...I definitely get the benefit of generating just the structs and not all of the envoys, but this is starting to bring up the question of whether this repo is the best place for codegen things to live or not again to me

jrdouceur: I think there should probably be a significant refactoring of these various libraries and tools, especially as the ownership becomes distributed.  For now, though, I think the most important thing is to get the code merged so it can be used.  At least, that is the priority as I understand it,.

wot-codegen/src/Azure.Iot.Operations.ProtocolCompiler/Program.cs
vaavva: ah, that's unfortunate that that can't be updated yet so that it doesn't get forgotten later on

vaavva (approved):

PR #1352, opened 2026-06-10
[Services] Modify Edge Registry WoTs and add generated Rust code

RLeclair:
This PR:
- Adds Rust codegen for the Edge Registry service under a new `edge_registry` cargo feature. Generates the base xRegistry and extensions under a single `edge_registry` namespace via one `gen.sh` invocation.
- Deduplicates shared `VersionId`/`ResourceXId` types into `core-xregistry` schemas referenced via ref so they're generated once across all four models instead of being duplicated per model.
 - Renames the `createResourceRequest`/`createVersionRequest` input schemas to `CreateResourceRequestPayload`/`CreateVersionRequestPayload`  to avoid a title collision with the auto-generated `{Action}Request` RPC alias (which produced ambiguous_glob_reexports), following the existing `{actionName}RequestPayload` convention used by ADR
 - Corrects the stale dov @ context URI in the three extension TMs
 - **Adds an explicit output to the delete operations to work around a Rust-codegen limitation on actions that return an error but no success payload**
- One known title collision remains intentionally unaddressed: JSON_LD11 (shared between the TD and TM format enums). It's harmless under the private module and will be handled in the wrapper PR.


eng/wot/edge-registry/core-xregistry/CreateResourceRequestPayload.schema.json
```diff
@@ -0,0 +1,34 @@
+{
+  "$schema": "https://json-schema.org/draft-07/schema",
+  "title": "CreateResourceRequestPayload",
```
vaavva, line 3: Is there an old file that should be deleted here or have this show as a rename instead of a new file?

eng/wot/edge-registry/core-xregistry/ResourceXidList.schema.json
```diff
@@ -0,0 +1,19 @@
+{
+  "$schema": "https://json-schema.org/draft-07/schema",
+  "title": "ResourceXidList",
```
vaavva: make the xId casing consistent - I'd go with XId for camel case uses. `XID` in documentation though is good

eng/wot/edge-registry/SchemaExtensions.TM.json
```diff
@@ -149,51 +149,8 @@
...
-                                    "type": "string"
-                                }
-                            }
-                        }
-                    }
-                },
-                "required": [
-                    "versions"
-                ]
+                "dov:ref": "./core-xregistry/VersionXidList.schema.json",
```
vaavva: Unfortunately, the extension ones should have versionId as an int

rust/azure_iot_operations_services/src/edge_registry/edge_registry_gen/edge_registry/create_resource_request_payload.rs
```diff
@@ -0,0 +1,35 @@
...
+use serde::{Deserialize, Serialize};
+use uuid::Uuid;
+
+use super::super::common_types::{b64::Bytes, date_only::Date, decimal::Decimal, time_only::Time};
+use super::resource_meta_attributes::ResourceMetaAttributes;
+use super::version_attributes::VersionAttributes;
+
+/// Request payload for creating a generic xRegistry Resource entity along with its default Version.
+#[derive(Serialize, Deserialize, Debug, Clone, Builder)]
+pub struct CreateResourceRequestPayload {
```
vaavva, line 18: This seems like effectively a duplicate of create_resourse_request - I vaguely remember there's a codegen reason for this, but just wanted to sanity check that these should both be here

eng/wot/edge-registry/core-xregistry/CreateResourceRequestPayload.schema.json
RLeclair: Good catch, there were two and are now deleted

eng/wot/edge-registry/core-xregistry/ResourceXidList.schema.json
RLeclair: Done!

eng/wot/edge-registry/SchemaExtensions.TM.json
RLeclair: I added new list and versionXId schemas for each extension

rust/azure_iot_operations_services/src/edge_registry/edge_registry_gen/edge_registry/create_resource_request_payload.rs
RLeclair: Create resource request should no longer be there

The real one should be `CreateResourceRequestPayload` to prevent a collision between the generated action name (`CreateResource` -> `CreateResourceRequest`) and the schema we had before `CreateResourceRequest`

vaavva (approved):

PR #1355, opened 2026-06-11
[Rust][MQTT] fix: race condition in `PlenaryAck`

cartertinney:
It was possible for a concurrency bug where two `PlenaryAckMembers` there were acknowledged at the same time could lead to a TOCTOU race when resolving the "all members acknowledged" trigger.


rust/azure_iot_operations_mqtt/src/session/plenary_ack.rs
```diff
@@ -171,18 +171,16 @@ impl InnerState {
         if let Some(total) = sealed {
             // Check if all members have acked
             if self.counter.load(Ordering::SeqCst) == total {
-                // Check if result is not yet set
-                if self.result.get().is_none() && self.manual_ack.lock().unwrap().is_some() {
```
vaavva, line 175: why doesn't this check still need to happen?
> if self.result.get().is_none()

vaavva (approved):

rust/azure_iot_operations_mqtt/src/session/plenary_ack.rs
cartertinney: It's redundant. Always was actually - only the single caller that successfully gets the `.take()` can ever set the result.

I could add it back as defensive code if you'd prefer though.

vaavva: no preference, especially if redundant! Was just curious if this change affected this or not

vaavva (approved):

PR #1358, opened 2026-06-12
Hardcode disabling Nagle algorithm (TCP_NODELAY) in Rust MQTT transport

huguesBouvier:
Set TCP_NODELAY on all TCP connect paths (plain TCP, kTLS, userspace TLS, and WebSocket) to minimize latency.


cartertinney:
Re-implemented in #1366

PR #1359, opened 2026-06-12
[Services][Edge Registry] Unify list/listWithLabel actions and move version IDs to create payloads

RLeclair:
This PR:

- Merges each paired list…/list…WithLabel action into a single action, since group type is no longer required for the non-label variant - Generic: listResources, listVersions
- Extensions: listSchemaVersions, listThingDescriptionVersions, listThingModelVersions
- Introduces *RequestPayload input schemas where every locator (group/type/resource id) is optional, and label is modeled as a single optional Label object ($ref), enforcing labelKey/labelValue both-or-neither
- Merged actions now return XID lists (*XIdList) so results spanning multiple groups/types are fully identified; the bare-{ids} outputs are dropped
- Updates topics to collision-free names (aio/registry/listResources, aio/registry/listVersions, aio/registry/<ext>Extension/listVersions), avoiding the aio/registry/list/+ wildcard used by listGroups
- Clarifies locator semantics in field docs: resourceType matches across all Groups when listing across Groups; resourceId matches across all Resource types when resourceType is unspecified

Update:
- Relocates the version identifier from the entity attribute schemas to the create-request payloads


eng/wot/edge-registry/core-xregistry/ListResourcesRequestPayload.schema.json
```diff
@@ -0,0 +1,30 @@
...
+  "type": "object",
+  "additionalProperties": false,
+  "required": [],
+  "properties": {
+    "groupType": {
+      "description": "Group type. Lists Resources across all Group types if not specified.",
+      "type": "string"
+    },
+    "groupId": {
+      "description": "Group identifier. Uses the default Group if not specified. Ignored if allGroups is true.",
```
vaavva: Did we decide that the SDK and service will just validate that this isn't default when groupType is None?

eng/wot/edge-registry/schema-extension/ListSchemaVersionsRequestPayload.schema.json
```diff
@@ -0,0 +1,26 @@
+{
+  "$schema": "https://json-schema.org/draft-07/schema",
+  "title": "ListSchemaVersionsRequestPayload",
```
vaavva: Do these need to be additional schema.jsons? Or can the core one be used and just the return type be extension specific?

eng/wot/edge-registry/core-xregistry/ListResourcesRequestPayload.schema.json
RLeclair: Yes, a defaulted group ID will only be meaningful when a group type is specified

I added some comments to clarify!

eng/wot/edge-registry/schema-extension/ListSchemaVersionsRequestPayload.schema.json
RLeclair: If we use the core ones we would expose groupType and resourceType which we can hide with the SDK but I thought it better to have a separate request payload

vaavva: I think I'd lean to reducing duplicates and covering it with the SDK - we're already doing that for other APIs (slightly different since those are full apis instead of just some structs, but I think the goals are the same)

RLeclair: Makes sense, went ahead and reduced the duplicates

vaavva (approved):

PR #1361, opened 2026-06-15
[Services]  Adding generic xRegistry client support

RLeclair:
This PR adds the Edge Registry feature, models, and generic xRegistry client wrapper

Tests and example will come in a separate PR.

### DIFF
--- a/eng/wot/edge-registry/core-xregistry/ListResourcesRequestPayload.schema.json
+++ b/eng/wot/edge-registry/core-xregistry/ListResourcesRequestPayload.schema.json
@@ -17,7 +17,7 @@
       "type": "string"
     },
     "allGroups": {
-      "description": "If true, lists Resources across all Groups (groupId is ignored). If false or unspecified, lists within the Group specified by groupId, or — only when groupType is specified — the default Group of that type if groupId is unspecified.",
+      "description": "If true, lists Resources across all Groups (groupId is ignored). If false, lists within the Group specified by groupId, or — only when groupType is specified — the default Group of that type if groupId is unspecified.",
       "type": "boolean"
     },
     "resourceType": {
--- a/eng/wot/edge-registry/core-xregistry/ListResourcesRequestPayload.schema.json
+++ b/eng/wot/edge-registry/core-xregistry/ListResourcesRequestPayload.schema.json
@@ -4,7 +4,9 @@
   "description": "Request payload for listing xRegistry Resources, optionally filtered by Group type, Group, Resource type, and/or label.",
   "type": "object",
   "additionalProperties": false,
-  "required": [],
+  "required": [
+    "allGroups"
+  ],
   "properties": {
     "groupType": {
       "description": "Group type. If not specified, Resources are listed across all Group types; in that case groupId must be specified explicitly or allGroups must be true, because the default Group only exists within a specific Group type.",
--- a/rust/azure_iot_operations_services/Cargo.toml
+++ b/rust/azure_iot_operations_services/Cargo.toml
@@ -36,6 +36,7 @@ edge_registry = [
   "bigdecimal",
   "time",
   "uuid",
+  "bytes",
 ]
 azure_device_registry = [
   "serde",
@@ -52,6 +53,7 @@ azure_device_registry = [
 [dependencies]
 azure_iot_operations_protocol = { version = "1.0", path = "../azure_iot_operations_protocol" }
 azure_iot_operations_mqtt = { version = "1.0", path = "../azure_iot_operations_mqtt" }
+bytes = { workspace = true, optional = true }
 derive_builder.workspace = true
 log.workspace = true
 thiserror.workspace = true
--- a/rust/azure_iot_operations_services/src/edge_registry.rs
+++ b/rust/azure_iot_operations_services/src/edge_registry.rs
@@ -152,8 +152,11 @@ impl From<Label> for client_gen::Label {
     }
 }
 
-impl From<(String, String)> for Label {
-    fn from((key, value): (String, String)) -> Self {
-        Self { key, value }
+impl From<client_gen::Label> for Label {
+    fn from(value: client_gen::Label) -> Self {
+        Self {
+            key: value.key,
+            value: value.value,
+        }
     }
 }
--- a/rust/azure_iot_operations_services/src/edge_registry.rs
+++ b/rust/azure_iot_operations_services/src/edge_registry.rs
@@ -102,6 +102,25 @@ impl From<GetVersionId> for Option<String> {
     }
 }
 
+/// Identifies the Version identifier to assign when creating a Version.
+#[derive(Debug, Clone, Default)]
+pub enum CreateVersionId {
+    /// Let the server assign the Version identifier.
+    #[default]
+    ServerAssigned,
+    /// Create the Version with this specific Version identifier.
+    Specified(String),
+}
+
+impl From<CreateVersionId> for Option<String> {
+    fn from(value: CreateVersionId) -> Self {
+        match value {
+            CreateVersionId::ServerAssigned => None,
+            CreateVersionId::Specified(id) => Some(id),
+        }
+    }
+}
+
 /// Selects which Groups a label query spans.
 pub enum GroupQuery {
     /// Search across all Group types. There is no default Group without a fixed Group type, so only
--- a/rust/azure_iot_operations_services/src/edge_registry.rs
+++ b/rust/azure_iot_operations_services/src/edge_registry.rs
@@ -67,7 +67,7 @@ impl From<rpc_command::invoker::RequestBuilderError> for ErrorKind {
 /// Identifies a Group within its Group type for a request.
 #[derive(Debug, Clone, Default)]
 pub enum GroupId {
-    /// Use the cloud default Group of the Group type.
+    /// Use the cloud default Group Id of the Group type.
     #[default]
     CloudDefault,
     /// Use the Group with the specified identifier.
@@ -138,7 +138,7 @@ pub enum GroupSelection {
 #[derive(Debug, Clone)]
 pub struct Label {
     /// The label key.
-    pub name: String,
+    pub key: String,
     /// The label value.
     pub value: String,
 }
@@ -146,14 +146,14 @@ pub struct Label {
 impl From<Label> for client_gen::Label {
     fn from(value: Label) -> Self {
         client_gen::Label {
-            key: value.name,
+            key: value.key,
             value: value.value,
         }
     }
 }
 
 impl From<(String, String)> for Label {
-    fn from((name, value): (String, String)) -> Self {
-        Self { name, value }
+    fn from((key, value): (String, String)) -> Self {
+        Self { key, value }
     }
 }
--- a/rust/azure_iot_operations_services/src/edge_registry.rs
+++ b/rust/azure_iot_operations_services/src/edge_registry.rs
@@ -3,5 +3,157 @@
 
 //! Types for Edge Registry (xRegistry) operations.
 
+use azure_iot_operations_protocol::common::aio_protocol_error::AIOProtocolError;
+use azure_iot_operations_protocol::rpc_command;
+use thiserror::Error;
+
+use crate::edge_registry::edge_registry_gen::edge_registry::client as client_gen;
+
 /// Edge Registry generated code
 mod edge_registry_gen;
+
+/// Edge Registry Client implementation wrapper
+pub mod client;
+pub mod models;
+
+pub use client::Client;
+
+// ~~~~~~~~~~~~~~~~~~~SDK Created Structs~~~~~~~~~~~~~~~~~~~~~~~~
+
+/// Represents an error that occurred in the Azure IoT Operations Edge Registry Client implementation.
+#[derive(Debug, Error)]
+#[error(transparent)]
+pub struct Error(#[from] ErrorKind);
+
+impl Error {
+    /// Returns the [`ErrorKind`] of the error.
+    #[must_use]
+    pub fn kind(&self) -> &ErrorKind {
+        &self.0
+    }
+}
+
+/// Represents the kinds of errors that occur in the Azure IoT Operations Edge Registry implementation.
+#[derive(Debug, Error)]
+pub enum ErrorKind {
+    /// An error occurred in the AIO Protocol. See [`AIOProtocolError`] for more information.
+    #[error(transparent)]
+    AIOProtocolError(#[from] AIOProtocolError),
+    /// An error was returned by the Edge Registry service.
+    #[error("{0:?}")]
+    ServiceError(client_gen::EdgeRegistryError),
+    /// An argument provided for a request was invalid.
+    #[error("{0}")]
+    ValidationError(String),
+    /// An error occurred while shutting down the Edge Registry Client.
+    #[error("Shutdown error occurred with the following protocol errors: {0:?}")]
+    ShutdownError(Vec<AIOProtocolError>),
+}
+
+impl From<rpc_command::invoker::Response<client_gen::EdgeRegistryError>> for ErrorKind {
+    fn from(value: rpc_command::invoker::Response<client_gen::EdgeRegistryError>) -> Self {
+        Self::ServiceError(value.payload)
+    }
+}
+
+impl From<rpc_command::invoker::RequestBuilderError> for ErrorKind {
+    fn from(e: rpc_command::invoker::RequestBuilderError) -> Self {
+        ErrorKind::ValidationError(e.to_string())
+    }
+}
+
+// ~~~~~~~~~~~~~~~~~~~SDK Created Helper Structs~~~~~~~~~~~~~~~~~~~~~~~~
+
+/// Identifies a Group within its Group type for a request.
+#[derive(Debug, Clone, Default)]
+pub enum GroupId {
+    /// Use the cloud default Group of the Group type.
+    #[default]
+    CloudDefault,
+    /// Use the Group with the specified identifier.
+    Specified(String),
+}
+
+impl From<GroupId> for Option<String> {
+    fn from(value: GroupId) -> Self {
+        match value {
+            GroupId::CloudDefault => None,
+            GroupId::Specified(id) => Some(id),
+        }
+    }
+}
+
+/// Identifies which Version of a Resource to retrieve.
+#[derive(Debug, Clone, Default)]
+pub enum GetVersionId {
+    /// Retrieve the default Version of the Resource.
+    #[default]
+    ResourceDefault,
+    /// Retrieve the Version with the specified identifier.
+    Specified(String),
+}
+
+impl From<GetVersionId> for Option<String> {
+    fn from(value: GetVersionId) -> Self {
+        match value {
+            GetVersionId::ResourceDefault => None,
+            GetVersionId::Specified(id) => Some(id),
+        }
+    }
+}
+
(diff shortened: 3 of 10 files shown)


rust/azure_iot_operations_services/src/edge_registry/models/xregistry.rs
```diff
@@ -0,0 +1,540 @@
...
+use bytes::Bytes;
+use chrono::{DateTime, Utc};
+
+use crate::edge_registry::edge_registry_gen::common_types::b64::{self};
+use crate::edge_registry::edge_registry_gen::edge_registry::client as client_gen;
+
+// ~~~~~~~~~~~~~~~~~~~Conversion helpers~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+/// Converts a generated list of `Label` key/value pairs into a map.
+fn labels_from_gen(labels: Vec<client_gen::Label>) -> HashMap<String, String> {
```
vaavva: labels needs to be a Vec<(String, String)>, because it's valid to have multiple labels with the same key

rust/azure_iot_operations_services/src/edge_registry/models/xregistry.rs
```diff
@@ -0,0 +1,540 @@
...
+    /// The date/time of when the entity was created.
+    pub created_at: DateTime<Utc>,
+    /// The date/time of when the entity was last updated.
+    pub modified_at: DateTime<Utc>,
+    /// Information about deprecation status of the entity, if applicable.
+    pub deprecated: Option<DeprecatedInfo>,
+    /// Map of the count of each Resource type contained within this Group, keyed by Resource type.
+    pub resources_counts: HashMap<String, u64>,
+    /// Extension-specific attributes (e.g., `envelope`, `protocol` for message Groups).
+    pub extensions: HashMap<String, Bytes>,
```
vaavva, line 215: Do we want this to be `HashMap<String, Bytes>` or `HashMap<String, Vec<u8>>`?

rust/azure_iot_operations_services/src/edge_registry/models/xregistry.rs
```diff
@@ -0,0 +1,540 @@
...
+    /// Version conforms to the rules defined by its `format` attribute.
+    pub format_validated: Option<Validated>,
+    /// When compatibility validation is enabled, indicates whether the server has validated that
+    /// the Version conforms to the rules defined by its Resource's `meta.compatibility` attribute.
+    pub compatibility_validated: Option<Validated>,
+    /// The raw document content for this Version as base64-encoded bytes. The interpretation
+    /// (schema, thing description, thing model, …) is determined by the parent Resource's type.
+    pub document: Option<Bytes>,
+    /// The hash of the document content for this Version.
+    pub document_hash: Option<String>,
```
vaavva, line 368: Do we want to combine the optionality on these in the SDK? Maybe not since then we have another vector of failure from the wire contract

rust/azure_iot_operations_services/src/edge_registry/models/xregistry.rs
```diff
@@ -0,0 +1,540 @@
...
+    fn from(value: client_gen::VersionXidList) -> Self {
+        value.versions.into_iter().map(VersionXid::from).collect()
+    }
+}
+
+// ~~~~~~~~~~~~~~~~~~~Read entities~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+/// A Group entity — container for related Resources.
+#[derive(Debug, Clone)]
+pub struct Group {
```
vaavva: In the service, I added an `Entity` suffix to all of these items that are really 1:1 with the xRegistry structs to kind of clarify the difference between other structs that we're making up/other wrappers vs the main data pieces. Not sure at the SDK level that's as needed, but something to keep in mind if all of the structs for Groups/Resources/etc start getting a bit confusing

rust/azure_iot_operations_services/src/edge_registry.rs
```diff
@@ -3,5 +3,157 @@
...
+        ErrorKind::ValidationError(e.to_string())
+    }
+}
+
+// ~~~~~~~~~~~~~~~~~~~SDK Created Helper Structs~~~~~~~~~~~~~~~~~~~~~~~~
+
+/// Identifies a Group within its Group type for a request.
+#[derive(Debug, Clone, Default)]
+pub enum GroupId {
+    /// Use the cloud default Group of the Group type.
```
vaavva:
```suggestion
    /// Use the cloud default Group Id of the Group type.
```

rust/azure_iot_operations_services/src/edge_registry/models/xregistry.rs
```diff
@@ -0,0 +1,540 @@
...
+            format: value.format,
+            document: value.document.map(|b| b64::Bytes(b.to_vec())),
+            extensions: extensions_to_gen(value.extensions),
+        }
+    }
+}
+
+/// Request payload for creating a generic xRegistry Resource entity along with its default Version.
+#[derive(Debug, Clone)]
+pub struct CreateResourceRequest {
```
vaavva: is this used?

rust/azure_iot_operations_services/src/edge_registry/models/xregistry.rs
```diff
@@ -0,0 +1,540 @@
...
+            documentation: value.documentation,
+        }
+    }
+}
+
+// ~~~~~~~~~~~~~~~~~~~Identifiers (XIDs)~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+/// The XID components that identify a Resource.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ResourceXid {
```
vaavva: Would case these as `ResourceXId`

rust/azure_iot_operations_services/src/edge_registry/client.rs
```diff
@@ -0,0 +1,868 @@
...
+    /// [`struct@Error`] of kind [`AIOProtocolError`](ErrorKind::AIOProtocolError) if there are
+    /// any underlying errors from the AIO RPC protocol.
+    ///
+    /// [`struct@Error`] of kind [`ServiceError`](ErrorKind::ServiceError) if an error is returned
+    /// by the Edge Registry service.
+    pub async fn list_resources(
+        &self,
+        group_query: GroupQuery,
+        resource_type: Option<String>,
+        label: Option<Label>,
```
vaavva, line 445: Here you're using a type `Label`, but in the returned structs, you use String, String - probably should align to one or the other

rust/azure_iot_operations_services/src/edge_registry.rs
```diff
@@ -3,5 +3,157 @@
...
+        match value {
+            GroupId::CloudDefault => None,
+            GroupId::Specified(id) => Some(id),
+        }
+    }
+}
+
+/// Identifies which Version of a Resource to retrieve.
+#[derive(Debug, Clone, Default)]
+pub enum GetVersionId {
```
vaavva, line 88: thoughts on this to match GroupId?
```suggestion
pub enum VersionId {
```

rust/azure_iot_operations_services/src/edge_registry/client.rs
```diff
@@ -0,0 +1,868 @@
...
+        resource_id: Option<String>,
+        label: Option<Label>,
+        timeout: Duration,
+    ) -> Result<Vec<VersionXid>, Error> {
+        let (group_type, group_id, all_groups) = Self::group_query_scope(group_query);
+
+        let payload = client_gen::ListVersionsRequestPayload {
+            group_type,
+            group_id,
+            all_groups: Some(all_groups),
```
vaavva: If we're always passing this as Some, does it need to be optional on the WoT?

rust/azure_iot_operations_services/src/edge_registry.rs
RLeclair: Changed!

rust/azure_iot_operations_services/src/edge_registry.rs
RLeclair: I spoke about this with @cartertinney

We thought `GetVersionId` was better because the name `VersionId` implies that it can be used in other non-get APIs which it can't. `GroupId` on the other hand can

rust/azure_iot_operations_services/src/edge_registry/models/xregistry.rs
```diff
@@ -0,0 +1,540 @@
...
+        .into_iter()
+        .map(|(key, value)| client_gen::Label { key, value })
+        .collect()
+}
+
+/// Converts a generated map of base64 extension values into byte buffers.
+fn extensions_from_gen(extensions: HashMap<String, b64::Bytes>) -> HashMap<String, Bytes> {
+    extensions
+        .into_iter()
+        .map(|(k, v)| (k, Bytes::from(v.0)))
```
cartertinney, line 35: Is there a cost do doing these `bytes::Bytes`/`b64::Bytes` conversions?

rust/azure_iot_operations_services/src/edge_registry/client.rs
RLeclair: Definitely not, can change

rust/azure_iot_operations_services/src/edge_registry/models/xregistry.rs
RLeclair: Changed!

rust/azure_iot_operations_services/src/edge_registry/models/xregistry.rs
RLeclair: We use `Bytes` already in the SDK so I picked `Bytes` after discussing with Carter

rust/azure_iot_operations_services/src/edge_registry/models/xregistry.rs
RLeclair: Good catch, we are not using it. I'm not sure why clippy didn't flag this

rust/azure_iot_operations_services/src/edge_registry/models/xregistry.rs
RLeclair: Done!

rust/azure_iot_operations_services/src/edge_registry/models/xregistry.rs
RLeclair: How could the wire contract fail here? In case the document hash doesn't match the document?

rust/azure_iot_operations_services/src/edge_registry/models/xregistry.rs
RLeclair: Makes sense to have that there, I added the `Entity` suffix

rust/azure_iot_operations_services/src/edge_registry/models.rs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Models for Edge Registry operations.
+
+pub mod xregistry;
+
+pub use xregistry::*;
+
+// TODO: Add the extensions here as well.
```
cartertinney, line 10: There are going to be other submodules here as well I presume?

rust/azure_iot_operations_services/src/edge_registry.rs
```diff
@@ -3,5 +3,157 @@
...
+
+/// Represents an error that occurred in the Azure IoT Operations Edge Registry Client implementation.
+#[derive(Debug, Error)]
+#[error(transparent)]
+pub struct Error(#[from] ErrorKind);
+
+impl Error {
+    /// Returns the [`ErrorKind`] of the error.
+    #[must_use]
+    pub fn kind(&self) -> &ErrorKind {
```
cartertinney, line 31: Obligatory question about error design: does the kind matter? Or are we better off using some other kind of axis?

rust/azure_iot_operations_services/src/edge_registry/client.rs
RLeclair: Went with type `Label` to be more clear

rust/azure_iot_operations_services/src/edge_registry/models/xregistry.rs
RLeclair: Also removed another struct we were not using

rust/azure_iot_operations_services/src/edge_registry/models/xregistry.rs
vaavva: if the service returned a document without the hash (or vice-versa) - it wouldn't, but you'd then have a failure point to handle trying to turn it to a struct or something where both are required

eng/wot/edge-registry/core-xregistry/ListResourcesRequestPayload.schema.json
vaavva:
```suggestion
      "description": "If true, lists Resources across all Groups (groupId is ignored). If false, lists within the Group specified by groupId, or — only when groupType is specified — the default Group of that type if groupId is unspecified.",
```

vaavva: and same in the other file

rust/azure_iot_operations_services/src/edge_registry.rs
```diff
@@ -3,5 +3,157 @@
...
+impl From<Label> for client_gen::Label {
+    fn from(value: Label) -> Self {
+        client_gen::Label {
+            key: value.key,
+            value: value.value,
+        }
+    }
+}
+
+impl From<(String, String)> for Label {
```
vaavva: is this still used? Or maybe should be From<client_gen::Label> now

rust/azure_iot_operations_services/src/edge_registry/models/xregistry.rs
```diff
@@ -0,0 +1,513 @@
...
+        .into_iter()
+        .map(|l| (l.key, l.value).into())
+        .collect()
+}
+
+/// Converts a list of [`Label`] into the generated list of `Label` key/value pairs.
+pub(crate) fn labels_to_gen(labels: Vec<Label>) -> Vec<client_gen::Label> {
+    labels
+        .into_iter()
+        .map(|label| client_gen::Label {
```
vaavva: I think you can just do .into() here since you impl From<Label> for client_gen::Label

rust/azure_iot_operations_services/src/edge_registry/client.rs
```diff
@@ -0,0 +1,868 @@
...
+
+    /// Create a new xRegistry Group.
+    ///
+    /// # Arguments
+    /// * `group_type` - The type of the Group to create.
+    /// * `group_id` - The identifier of the Group to create. If [`CloudDefault`](GroupId::CloudDefault), create the default Group of the Group type.
+    /// * `attributes` - The [`GroupAttributes`] for the new Group.
+    /// * `timeout` - The duration until the client stops waiting for a response to the request, it is rounded up to the nearest second.
+    ///
+    /// Returns the created [`Group`] with epoch 1.
```
vaavva: doc updates needed for the Entity renames

rust/azure_iot_operations_services/Cargo.toml
```diff
@@ -52,6 +53,7 @@ azure_device_registry = [
 [dependencies]
 azure_iot_operations_protocol = { version = "1.0", path = "../azure_iot_operations_protocol" }
 azure_iot_operations_mqtt = { version = "1.0", path = "../azure_iot_operations_mqtt" }
+bytes = { workspace = true, optional = true }
```
cartertinney, line 56: Optional?

rust/azure_iot_operations_services/src/edge_registry.rs
cartertinney: Yeah the issue is that in other contexts a Version ID is just a String. This construct is exclusively used for the get method.

Open to other ideas if you got 'em @vaavva

rust/azure_iot_operations_services/src/edge_registry.rs
RLeclair: It should be the latter, good catch!

rust/azure_iot_operations_services/src/edge_registry/client.rs
RLeclair: Done, need to ask for a vscode feature to make f2 also replaces doc references

rust/azure_iot_operations_services/src/edge_registry/models/xregistry.rs
RLeclair: Yup, done

eng/wot/edge-registry/core-xregistry/ListResourcesRequestPayload.schema.json
RLeclair: Done!

rust/azure_iot_operations_services/src/edge_registry.rs
vaavva: Sync'd with Carter - I'm fine with GetVersionId, but then we also should probably have CreateVersionId or something for the semantics of the Option<String> version id where you can provide it or have the server generate the ID

vaavva (approved):

rust/azure_iot_operations_services/src/edge_registry.rs
RLeclair: I agree, I've added CreateVersionId

rust/azure_iot_operations_services/src/edge_registry/models.rs
RLeclair: Correct, the next PR will add schemas, thing descriptions and thing models

rust/azure_iot_operations_services/src/edge_registry/models/xregistry.rs
RLeclair: This should be an O(1) operation. The Bytes takes ownership of the Vec's existing allocation

rust/azure_iot_operations_services/src/edge_registry.rs
RLeclair: If you get a service error there is potential to retry

I mostly followed what we had before but open to changing it

rust/azure_iot_operations_services/Cargo.toml
RLeclair: I marked it as optional because not all of the features in this crate use this dependency. If you're only interested in ADR you don't need it

rust/azure_iot_operations_services/src/edge_registry.rs
RLeclair: We can chat about this for the next PR

rust/azure_iot_operations_services/src/edge_registry/client.rs
```diff
@@ -0,0 +1,869 @@
...
+    GroupSelection, Label,
+};
+
+const GROUP_TYPE_TOPIC_TOKEN: &str = "groupType";
+const RESOURCE_TYPE_TOPIC_TOKEN: &str = "resourceType";
+const RESOURCE_ID_TOPIC_TOKEN: &str = "resourceId";
+const VERSION_ID_TOPIC_TOKEN: &str = "versionId";
+
+/// Edge Registry client implementation.
+#[allow(clippy::struct_field_names)]
```
vaavva, line 34: is this clippy bypass needed?

vaavva (approved):

PR #1362, opened 2026-06-15
Codegen with error response without normal response

jrdouceur:


RLeclair (approved):
Confirmed output is in expected shape from generating my own sample

PR #1364, opened 2026-06-16
[Services] Add Schema, Thing Document and Thing Description xRegistry Extensions to client

RLeclair:
This PR:
- Implements `Schema`, `Thing_Description` and `Thing Model` extensions for the Edge Registry Client. 
  - Adds `create`, `get`, `list` and `delete` operations for each extension.
- Adds models for each extension

### DIFF
--- a/rust/azure_iot_operations_services/src/edge_registry.rs
+++ b/rust/azure_iot_operations_services/src/edge_registry.rs
@@ -18,6 +18,13 @@ pub mod models;
 
 pub use client::Client;
 
+// `client_gen::JSON_LD11` cannot be used in the models: the generated `client` module flattens both
+// the Thing Description and Thing Model `JSON_LD11` consts into one namespace (and their source
+// modules are private), so the identifier is ambiguous.
+// TODO: consider generating the format identifiers into separate namespaces to avoid this ambiguity.
+/// JSON-LD 1.1 format.
+const JSON_LD11: &str = "JSON-LD/1.1";
+
 // ~~~~~~~~~~~~~~~~~~~SDK Created Structs~~~~~~~~~~~~~~~~~~~~~~~~
 
 /// Represents an error that occurred in the Azure IoT Operations Edge Registry Client implementation.
--- a/rust/azure_iot_operations_services/src/edge_registry.rs
+++ b/rust/azure_iot_operations_services/src/edge_registry.rs
@@ -148,9 +148,6 @@ impl From<GroupId> for Option<String> {
 }
 
 /// Identifies which Version of a Resource to retrieve.
-///
-/// The type parameter `T` is the Version identifier type: [`String`] for generic xRegistry
-/// Versions and [`u64`] for typed extension (Schema, Thing Description, Thing Model) Versions.
 #[derive(Debug, Clone, Default)]
 pub enum GetVersionId<T> {
     /// Retrieve the default Version of the Resource.
--- a/rust/azure_iot_operations_services/src/edge_registry.rs
+++ b/rust/azure_iot_operations_services/src/edge_registry.rs
@@ -148,17 +148,20 @@ impl From<GroupId> for Option<String> {
 }
 
 /// Identifies which Version of a Resource to retrieve.
+///
+/// The type parameter `T` is the Version identifier type: [`String`] for generic xRegistry
+/// Versions and [`u64`] for typed extension (Schema, Thing Description, Thing Model) Versions.
 #[derive(Debug, Clone, Default)]
-pub enum GetVersionId {
+pub enum GetVersionId<T> {
     /// Retrieve the default Version of the Resource.
     #[default]
     ResourceDefault,
     /// Retrieve the Version with the specified identifier.
-    Specified(String),
+    Specified(T),
 }
 
-impl From<GetVersionId> for Option<String> {
-    fn from(value: GetVersionId) -> Self {
+impl<T> From<GetVersionId<T>> for Option<T> {
+    fn from(value: GetVersionId<T>) -> Self {
         match value {
             GetVersionId::ResourceDefault => None,
             GetVersionId::Specified(id) => Some(id),
--- a/rust/azure_iot_operations_services/src/edge_registry.rs
+++ b/rust/azure_iot_operations_services/src/edge_registry.rs
@@ -56,6 +56,70 @@ impl From<rpc_command::invoker::Response<client_gen::EdgeRegistryError>> for Err
     }
 }
 
+impl From<rpc_command::invoker::Response<client_gen::SchemaExtensionError>> for ErrorKind {
+    fn from(value: rpc_command::invoker::Response<client_gen::SchemaExtensionError>) -> Self {
+        Self::ServiceError(value.payload.into())
+    }
+}
+
+impl From<rpc_command::invoker::Response<client_gen::ThingDescriptionExtensionError>>
+    for ErrorKind
+{
+    fn from(
+        value: rpc_command::invoker::Response<client_gen::ThingDescriptionExtensionError>,
+    ) -> Self {
+        Self::ServiceError(value.payload.into())
+    }
+}
+
+impl From<rpc_command::invoker::Response<client_gen::ThingModelExtensionError>> for ErrorKind {
+    fn from(value: rpc_command::invoker::Response<client_gen::ThingModelExtensionError>) -> Self {
+        Self::ServiceError(value.payload.into())
+    }
+}
+
+impl From<client_gen::SchemaExtensionError> for client_gen::EdgeRegistryError {
+    fn from(value: client_gen::SchemaExtensionError) -> Self {
+        client_gen::EdgeRegistryError {
+            code: value.code,
+            detail: value.detail,
+            source: value.source,
+            status: value.status,
+            subject: value.subject,
+            title: value.title,
+            r#type: value.r#type,
+        }
+    }
+}
+
+impl From<client_gen::ThingDescriptionExtensionError> for client_gen::EdgeRegistryError {
+    fn from(value: client_gen::ThingDescriptionExtensionError) -> Self {
+        client_gen::EdgeRegistryError {
+            code: value.code,
+            detail: value.detail,
+            source: value.source,
+            status: value.status,
+            subject: value.subject,
+            title: value.title,
+            r#type: value.r#type,
+        }
+    }
+}
+
+impl From<client_gen::ThingModelExtensionError> for client_gen::EdgeRegistryError {
+    fn from(value: client_gen::ThingModelExtensionError) -> Self {
+        client_gen::EdgeRegistryError {
+            code: value.code,
+            detail: value.detail,
+            source: value.source,
+            status: value.status,
+            subject: value.subject,
+            title: value.title,
+            r#type: value.r#type,
+        }
+    }
+}
+
 impl From<rpc_command::invoker::RequestBuilderError> for ErrorKind {
     fn from(e: rpc_command::invoker::RequestBuilderError) -> Self {
         ErrorKind::ValidationError(e.to_string())
--- a/rust/azure_iot_operations_services/src/edge_registry.rs
+++ b/rust/azure_iot_operations_services/src/edge_registry.rs
@@ -131,7 +131,7 @@ impl From<rpc_command::invoker::RequestBuilderError> for ErrorKind {
 /// Identifies a Group within its Group type for a request.
 #[derive(Debug, Clone, Default)]
 pub enum GroupId {
-    /// Use the cloud default Group of the Group type.
+    /// Use the cloud default Group Id of the Group type.
     #[default]
     CloudDefault,
     /// Use the Group with the specified identifier.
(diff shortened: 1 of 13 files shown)


rust/azure_iot_operations_services/src/edge_registry/models/schema.rs
```diff
@@ -0,0 +1,178 @@
...
+            .versions
+            .into_iter()
+            .map(SchemaVersionXId::from)
+            .collect()
+    }
+}
+
+/// Attributes needed to create a Schema Version.
+#[derive(Debug, Clone)]
+pub struct SchemaVersionAttributes {
```
vaavva, line 150: might be useful to have an optional builder for this since it can't implement default?

rust/azure_iot_operations_services/src/edge_registry/models/schema.rs
```diff
@@ -0,0 +1,178 @@
...
+    /// A URL to a graphical icon for the owning entity.
+    pub icon: Option<String>,
+    /// Queryable Key Value pairs to be added to the Version.
+    pub labels: Vec<Label>,
+    /// The versionId of this Version's ancestor if it has an ancestor.
+    pub ancestor: Option<u64>,
+    /// Content type of the Version document.
+    pub content_type: Option<String>,
+    /// Format identifier of the Version document.
+    pub format: String,
```
vaavva: we should have consts somewhere for the supported formats (allow them to pass in something custom, but make it easy to not mess it up). The consts are defined in the WoT that we can use

rust/azure_iot_operations_services/src/edge_registry/models/xregistry.rs
```diff
@@ -19,7 +19,7 @@ use crate::edge_registry::edge_registry_gen::edge_registry::client as client_gen
 // ~~~~~~~~~~~~~~~~~~~Conversion helpers~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 /// Converts a generated list of `Label` key/value pairs into a vector of [`Label`].
-fn labels_from_gen(labels: Vec<client_gen::Label>) -> Vec<Label> {
+pub(crate) fn labels_from_gen(labels: Vec<client_gen::Label>) -> Vec<Label> {
```
vaavva: if these are used by the other models, should they be in models.rs instead of here? I feel like there could be arguments for either since the others definitely do build off of this one

rust/azure_iot_operations_services/src/edge_registry.rs
```diff
@@ -85,16 +149,16 @@ impl From<GroupId> for Option<String> {
 
 /// Identifies which Version of a Resource to retrieve.
 #[derive(Debug, Clone, Default)]
-pub enum GetVersionId {
+pub enum GetVersionId<T> {
```
vaavva, line 159: should we do the same for VersionXId?

rust/azure_iot_operations_services/src/edge_registry/client.rs
```diff
@@ -851,6 +1609,24 @@ impl Client {
         tokens
     }
 
+    /// Builds the topic tokens for an extension Version request, keyed by the extension's
+    /// Resource-identifier token (e.g. `schemaId`, `thingDescriptionId`, `thingModelId`).
+    fn extension_topic_tokens(id_token: &str, resource_id: String) -> HashMap<String, String> {
```
vaavva:
```suggestion
    fn extension_topic_tokens(resource_id_token: &str, resource_id: String) -> HashMap<String, String> {
```

rust/azure_iot_operations_services/src/edge_registry/client.rs
```diff
@@ -851,6 +1609,24 @@ impl Client {
...
+        HashMap::from([(id_token.to_string(), resource_id)])
+    }
+
+    /// Builds the topic tokens for an extension Version request that carries the Version Id in the
+    /// topic, keyed by the extension's Resource-identifier token.
+    fn extension_version_topic_tokens(
+        id_token: &str,
```
vaavva: just to keep things unambiguous
```suggestion
        resource_id_token: &str,
```

rust/azure_iot_operations_services/src/edge_registry/client.rs
vaavva: would this match the naming of the generic one better?
```suggestion
    fn extension_resource_topic_tokens(id_token: &str, resource_id: String) -> HashMap<String, String> {
```

rust/azure_iot_operations_services/src/edge_registry/client.rs
```diff
@@ -749,6 +858,615 @@ impl Client {
...
+        let response = self
+            .get_schema_version_command_invoker
+            .invoke(request)
+            .await
+            .map_err(ErrorKind::from)?
+            .map_err(ErrorKind::from)?;
+        Ok(response.payload.into())
+    }
+
+    /// List the XIDs of the xRegistry Schema Versions of a Schema.
```
vaavva: shouldn't this have the full list_versions support like with the generic where schema_id is optional and groupId could be all, default, or custom?

rust/azure_iot_operations_services/src/edge_registry/client.rs
RLeclair: Done!

rust/azure_iot_operations_services/src/edge_registry/client.rs
RLeclair: Changed as well

rust/azure_iot_operations_services/src/edge_registry/client.rs
RLeclair: That makes sense, I changed it

rust/azure_iot_operations_services/src/edge_registry/models/xregistry.rs
RLeclair: Yeah, they could go in either place but I think it makes more sense for it to be in `models.rs`. Moved!

rust/azure_iot_operations_services/src/edge_registry.rs
RLeclair: Good idea! I'll make the change

rust/azure_iot_operations_services/src/edge_registry/client.rs
RLeclair: Changed it!

rust/azure_iot_operations_services/src/edge_registry/models/schema.rs
RLeclair: Done, turned into an enum

rust/azure_iot_operations_services/src/edge_registry/models/schema.rs
RLeclair: Done as well, only for extensions

rust/azure_iot_operations_services/src/edge_registry/models/schema.rs
```diff
@@ -117,38 +104,76 @@ impl From<client_gen::SchemaVersionXid> for SchemaVersionXId {
...
-            .collect()
+        value.versions.into_iter().map(Into::into).collect()
+    }
+}
+
+/// Format of a Schema Version document.
+///
+/// The known variants mirror the formats defined by the xRegistry Schema extension; any other
+/// identifier can be supplied via [`Custom`](SchemaFormat::Custom).
+#[derive(Debug, Clone, PartialEq, Eq)]
```
vaavva, line 117: non-exhaustive?

rust/azure_iot_operations_services/src/edge_registry/models/thing_description.rs
```diff
@@ -117,38 +104,76 @@ impl From<client_gen::ThingDescriptionVersionXid> for ThingDescriptionVersionXId
...
+}
+
+// `client_gen::JSON_LD11` cannot be used here: the generated `client` module flattens both the
+// Thing Description and Thing Model `JSON_LD11` consts into one namespace (and their source modules
+// are private), so the identifier is ambiguous. The literal is the single shared value.
+// TODO: consider generating the format identifiers into separate namespaces to avoid this ambiguity.
+impl From<ThingDescriptionFormat> for String {
+    fn from(value: ThingDescriptionFormat) -> Self {
+        match value {
+            ThingDescriptionFormat::JsonLd11 => "JSON-LD/1.1".to_string(),
```
vaavva: can we define a constant ourselves in the meantime just to avoid having the raw string in a few places?

rust/azure_iot_operations_services/src/edge_registry/client.rs
```diff
@@ -983,17 +984,18 @@ impl Client {
...
-        schema_id: String,
+        groups: GroupSelection,
+        schema_id: Option<String>,
         label: Option<Label>,
         timeout: Duration,
-    ) -> Result<Vec<SchemaVersionXId>, Error> {
+    ) -> Result<Vec<VersionXId<u64>>, Error> {
+        let (group_id, all_groups) = Self::group_selection_scope(groups);
         let payload = client_gen::ListVersionsRequestPayload {
             group_type: None,
```
vaavva: this should be filled in so that we filter to only schema extension grouptypes and schema resourcetypes - I think the consts are in the WoTs

rust/azure_iot_operations_services/src/edge_registry/client.rs
```diff
@@ -962,15 +960,18 @@ impl Client {
         Ok(response.payload.into())
     }
 
-    /// List the XIDs of the xRegistry Schema Versions of a Schema.
+    /// List the XIDs of the xRegistry Schema Versions of a schema.
```
vaavva: suggestion```
 /// List the XIDs of the xRegistry Schema Versions matching the provided constraints.
```

File: rust/azure_iot_operations_services/src/edge_registry/models/thing_description.rs (hunk 9 above)
Comment (RLeclair) [reply rust/azure_iot_operations_services/src/edge_registry/models/thing_description.rs]: Sure, I defined a constant at the edge_registry module level

File: rust/azure_iot_operations_services/src/edge_registry/models/schema.rs (hunk 8 above)
Comment (RLeclair) [reply rust/azure_iot_operations_services/src/edge_registry/models/schema.rs:117]: Great catch!

File: rust/azure_iot_operations_services/src/edge_registry/client.rs (hunk 10 above)
Comment (RLeclair) [reply rust/azure_iot_operations_services/src/edge_registry/client.rs]: They are, I added them to this file as well! Thanks for calling it out, I had the wrong idea. I thought that since we have specific request topics for list on each extension the service would know about the types

File: rust/azure_iot_operations_services/src/edge_registry/client.rs (hunk 11 above)
Comment (RLeclair) [reply rust/azure_iot_operations_services/src/edge_registry/client.rs]: Done!

File: rust/azure_iot_operations_services/src/edge_registry/client.rs [hunk 12]
```diff
@@ -30,9 +30,15 @@ const GROUP_TYPE_TOPIC_TOKEN: &str = "groupType";
 const RESOURCE_TYPE_TOPIC_TOKEN: &str = "resourceType";
 const RESOURCE_ID_TOPIC_TOKEN: &str = "resourceId";
 const VERSION_ID_TOPIC_TOKEN: &str = "versionId";
+const SCHEMA_GROUP_TYPE: &str = client_gen::SCHEMA_GROUP_TYPE;
```
vaavva: tiniest of nits - I'd maybe group these by concept rather than full alphabetically - topic token keys vs xid consts, especially since some are hardcoded here vs some being defined in codegen (which is now making me think that codegen should probably generate consts for the known topic tokens lol)

RLeclair: Addressed!

rust/azure_iot_operations_services/src/edge_registry/client.rs
```diff
@@ -749,6 +865,627 @@ impl Client {
...
+        let response = self
+            .get_schema_version_command_invoker
+            .invoke(request)
+            .await
+            .map_err(ErrorKind::from)?
+            .map_err(ErrorKind::from)?;
+        Ok(response.payload.into())
+    }
+
+    /// List the XIDs of the xRegistry Schema Versions of a Schema.
```
vaavva: I think my previous github comment didn't apply correctly - this is what I meant
```suggestion
    /// List the XIDs of the xRegistry Schema Versions matching the provided constraints.
```
would be the same for the other extension list fns

rust/azure_iot_operations_services/src/edge_registry/client.rs
vaavva: That's true, the service could, but I think it's better to provide the correct parameters in the SDK so there's no ambiguity (although I think I agree with you that the service would just ignore what these are set to to avoid an error case)

vaavva (approved):

rust/azure_iot_operations_services/src/edge_registry/client.rs
RLeclair: Done!

PR #1374, opened 2026-06-18
[.NET] Core xRegistry APIs 

maximsemenov80:
Core xRegistry APIs


dotnet/src/Azure.Iot.Operations.Services/EdgeRegistry/Models/Group.cs
```diff
@@ -0,0 +1,75 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Azure.Iot.Operations.Services.EdgeRegistry.Models;
+
+/// <summary>
+/// A Group entity — container for related Resources.
+/// </summary>
+public class Group
```
vaavva: For consistency with Rust, I'd call these `GroupEntity` (ResourceEntity, VersionEntity)

dotnet/src/Azure.Iot.Operations.Services/EdgeRegistry/Models/Group.cs
```diff
@@ -0,0 +1,75 @@
...
+{
+    /// <summary>
+    /// Group identifier.
+    /// </summary>
+    public required string Id { get; set; }
+
+    /// <summary>
+    /// Full XID path, e.g. /schemagroups/mygroup
+    /// </summary>
+    public required string Xid { get; set; }
```
vaavva: small nit - I'd case this XId, since it's short for xRegistry Id, which feels more like two words :)

dotnet/src/Azure.Iot.Operations.Services/EdgeRegistry/Models/ResourceMeta.cs
```diff
@@ -0,0 +1,75 @@
...
+
+    /// <summary>
+    /// Full XID path.
+    /// </summary>
+    public required string Xid { get; set; }
+
+    /// <summary>
+    /// Indicates that this Resource is a reference to another Resource within the same Registry. The XID path of the referenced Resource.
+    /// </summary>
+    public string? Xref { get; set; }
```
vaavva: same here, I'd probably case it XRef

dotnet/src/Azure.Iot.Operations.Services/EdgeRegistry/AnyGroupSelection.cs
```diff
@@ -0,0 +1,40 @@
...
+namespace Azure.Iot.Operations.Services.EdgeRegistry;
+
+/// <summary>
+/// Selects which Groups to query when no Group type is specified: all Groups, or a specific
+/// Group identifier. There is no cloud-default Group without a Group type, so that case is not
+/// representable here.
+/// </summary>
+public readonly struct AnyGroupSelection
+{
+    private enum Kind : byte
```
vaavva: to avoid confusion between `Kind` and `type`, I might call this `Filter`

dotnet/src/Azure.Iot.Operations.Services/EdgeRegistry/EdgeRegistryClient.cs
```diff
@@ -0,0 +1,312 @@
...
+    // ---- Group APIs ----
+
+    /// <inheritdoc/>
+    public async Task<IReadOnlyList<string>> ListGroupsAsync(string groupType, TimeSpan? timeout = null, CancellationToken cancellationToken = default)
+    {
+        cancellationToken.ThrowIfCancellationRequested();
+        ObjectDisposedException.ThrowIf(_disposed, this);
+
+        var output = await _stub.ListGroupsAsync(
+            requestMetadata: null,
```
vaavva: what is this requestMetadata?

dotnet/src/Azure.Iot.Operations.Services/EdgeRegistry/gen.sh
```diff
@@ -0,0 +1,17 @@
...
+dotnet run --project ../../../../wot-codegen/src/Azure.Iot.Operations.ProtocolCompiler/ \
+ --clientThings ../../../../eng/wot/edge-registry/EdgeRegistry.TM.json ../../../../eng/wot/edge-registry/SchemaExtensions.TM.json ../../../../eng/wot/edge-registry/ThingDescriptionExtensions.TM.json ../../../../eng/wot/edge-registry/ThingModelExtensions.TM.json \
+ --schemas '../../../../eng/wot/edge-registry/core-xregistry/*.schema.json' '../../../../eng/wot/edge-registry/schema-extension/*.schema.json' '../../../../eng/wot/edge-registry/thing-description-extension/*.schema.json' '../../../../eng/wot/edge-registry/thing-model-extension/*.schema.json' \
+ --outDir ./Azure.Iot.Operations.Services.EdgeRegistry \
+ --lang csharp \
+ --namespace Generated \
+ --common Generated.Common \
+ --noProj
+
+mv -f ./Azure.Iot.Operations.Services.EdgeRegistry/** .
```
vaavva, line 16: is there not a way to get the output folder you're looking for here using the protocol compiler arguments? I don't think we have to do this for Rust

vaavva (approved):

dotnet/src/Azure.Iot.Operations.Services/EdgeRegistry/gen.ps1
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env pwsh
```
timtay-microsoft: Feel free to not bother with the powershell version of this script. I think all the other gen.ps1 scripts have rotted over time and only the gen.sh scripts still work as expected

dotnet/src/Azure.Iot.Operations.Services/EdgeRegistry/AnyGroupSelection.cs
maximsemenov80: I reworked this type a bit and it is now GroupSelector, so _Kind_ fits better then _Filter_ since Selector implies that it is a filter and _Kind_ says what type of the filter it is.

maximsemenov80: Plus it is a private enum and all user will see is:
``GroupSelector.Default |  GroupSelector.All | GroupSelector.Specific("id") ``

dotnet/src/Azure.Iot.Operations.Services/EdgeRegistry/gen.sh
maximsemenov80: tbh, I adopted existing script, I'll double check

RLeclair (approved):

dotnet/src/Azure.Iot.Operations.Services/EdgeRegistry/EdgeRegistryClient.cs
```diff
@@ -0,0 +1,298 @@
...
+        var output = await _stub.ListGroupsAsync(
+            additionalTopicTokenMap: GroupTopicTokens(groupType),
+            commandTimeout: timeout ?? s_defaultCommandTimeout,
+            cancellationToken: cancellationToken);
+
+        return output.Ids;
+    }
+
+    /// <inheritdoc/>
+    public async Task<Models.GroupEntity> GetGroupAsync(string groupType, GroupId groupId, TimeSpan? timeout = null, CancellationToken cancellationToken = default)
```
timtay-microsoft, line 47: I don't quite understand the GroupId type. It appears to just wrap a single string, so why not just take that string like:

```csharp
    public async Task<Models.GroupEntity> GetGroupAsync(string groupType, string groupId = default, TimeSpan? timeout = null, CancellationToken cancellationToken = default)
```

dotnet/src/Azure.Iot.Operations.Services/EdgeRegistry/IThingModelClient.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Azure.Iot.Operations.Services.EdgeRegistry;
+
+public interface IThingModelClient
+{
+
```
timtay-microsoft, line 8: Is this intentionally empty?

timtay-microsoft: Same question for ISchemaRegistryClient

maximsemenov80: yes, that is for future implementation (next PR for extentions)

dotnet/src/Azure.Iot.Operations.Services/EdgeRegistry/ICoreClient.cs
```diff
@@ -0,0 +1,55 @@
...
+/// Versions identified by XID path components. The extension clients (Schema, Thing Description,
+/// Thing Model) delegate to this surface for the group/resource/version operations that have no
+/// dedicated extension action.
+/// </summary>
+public interface ICoreClient : IAsyncDisposable
+{
+    // ---- Group APIs ----
+
+    /// <summary>Lists the IDs of all Groups of the given <paramref name="groupType"/>.</summary>
+    Task<IReadOnlyList<string>> ListGroupsAsync(string groupType, TimeSpan? timeout = null, CancellationToken cancellationToken = default);
```
timtay-microsoft, line 21: Since these xml headers are inherited into the EdgeRegistryClient, I'd suggest adding the "params" and "returns" sections to the xml

dotnet/src/Azure.Iot.Operations.Services/EdgeRegistry/EdgeRegistryClient.cs
```diff
@@ -0,0 +1,298 @@
...
+    /// <summary>Builds the topic tokens for a Version-scoped request that carries the Version id in the topic.</summary>
+    private static Dictionary<string, string> VersionTopicTokens(string groupType, string resourceType, string resourceId, string versionId)
+    {
+        Dictionary<string, string> tokens = ResourceTopicTokens(groupType, resourceType, resourceId);
+        tokens["versionId"] = versionId;
+        return tokens;
+    }
+
+    /// <inheritdoc/>
+    public async ValueTask DisposeAsync()
```
timtay-microsoft, line 299: Should this client have a ```StopAsync()``` as well? The stub class it wraps has one

dotnet/src/Azure.Iot.Operations.Services/EdgeRegistry/EdgeRegistryClient.cs
RLeclair: The `GroupId` is meant to help users distinguish between a specific group and the cloud default which they might not know the name for

Keeping it as just a string can be confusing as a null string in other APIs doesn't mean default. It could mean something else. For example, when listing versions a null resource ID means that all versions should be listed as opposed to the "default" resource ID should be listed

RLeclair: In short, it is to make "null" less ambiguous. That way a customer knows that if groupId is cloud default that's what they're getting

timtay-microsoft: Interesting. Thanks for clarifying!

timtay-microsoft (approved):

PR #1386, opened 2026-06-29
[Rust][.NET][Services] Add delete options, Core-prefix generic models, and drop dummy delete output

RLeclair:
This PR:

- Removes the placeholder "dummy" output from all delete actions in the Edge Registry WoT and regenerates Rust + .NET.
- Adds `DeleteOptions` (optional `expectedEpoch`) as a shared `DeleteRequestPayload`, threaded through all six delete APIs in both SDKs.
- Modeled as a standalone schema so future delete options can be added in one place.
- Renames the generic xRegistry model types to a `Core` prefix (`CoreGroupEntity`,  `CoreVersionAttributes`, …) to disambiguate them from the extension models.
- Aligns the extension `Version` model field order with the WoT schema, placing `resourceId` before `versionId`.


maximsemenov80 (approved):

eng/wot/edge-registry/core-xregistry/DeleteRequestPayload.schema.json
```diff
@@ -0,0 +1,18 @@
...
+  "description": "Request payload for deleting an xRegistry entity.",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [],
+  "properties": {
+    "groupId": {
+      "description": "Group identifier. Uses the default if not specified.",
+      "type": "string"
+    },
+    "options": {
```
vaavva, line 15: I think this one should be required, it's just defaultable since the internals of it aren't required

dotnet/src/Azure.Iot.Operations.Services/EdgeRegistry/EdgeRegistryClient.Core.cs
```diff
@@ -61,12 +61,12 @@ public async Task<IReadOnlyList<string>> ListGroupsAsync(string groupType, TimeS
     }
 
     /// <inheritdoc/>
-    public async Task DeleteGroupAsync(string groupType, GroupId groupId, TimeSpan? timeout = null, CancellationToken cancellationToken = default)
+    public async Task DeleteGroupAsync(string groupType, GroupId groupId, Models.DeleteOptions options, TimeSpan? timeout = null, CancellationToken cancellationToken = default)
```
vaavva: to be dotnet-y, I think that options should probably have a default value that doesn't specify the epoch here

eng/wot/edge-registry/core-xregistry/DeleteRequestPayload.schema.json
RLeclair: Changed to required!

dotnet/src/Azure.Iot.Operations.Services/EdgeRegistry/EdgeRegistryClient.Core.cs
RLeclair: Wasn't too sure how to do that so on null these become Default (no epoch)

vaavva (approved):

PR #1396, opened 2026-06-30
[Rust][Services][Protocol] Version bump for 07-10-2026 release

RLeclair:
Beta release of the Rust edge registry client


rust/azure_iot_operations_protocol/Cargo.toml
```diff
@@ -3,7 +3,7 @@
 
 [package]
 name = "azure_iot_operations_protocol"
-version = "1.0.1"
+version = "1.1.0"
```
vaavva, line 6: Is this just the invoker change, or the executor one as well? If it's just the invoker change, I think it should only be a patch version bump - it's just a scale bug fix

cartertinney: Agree, I'm not sure what would make this minor instead of patch, please clarify

RLeclair: With the executor change this is now a minor release

vaavva (approved):

PR #1398, opened 2026-07-02
[Rust][Protocols][Services] Have invoker use dispatcher for command responses

vaavva:
Context: Because the invoker used a bounded broadcast channel of size 5, if more than 5 invokes were in flight, it was possible to lose response messages. Switching to using a dispatcher where responses only go to the correct invoke request means that there's no longer a competition for resources between concurrent invokes.

This PR:
- switches the invoker to use a dispatcher instead of a broadcast channel
- moves the common dispatcher logic from the services crate to the protocol crate so that it can be used in both. 
- because it's used in multiple crates, it's now pub, but I put it under a feature flag indicating that it's still under development so that we can make changes to it without it being a breaking change. This also discourages blind customer use, while allowing us to use it in our 1P connectors as well.
- adds unit tests for the dispatcher now that it is used more widely
- switched the dispatcher to use a parking_lot mutex to avoid unwraps


rust/azure_iot_operations_protocol/src/common/dispatcher.rs
```diff
@@ -0,0 +1,291 @@
...
+}
+
+/// Error kind when dispatching a message to a receiver
+#[derive(Debug, Eq, PartialEq, Clone, Error)]
+pub enum DispatchErrorKind {
+    /// The message could not be sent to the receiver
+    #[error("Failed to send message")]
+    SendError,
+    /// There was no receiver with the given ID registered
+    #[error("Receiver with ID '{0}' not found")]
```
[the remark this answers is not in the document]
cartertinney: Good suggestion

rust/azure_iot_operations_protocol/src/common/dispatcher.rs
```diff
@@ -0,0 +1,291 @@
...
+        let dispatcher = create_dispatcher();
+        let _rx1 = dispatcher.register_receiver("a".to_string()).unwrap();
+        let _rx2 = dispatcher.register_receiver("b".to_string()).unwrap();
+        let _rx3 = dispatcher.register_receiver("c".to_string()).unwrap();
+        let mut ids = dispatcher.get_all_receiver_ids();
+        ids.sort();
+        assert_eq!(ids, vec!["a", "b", "c"]);
+    }
+
+    #[test_matrix([true, false], [true, false])]
```
RLeclair, line 171: Aren't both of these the same test case?

rust/azure_iot_operations_protocol/src/common/dispatcher.rs
```diff
@@ -0,0 +1,291 @@
...
+            assert_eq!(result, register_first);
+        }
+        let expected_count = usize::from(register_first && !unregister);
+        assert_eq!(dispatcher.get_all_receiver_ids().len(), expected_count);
+    }
+
+    #[test_case(0, 0; "unregister all with none registered")]
+    #[test_case(1, 1; "unregister all with one registered")]
+    #[test_case(3, 3; "unregister all with three registered")]
+    fn test_unregister_all(register_count: usize, expected_unregistered: usize) {
```
RLeclair: If we're registering and unregistering the same amounts could we have a single argument?

rust/azure_iot_operations_protocol/src/common/dispatcher.rs
```diff
@@ -0,0 +1,291 @@
...
+    #[test]
+    fn test_register_after_unregister_same_id() {
+        let dispatcher = create_dispatcher();
+        let _rx1 = dispatcher.register_receiver("id".to_string()).unwrap();
+        dispatcher.unregister_receiver(&"id".to_string());
+        let mut rx2 = dispatcher.register_receiver("id".to_string()).unwrap();
+        dispatcher
+            .dispatch(&"id".to_string(), "new".to_string())
+            .unwrap();
+        assert_eq!(rx2.try_recv().unwrap(), "new");
```
RLeclair, line 264: Should you check that rx1 doesn't receive anything?

rust/azure_iot_operations_protocol/src/common/dispatcher.rs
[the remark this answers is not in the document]
vaavva: to avoid changing existing behavior that uses this code (since the location of the code changed, but not the behavior), I'm not going to make this adjustment even though it's a good suggestion

rust/azure_iot_operations_protocol/src/common/dispatcher.rs
vaavva: it's a matrix, so it's saying to set up a matrix with every combination between argument 1 having the values [t, f] and argument 2 having the values [t, f], so it tests the full argument combos of (t, t), (f, t), (t, f), and (f, f)

rust/azure_iot_operations_protocol/src/rpc_command/invoker.rs
```diff
@@ -1209,8 +1191,34 @@ where
...
         }
 
-        // Create receiver for response
-        let mut response_rx = self.response_tx.subscribe();
+        // Create correlation id and receiver for response
+        let (correlation_data, mut response_rx) = {
+            loop {
+                let correlation_id = Uuid::new_v4();
+                let correlation_data = Bytes::from(correlation_id.as_bytes().to_vec());
+
```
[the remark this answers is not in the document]
vaavva: suggestion taken

rust/azure_iot_operations_protocol/src/common/dispatcher.rs
vaavva: updated!

rust/azure_iot_operations_protocol/src/common/dispatcher.rs
vaavva: good idea, added that

rust/azure_iot_operations_protocol/src/common/dispatcher.rs
RLeclair: Oh I didn't notice the matrix, that makes sense then!

RLeclair (approved):

rust/azure_iot_operations_services/Cargo.toml
```diff
@@ -15,7 +15,7 @@ publish = true
 [features]
 default = []
 all = ["state_store", "schema_registry", "leased_lock", "azure_device_registry", "edge_registry"]
-state_store = []
+state_store = ["azure_iot_operations_protocol/unstable"]
```
cartertinney: "internal-utils" as a feature name?

cartertinney (approved):

rust/azure_iot_operations_services/Cargo.toml
vaavva: sure, works for me. Updated

cartertinney (approved):

PR #1404, opened 2026-07-07
mRPC Authorization policies

varunpuranik:


vaavva (approved):
LGTM - only thing that might be worth calling out is that the `{executorId}` probably shouldn't appear in the WoT contract/invoker setup as a token (at least not in all cases), since it would be better to hardcode this to the actual Id in those places rather than have it need to be provided by the end user

doc/reference/rpc-authorization.md
```diff
@@ -0,0 +1,280 @@
...
+[AIO MQTT Broker authorization](https://learn.microsoft.com/en-us/azure/iot-operations/manage-mqtt-broker/howto-configure-authorization?tabs=portal) doc.
+
+## Why mRPC is a special case
+
+In ordinary MQTT pub/sub the publisher does not care which subscriber receives a
+message — any authorized subscriber may. mRPC breaks both of those assumptions:
+
+1. A single logical call spans **two correlated topics** — a request topic and a
+   response topic.
+2. The publisher **intends a specific recipient**: a request is meant for *one*
```
cartertinney, line 24: Not entirely true I don't think - don't we allow for load balancing between executors on different client/sessions using shared subscriptions? Or is that only supporting specifying the exact executor instance?

doc/reference/rpc-authorization.md
```diff
@@ -0,0 +1,280 @@
...
+   executor (service); a response is meant for *the one* invoker that issued it.
+
+Therefore authorization must do something pub/sub normally doesn't: **bind each
+topic to the identity of the single client allowed to receive on it.** The lever
```
cartertinney, line 28: A single client can have many executors though. This paragraph appears to conflate a service, a client and an executor identity, all of which are rather different. Services can have more than one client, clients can have more than one invoker/executor.

doc/reference/rpc-authorization.md
```diff
@@ -0,0 +1,280 @@
...
+for that is restricting the `Subscribe` action — not just `Publish` — per
+principal, using per-identity topic segments.
+
+## Topic shapes
+
+Place the routing identity in its own path segment; broker token substitution
+requires a token to be the *entire* segment (e.g. `commands/{executorId}/+` is
```
cartertinney, line 35: To be clear here, currently, there is no actual requirement to use an executorId token at all, as we support any arbitrary pattern. The only requirement is that a token provided must be resolvable with the corresponding token map. There are no special semantic treatment or value to any type of token - tokens are considered to be generic with no special properties for any given token, nor any requirement to use a particular token.

doc/reference/rpc-authorization.md
```diff
@@ -0,0 +1,280 @@
...
+  ```
+
+* Response topic (invoker-keyed), using the default response prefix from
+  [ADR 14](../dev/adr/0014-response-topic-prefix.md):
+
+  ```text
+  clients/{invokerClientId}/commands/{executorId}/{commandName}
+  ```
+
+The invoker learns nothing about the executor's MQTT client ID, and the executor
```
cartertinney, line 51: Under the current paradigm, this isn't quite possible due to the fact that the executor subscribes based on a topic pattern, which is filled in with topic tokens, including the Executor Id, so it is explicitly known if provided. If not specified in the provided tokens, it will get replaced with a "+" wildcard, meaning the executor will end up subscribing to all requests to any executor id.

doc/reference/rpc-authorization.md
```diff
@@ -0,0 +1,280 @@
...
+---
+
+## Option A (preferred) — Executor identified by a pre-known `executorId` (well-known topic)
+
+> [!NOTE]
+> This is the recommended model. Use it when the invoker does not (and should not)
+> know the executor's MQTT client ID, and/or when the executor is horizontally
+> scaled.
+
+The executor is addressed by a **stable, operator-assigned service identity**
```
cartertinney, line 67: How is this different from the shared sub load balancing feature we already have using $partition?

doc/reference/rpc-authorization.md
```diff
@@ -0,0 +1,280 @@
...
+* Response topic (invoker-keyed), using the default response prefix from
+  [ADR 14](../dev/adr/0014-response-topic-prefix.md):
+
+  ```text
+  clients/{invokerClientId}/commands/{executorId}/{commandName}
+  ```
+
+The invoker learns nothing about the executor's MQTT client ID, and the executor
+learns the response topic at runtime from the `ResponseTopic` property of the
+request message. The only identities that appear in the topics are:
```
cartertinney, line 53: Again, not strictly true, the tokens can be anything

doc/reference/rpc-authorization.md
```diff
@@ -0,0 +1,280 @@
...
+> This is the recommended model. Use it when the invoker does not (and should not)
+> know the executor's MQTT client ID, and/or when the executor is horizontally
+> scaled.
+
+The executor is addressed by a **stable, operator-assigned service identity**
+(`executorId`) that is *decoupled from the MQTT client ID*. The invoker publishes
+to a well-known topic such as `commands/temperature-service/getStatus` without
+knowing which pod (client ID) will handle it.
+
+The `executorId` is asserted by the broker from the **SAT-derived attributes** of
```
cartertinney, line 72: Are we suggesting that the user-customizable topic pattern support and support for arbitrary tokens is being dropped? This suggests we are beginning to enforce meaningful semantic value to tokens, rather than treating them as arbitrary values defined by the end user.

cartertinney (commented):
I'm not sure quite how this proposal intersects with the existing reality, and what the ramifications we're willing to accept are.

doc/reference/rpc-authorization.md
RLeclair: That's what I am reading this as too

doc/reference/rpc-authorization.md
vaavva: 👍 that's true, but we do have some special handling for `invokerClientId` and `executorId` on their respective sides in codegen iirc

doc/reference/rpc-authorization.md
vaavva: I don't think so - the way I interpreted this was as a way that a customer could use our existing features to enforce meaningful semantic values in their solutions. I don't read this doc as requiring any code changes, it's just a guide for customers to implement secure rpc with our tools. The topic tokens in this doc are arbitrary as long as the customer is consistent about what they choose

PR #1421, opened 2026-07-13
Move TMs from OPC UA Companion specs to https://github.com/Azure/smd

koepalex:
Removes all Thing Model (TM) definitions for several OPC UA companion specifications, those files will be going forward stored in https://github.com/Azure/smd


timtay-microsoft:
Why are these thing models being added to this repo? Do we expect other users to use these types?

marcschier (approved):

koepalex:
@avishekpant / @timtay-microsoft could you restart the (unrelated) failing check and afterwards go ahead and merge?

PR #1426, opened 2026-07-15
[Rust][Scaffolding] Wire scaffolding SDK deps as { version, path }

yophilav:
## What

Wire the AIO SDK dependencies of `sample_connector_scaffolding` as **`{ version, path }`** — the same way the SDK crates wire their own deps (e.g. [`azure_iot_operations_connector/Cargo.toml`](https://github.com/Azure/iot-operations-sdks/blob/main/rust/azure_iot_operations_connector/Cargo.toml)) — so on **`main`** the sample builds against the **in-repo dev crates via `path`**, not crates.io.

The version requirements are pinned at **minor level** (matching how the SDK crates declare their own deps), so they track `main` without churn on every patch bump:

| Crate | Before (crates.io) | After (`{ version, path }`) | Resolves to (in-repo `main`) |
|---|---|---|---|
| `azure_iot_operations_connector` | `2.0.0-rc3` | `2.1` | `2.1.0` (GA, released 2026-07-24) |
| `azure_iot_operations_services` | `1.2.0-rc4` | `1.4.0-beta1` | `1.4.0-beta1` |
| `azure_iot_operations_protocol` | `1.0` | `1.1` | `1.1.0` |

## Why

Previously the scaffold pinned crates.io versions, so it could silently rot: an SDK API change could land in `main` and the sample would keep compiling against the old published crates. Building against the **dev crates (path)** means **this sample&#39;s CI now fails in `main`** if an SDK change isn&#39;t reflected here — we can&#39;t forget to update it.

During a milestone release the `path` attributes are **stripped**, leaving the version pins — the only release-branch edit needed (after cherry-picking any fixes from `main`).

### Why `services = 1.4.0-beta1` (not a minor-level pin)
A `path` dependency&#39;s version requirement must be **satisfiable by the local crate**, and the in-repo `azure_iot_operations_services` is at `1.4.0-beta1` — a **prerelease**. Cargo prerelease rules require the requirement itself to carry a prerelease tag to match, so a bare `1.4` (or `1.3`) requirement can&#39;t select it. Hence the exact `1.4.0-beta1` pin here, mirroring what `azure_iot_operations_connector/Cargo.toml` itself declares. The release branch re-pins it to the shipped version.

## Notes
- The scaffold **stays excluded** from the workspace so it remains **standalone-buildable** (downstream tooling copies the dir and builds it in isolation). The `exclude` comment is updated to match the new model.
- `create_device_endpoint_client_create_observation()` now returns `Result<_, String>`; propagated with `?` to match `main()`&#39;s fallible setup.

## Validation

Ran the exact `_rust-tool-check.yml` matrix locally **against the in-repo dev crates** (toolchain 1.88 from `rust/rust-toolchain.toml`):

- `cargo fmt --all --check` — clean
- `cargo hack clippy --each-feature --all-targets -- --deny=warnings` — clean
- `cargo machete` — no unused deps
- `cargo hack build --each-feature` — success (compiled `connector 2.1.0`, `services 1.4.0-beta1`, `protocol 1.1.0`, `mqtt 1.1.1` from `/rust/...` paths)

The dedicated `CI-rust-connector-scaffolding` check runs on this PR; its `push: main` trigger also rebuilds the sample on every merge to `main`, catching SDK drift there.


rust/Cargo.toml
```diff
@@ -15,9 +15,8 @@ members = [
 ]
 
 exclude = [
-  # Consider moving this outside the Rust SDK since it is built on released SDK versions,
-  # not the dev version in this directory.
-  # Or, at the very least, outside of sample_applications.
+  # Excluded to keep it standalone-buildable: downstream tooling copies this
+  # directory and builds it in isolation.
   "sample_applications/sample_connector_scaffolding"
```
RLeclair, line 20: I think now that we're specifying the scaffolding AIO dependencies with the path it is a good time for us to add it to members

Any thoughts @cartertinney?

cartertinney (approved):

RLeclair (approved):

rust/Cargo.toml
cartertinney: Discussed offline and resolved

cartertinney (approved):

PR #1429, opened 2026-07-20
Replace PAT usage with SSH deployment key approach

timtay-microsoft:
To access the private repo we use during CI builds (iot-operations-sdks-action), we will now use SSH keys (which last indefinitely) rather than personal access tokens (which need constant renewal).

https://docs.github.com/en/authentication/connecting-to-github-with-ssh/managing-deploy-keys#deploy-keys

I have updated the documentation where applicable, but hopefully no one should have to tinker with this setup again.


doc/dev/github.md
```diff
@@ -16,60 +16,3 @@ By default, Codespaces doesn't allow access to other private repositories, espec
     gh auth login
     gh auth setup-git
     ```
-
-## Developing on a fork
```
timtay-microsoft, line 20: I don't think this setup ever worked, but it won't be applicable anymore now that our pipelines won't take a PAT, so I'm removing this documentation

maximsemenov80 (approved):

PR #1432, opened 2026-07-23
API For SCAN and MGET

olivakar:


rust/azure_iot_operations_services/src/state_store/client.rs
```diff
@@ -361,6 +361,50 @@ impl Client {
         .await
     }
 
+    /// Scans the State Store for keys matching `pattern`, returning one collection of
+    /// results plus an optional continuation token to resume the scan.
+    ///
+    /// Pass `continuation_token: None` for the first request. On each response,
+    /// if [`ScanResult::continuation_token`](state_store::ScanResult::continuation_token)
+    /// is `Some`, pass it to the next `scan` call to get the next page. A `None`
```
vaavva: should we do the continuation internally (or at least offer a convenience function to do so) rather than making them coordinate this? Alternatively, we could create a surface where they don't need to provide the pattern again each time and don't get a continuation token on the last one to make it easier to continue to the next call

rust/azure_iot_operations_services/src/state_store/client.rs
```diff
@@ -361,6 +361,50 @@ impl Client {
...
+        pattern: Vec<u8>,
+        continuation_token: Option<Vec<u8>>,
+        timeout: Duration,
+    ) -> Result<state_store::Response<state_store::ScanResult>, Error> {
+        todo!("Implement scan method");
+    }
+
+    /// Gets the values of multiple keys concurrently.
+    ///
+    /// There is no MGET command in the State Store. This is an SDK side
```
vaavva: what's the reason for us to provide this? It seems easy enough for a customer to do this themselves, and just gets us into weird-ish error scenarios that we have to take opinions on (if some fail, do we retry? Do we provide a convenient way for them to retry those that failed? Do we have a max concurrency limit? If SCAN has pagination, should this have pagination too for large responses?)

rust/azure_iot_operations_services/src/state_store/client.rs
```diff
@@ -361,6 +361,50 @@ impl Client {
         .await
     }
 
+    /// Scans the State Store for keys matching `pattern`, returning one collection of
```
vaavva: should add some documentation on what the format of pattern should be/how it will be interpreted (is it regex/is it prefix matching/etc)

rust/azure_iot_operations_services/src/state_store/client.rs
```diff
@@ -361,6 +361,50 @@ impl Client {
...
+    /// Returns a vector of `Result`s, each containing either the value of the key (if found) or an error.
+    pub async fn mget(
+        &self,
+        keys: &[Vec<u8>],
+        timeout: Duration,
+    ) -> Vec<Result<state_store::Response<Option<Vec<u8>>>, Error>> {
+        todo!("Implement mget method");
+    }
+
     async fn del_internal(
```
vaavva, line 364: nit: I wouldn't put the new APIs in the middle of the delete APIs, just so all of the delete things stay together

rust/azure_iot_operations_services/src/state_store/resp3.rs
```diff
@@ -31,6 +31,10 @@ pub(crate) enum Request {
         key: Vec<u8>,
         options: KeyNotifyOptions,
     },
+    Scan {
+        pattern: Vec<u8>,
+        continuation_token: Option<Vec<u8>>,
```
vaavva, line 36: Do we want to have a more concrete struct for the continuation token so that we can validate that the customer is passing in one in the correct format? Or make this an opaque struct so that they can't create their own unless it's something we've returned to them? The second option might be limiting though, maybe the customer should be able to recreate their own continuation token if they saved the previous one and then crashed or something. But in that case, they should probably have to pass one in in the right format rather than having the service fail the request for being in a bad format, so we should structify (option 1)

rust/azure_iot_operations_services/src/state_store/client.rs
```diff
@@ -361,6 +361,50 @@ impl Client {
...
+    /// continuation token means the scan is complete.
+    ///
+    /// Returns a [`Response`] containing a [`ScanResult`] with the matching keys and an optional continuation token.
+    pub async fn scan(
+        &self,
+        pattern: Vec<u8>,
+        continuation_token: Option<Vec<u8>>,
+        timeout: Duration,
+    ) -> Result<state_store::Response<state_store::ScanResult>, Error> {
+        todo!("Implement scan method");
```
cartertinney: I'm confused - why are we checking in stubs? This would need to be feature branched if you're going to do this so it doesn't interfere with our ability to release.

rust/azure_iot_operations_services/src/state_store/resp3.rs
```diff
@@ -227,6 +235,11 @@ fn serialize_key_notify(key: &[u8], options: &KeyNotifyOptions) -> Vec<u8> {
     builder.get_buffer()
 }
 
+/// Builds a RESP3 payload to `SCAN(pattern, continuation_token)`
+fn serialize_scan(pattern: &[u8], continuation_token: Option<&[u8]>) -> Vec<u8> {
+    print!("serialize_scan called with pattern: {:?}, continuation_token: {:?}\n", pattern, continuation_token);
```
cartertinney: Why print?

cartertinney (changes requested):

rust/azure_iot_operations_services/src/state_store/client.rs
vaavva: +1

rust/azure_iot_operations_services/src/state_store/resp3.rs
```diff
@@ -466,6 +489,10 @@ fn parse_value(payload: &[u8], prefix: &[u8]) -> Result<Vec<u8>, String> {
     }
 }
 
+/// Parses a RESP3 `SCAN` response array into its keys and optional continuation token.
+fn parse_scan(payload: &[u8]) -> Result<(Vec<Vec<u8>>, Option<Vec<u8>>), String> {
```
RLeclair, line 492: Does this need to be a helper if it is only used once?

rust/azure_iot_operations_services/src/state_store/resp3.rs
olivakar: left over. not real.

rust/azure_iot_operations_services/src/state_store/client.rs
olivakar: I am not planning to check this in. Just wanted a review. This is a good way to leave comments.

rust/azure_iot_operations_services/src/state_store/client.rs
olivakar: - ADR 0092 decided not to implement MGET server-side, but the same decision says that "we will have the SDK synthesize an MGET request by sending out multiple GETs in parallel." We can disagree to this and have the application do this. 
-  For retry, I did not want to retry from this implementation. Every result is reported, the application can decide. 
- For convenience, it's easy to retry on application side. Filter the failed ones, call `mget` again with that subset. No new API should be needed.
- Pagination , I don't think it applies, these are N independent gets and doesn't have any position to resume from
- On max concurrency, Agreed, we should cap this internally at n gets in flight. 
This is not a contract limit: the caller still passes as many keys as they like and gets one result per key from one call. It's invisible except that a large call takes longer.
- A separate question is whether we should also cap the number of keys per call. That one would be a contract limit and would belong in the signature. 

Worth discussing.

rust/azure_iot_operations_services/src/state_store/resp3.rs
olivakar: I followed the structure of the previous match arms. I am okay either way.

rust/azure_iot_operations_services/src/state_store/resp3.rs
olivakar: On validating the token: We can do a light check — that the token has the shape `digits;digits;bytes`. We can't validate beyond that (real chain id, topology, exact version) because that needs knowledge of the broker's chains. Also, the broker already returns a malformed-token error if something is wrong. One caveat: the shape check only validates the grammar, not the values.

On opacity: Instead of some concrete `{ chain_id, key }` struct, we can use an opaque `ContinuationToken(Vec<u8>)`. 

On crash recovery: We add `as_bytes()` and `from_bytes()` so a customer can persist a token and rebuild it after a restart. The tradeoff: once `from_bytes` is public, a customer can pass arbitrary bytes, so the final format check still lives in the broker.

rust/azure_iot_operations_services/src/state_store/client.rs
olivakar: `scan(pattern, token)` maps 1:1 to the way ADR described it and is needed for fine-grained control and resuming from a saved token.

But we can add a convenience method so that pattern is not passed multiple times. Are you thinking of a stream here, or do you have a something else in mind?

rust/azure_iot_operations_services/src/state_store/client.rs
olivakar: can be provided as a sample.  no sdk implementation.

rust/azure_iot_operations_services/src/state_store/resp3.rs
arsing: No, you shouldn't validate it. Treat it as an opaque sequence of bytes.

huguesBouvier:
Can you put the test you have run in the description?

rust/azure_iot_operations_services/src/state_store/client.rs
avishekpant: > Pagination , I don't think it applies, these are N independent gets and doesn't have any position to resume from

I believe Valerie meant: how do we handle it if the combined response of those N independent GETs is very large. A solution would be to return the response for a subset of the keys and let the application retry for any missing or failed keys.

rust/azure_iot_operations_services/src/state_store/client.rs
```diff
@@ -562,6 +562,88 @@ impl Client {
...
+    ///
+    /// # Arguments
+    /// * `pattern` - The glob-style pattern to match keys against, as described above.
+    /// * `continuation_token` - `None` for the first request; otherwise the token returned by the
+    ///   previous `scan` response, to resume where the previous page left off.
+    /// * `timeout` - The duration until the client stops waiting for a `Scan` response, rounded up
+    ///   to the nearest second.
+    ///
+    /// Returns a [`Response`] containing a [`ScanResult`] with the matching keys and an optional continuation token.
+    pub async fn scan(
```
avishekpant, line 612: Should we rename it: listKeys over scan? Scan is an implementation detail while listKeys expresses the user's intent and aligns better with common SDK naming conventions. It also gives us flexibility to change the underlying implementation without affecting the public API.
@olivakar / @vaavva - Thoughts?

olivakar:
discussion held. ADR written and implementation complete for SCAN

PR #1451, opened 2026-08-04
Add ADR 33: Defer SDK-level MGET for State Store

olivakar:
Documented the decision not to implement an MGET API for the State Store SDKs, outlining concerns and alternatives.


vaavva (approved):

arsing (approved):

doc/dev/adr/0033-mget.md
```diff
@@ -0,0 +1,98 @@
...
+
+The State Store has no `MGET` command. Broker ADR 0092, "DSS Scan and MGET,"
+decided not to add one, and said the SDKs would provide similar behavior by
+sending several `GET` requests in parallel. This ADR revisits that SDK-side
+decision only; the protocol decision stands.
+
+The gap between what the `MGET` name suggests and what the SDK would actually do
+is the root of the concerns below.
+
+### Concerns
```
avishekpant, line 16: I see these more as trade-offs we have to weigh rather than strict technical problems but I totally agree: maintaining a separate MGET API isn't worth the overhead. Plus locking in those design choices won't work for everyone since clients are naturally going to want different response behaviors.

doc/dev/adr/0033-mget.md
```diff
@@ -0,0 +1,98 @@
...
+  all SDKs, with no strong basis for the values we'd choose.
+
+### Option 2: Ship the helper under a different name
+
+Something like `get_many`, with an application-supplied parallel-request limit,
+documented clearly as sending a separate `GET` request for each key.
+
+- Pro: avoids implying a real `MGET`, and lets each application pick a limit that
+  fits its workload.
+- Con: still permanent cross-SDK API wrapping a plain loop, and still forces us
```
varunpuranik: Why is this a con? This goes for all SDK APIs

doc/dev/adr/0033-mget.md
```diff
@@ -0,0 +1,98 @@
...
+  is per key, there is no single overall timeout, so the call may take much
+  longer than the per-key timeout.
+- **It's a permanent public API for a small amount of application code.** Whatever
+  we pick must be implemented, tested, and kept compatible across all three SDKs,
+  while applications can provide the same behavior with bounded parallel
+  requests.
+
+## Options considered
+
+### Option 1: Ship `MGET` as a helper
```
varunpuranik: I think the ask was to build an MGET api equivalent. Like we do for other APIs. 
So Option 1 and 2 are the same, imo.

doc/dev/adr/0033-mget.md
```diff
@@ -0,0 +1,98 @@
...
+  as one request fails or to wait and return the outcome for every key. It must also
+  define how repeated input keys are handled and whether each repeated key gets
+  its own result. These choices become part of the public API.
+- **We need to decide how the requests run and set limits.** The requests could
+  run one at a time, all at once, or with a fixed number in progress. If the SDK
+  sets that number, every application is bound by our choice. We may also need
+  to limit how many keys are allowed in one call, such as 1,000 or 10,000. Broker
+  ADR 0092 does not define either limit, and we do not have one value that is
+  right for every application and deployment.
+- **Large calls use more memory.** An API that returns all results together may
```
varunpuranik: how is this different than what the application might have to do?

doc/dev/adr/0033-mget.md
```diff
@@ -0,0 +1,98 @@
...
+  need to keep many values and errors in memory until the call finishes.
+- **Retries belong to the application.** A key can be absent, or a single `get()` call
+  can fail transiently. If the API doesn't retry, it adds little beyond separate
+  `GET` requests. If it does retry, we own the count and backoff. Since `timeout`
+  is per key, there is no single overall timeout, so the call may take much
+  longer than the per-key timeout.
+- **It's a permanent public API for a small amount of application code.** Whatever
```
varunpuranik: This can be said about a lot of our APIs,. That does not mean we should not support it in the SDK.

doc/dev/adr/0033-mget.md
```diff
@@ -0,0 +1,98 @@
...
+  its own result. These choices become part of the public API.
+- **We need to decide how the requests run and set limits.** The requests could
+  run one at a time, all at once, or with a fixed number in progress. If the SDK
+  sets that number, every application is bound by our choice. We may also need
+  to limit how many keys are allowed in one call, such as 1,000 or 10,000. Broker
+  ADR 0092 does not define either limit, and we do not have one value that is
+  right for every application and deployment.
+- **Large calls use more memory.** An API that returns all results together may
+  need to keep many values and errors in memory until the call finishes.
+- **Retries belong to the application.** A key can be absent, or a single `get()` call
```
varunpuranik: Why is a retry needed if key is absent? 
Why can the get call fail transiently?

doc/dev/adr/0033-mget.md
```diff
@@ -0,0 +1,98 @@
...
+  can fail transiently. If the API doesn't retry, it adds little beyond separate
+  `GET` requests. If it does retry, we own the count and backoff. Since `timeout`
+  is per key, there is no single overall timeout, so the call may take much
+  longer than the per-key timeout.
```
varunpuranik: What does this mean? Are you suggesting that each key needs a different timeout? Why??

doc/dev/adr/0033-mget.md
```diff
@@ -0,0 +1,98 @@
...
+application can read multiple keys.
+
+- Pro: no new API to keep aligned across SDKs; applications keep control of
+  limits, retries, timeouts, cancellation and error handling.
+- Con: applications write a small amount of code, and may over-parallelize if
+  they don't follow the guidance.
+
+## Decision
+
+**Option 3.** We will not add `MGET` or `get_many` to the State Store SDKs.
```
varunpuranik: I don't see any of the reasons above as unsurmountable.. I feel we can easily add a get_many API with a shape that customers can use, as a helper. 
What I do feel is that this is probably not very pressing. So we can make it a P2 or P3.

doc/dev/adr/0033-mget.md
olivakar: That is fair. Being a permanent API is not a con by itself. The concern is that this API would define client-side behavior that the State Store protocol does not define, and we would need to keep that behavior consistent across SDKs.

doc/dev/adr/0033-mget.md
olivakar: Both options add the same SDK-side multi-key behavior and have the same API design questions. The only difference is the name.  I separated them because  MGET  suggests a real State Store operation, while  get_many  clearly suggests SDK-side behavior. I will combine them into one option and discuss MGET  versus  get_many  as a naming.

doc/dev/adr/0033-mget.md
olivakar: removed.

doc/dev/adr/0033-mget.md
vaavva: I actually don't think that's true - almost all of our Service Client SDKs are _very_ close to the WoT/DTDL contract. The two major exceptions are:
1. Observation/notification patterns, where we combine a mRPC request to observe with a telemetry receiver that listens for notifications and wrap that nicely for the user.
2. Health status reporting convenience background task with the ADR service

However, both of these two exceptions obfuscate a _large_ amount of code for the user that would be challenging for them to handle the logic for, and the code that we provide is _always_ exactly the logic that the customer needs (because of the service behavior, there is no other possible way they should be coding these), which I think is fairly different than us providing a convenience wrapper for a for loop that may or may not actually have the behavior that the customer wants

doc/dev/adr/0033-mget.md
olivakar: A missing key should not be retried. The  GET  API returns it as a successful result with no value. A  GET  can still fail because of a timeout, State Store error, or underlying protocol error. The concern is whether an SDK-level  MGET  should retry those failed requests. If it does, the SDK must choose which errors to retry, how many times, and with what delay. Different applications may need different retry behavior. I updated the wording to clarify this.

doc/dev/adr/0033-mget.md
olivakar: No, I am not suggesting that each key needs a different timeout. The question is how one timeout passed to  MGET  should apply to its multiple  GET  requests. Updated wording.

doc/dev/adr/0033-mget.md
olivakar: reworded

avishekpant (approved):

PR #1455, opened 2026-08-10
ADR 34 for State Store SCAN API

olivakar:
This ADR outlines the design decisions for the State Store SCAN API, including pagination options, result delivery methods, and continuation token handling.

### DIFF
--- a/doc/dev/adr/0034-state-store-scan.md
+++ b/doc/dev/adr/0034-state-store-scan.md
@@ -1,9 +1,8 @@
-# ADR 34: State Store SCAN API
+# ADR 34: State Store Filter Keys API
 
 ## Context
 
-Broker ADR 0092, "DSS Scan and MGET"
-(`Azure-MQ/docs-dev/adr/dmqtt/0092-DSS-scan-mget.md`), defines the State Store
+Broker ADR 0092, "DSS Scan and MGET" , defines the State Store
 `SCAN` protocol and selects cursor-based pagination. A response contains one
 page of matching keys and, unless the scan is complete, a continuation token
 for the next request.
@@ -44,8 +43,8 @@ also add public API that must be maintained.
 
 Keeping the token internal still supports lazy paging. Token access can be
 added later without breaking existing applications, but cannot be removed
-after becoming public. Without token access, a scan restarts after an
-application restart.
+after becoming public. Without token access, a filter-keys operation restarts
+after an application restart.
 
 The `redis-rs` crate exposes async SCAN through
 [`AsyncCommands::scan`](https://docs.rs/redis/latest/redis/trait.AsyncCommands.html#method.scan)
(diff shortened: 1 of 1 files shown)


arsing (approved):

doc/dev/adr/0034-state-store-scan.md
```diff
@@ -0,0 +1,128 @@
+# ADR 34: State Store SCAN API
+
+## Context
+
+Broker ADR 0092, "DSS Scan and MGET"
+(`Azure-MQ/docs-dev/adr/dmqtt/0092-DSS-scan-mget.md`), defines the State Store
```
avishekpant: Let's not add details about internal repo on public repo. Let's remove the reference to broker ADR.

doc/dev/adr/0034-state-store-scan.md
```diff
@@ -0,0 +1,128 @@
...
+while let Some(keys) = scan.next().await? {
+     // Process this page of keys.
+}
+```
+
+This keeps one page in memory and makes requests only when the application
+calls `next()`. The application may stop when its processing condition is met,
+without requesting the remaining pages.
+
+### Use an SDK-owned async `next()` pattern
```
vaavva, line 102: could be worth throwing something in here about the downside of using stream for this application as well - Stream has much more capabilities that aren't needed for this application. For example, the main advantage of Stream is it's filtering capabilities, but there shouldn't be any need to filter on `scan` since it's already effectively a filtered list keys request

vaavva (approved):

doc/dev/adr/0034-state-store-scan.md
```diff
@@ -0,0 +1,128 @@
...
+stream or async-iterator dependency.
+
+The object keeps the client reference, pattern, timeout, current token, and
+completion state. It cannot outlive the State Store client.
+
+### Keep continuation tokens internal
+
+The pagination object keeps and forwards the continuation token internally. It
+does not expose the token. After an application restart, a scan starts again
+from the beginning. This is an accepted limitation.
```
cartertinney: Do we want to clarify why this is an accepted limitation (the impracticality of an application stashing continuation tokens, etc.) or would that expand the scope too much?

cartertinney (approved):

doc/dev/adr/0034-state-store-scan.md
```diff
@@ -0,0 +1,128 @@
...
+### Fetch pages lazily
+
+The SDK returns a caller-driven pagination object. Each async `next()` call
+requests one page and uses the previous response token. It returns `None` when
+the broker does not return a continuation token.
+
+Conceptually:
+
+```rust
+pub struct ScanPager<'a> {
```
RLeclair: Maybe worth explicitly calling out the lifetime here so it is clear the pager doesn't exceed the client's lifetime:

```rust
pub struct ScanPager<'client> ...

impl Client {
    pub fn scan<'client>(
        ...
    ) -> Result<ScanPager<'client>, Error> { ... }
}
```

RLeclair: In the implementation that isn't needed

RLeclair (approved):

avishekpant (approved):

PR #1462, opened 2026-08-19
Scan Implementation

olivakar:


rust/azure_iot_operations_services/src/state_store/resp3.rs
```diff
@@ -296,6 +428,7 @@ impl PayloadSerialize for Response {
             _ if payload.starts_with(Self::RESPONSE_LENGTH_PREFIX) => Ok(Response::Value(
                 parse_value(payload, Self::RESPONSE_LENGTH_PREFIX)?,
             )),
+            _ if payload.starts_with(b"*") => Ok(parse_scan(payload)?),
```
vaavva: Save `*` as a const like the others: `SCAN_RESPONSE_PREFIX`

vaavva: have this be `Self::parse_scan(...` like `Self::parse_error(...` just for consistency

rust/azure_iot_operations_services/src/state_store/resp3.rs
```diff
@@ -506,6 +639,38 @@ mod tests {
...
+            .unwrap(),
+            Response::KeysScanned {
+                keys: vec![b"key1".to_vec(), b"key2".to_vec()],
+                continuation_token: None,
+            }
+        );
+    }
+
+    #[test]
+    fn test_scan_response_empty_page_with_binary_continuation_token() {
```
vaavva, line 678: should we have a test for a non-empty page with continuation token too?

rust/azure_iot_operations_services/src/state_store/resp3.rs
```diff
@@ -229,6 +249,113 @@ fn serialize_key_notify(key: &[u8], options: &KeyNotifyOptions) -> Vec<u8> {
...
+                self.index,
+                self.payload
+            ));
+        }
+        self.index += 1;
+
+        let remaining = self
+            .payload
+            .get(self.index..)
+            .ok_or_else(|| format!("Invalid RESP3 payload: {:?}", self.payload))?;
```
vaavva: I think you could simplify this a bit by doing strip_prefix like in parse_value (https://github.com/Azure/iot-operations-sdks/pull/1462/changes#diff-7b3b13a4fccc18a6e18dea08a69a820006308d38590bf0b2fb7915bd0f75b11aR573) to combine the two error checks

rust/azure_iot_operations_services/src/state_store/resp3.rs
```diff
@@ -229,6 +249,113 @@ fn serialize_key_notify(key: &[u8], options: &KeyNotifyOptions) -> Vec<u8> {
...
+        self.index = element_end;
+        Ok(value)
+    }
+
+    fn is_complete(&self) -> bool {
+        self.index == self.payload.len()
+    }
+}
+
+fn parse_scan(payload: &[u8]) -> Result<Response, String> {
```
vaavva: have you been able to test the serialize and deserialize with the state store yet?
One other thing that would be good to verify is to make sure there isn't any new error we need to handle getting back if an invalid pattern is sent

rust/azure_iot_operations_services/src/state_store/client.rs
```diff
@@ -741,6 +862,47 @@ mod tests {
         ));
     }
 
+    #[tokio::test]
```
vaavva, line 865: since `filter_keys` isn't async, these might not need to be tokio tests

olivakar: Tried this — both tests panic as plain  #[test] :
panicked at azure_iot_operations_protocol/src/rpc_command/invoker.rs:926:
there is no reactor running, must be called from the context of a Tokio 1.x runtime

rust/azure_iot_operations_services/src/state_store/resp3.rs
olivakar: in process for e2e testing

rust/azure_iot_operations_services/src/state_store/client.rs
```diff
@@ -59,6 +59,73 @@ impl KeyObservation {
     // that was observed where the receiver was dropped and a key that was never observed
 }
 
+/// A caller-driven, page-at-a-time filter of State Store keys.
+///
+/// Call [`FilterKeysPager::next`] to request each page. The continuation token is managed
+/// internally, and the pager cannot outlive the [`Client`] that created it.
+pub struct FilterKeysPager<'a> {
+    client: &'a Client,
```
cartertinney, line 67: Does this need to be a ref? Generally clients are clonable extremely cheaply, although now that I think about it, DSS might not be for reasons

rust/azure_iot_operations_services/src/state_store/client.rs
```diff
@@ -59,6 +59,73 @@ impl KeyObservation {
...
+    /// # Errors
+    /// [`struct@Error`] of kind [`InvalidArgument`](ErrorKind::InvalidArgument) if the `timeout`
+    /// that the pager was created with is zero or > `u32::MAX`
+    ///
+    /// [`struct@Error`] of kind [`ServiceError`](ErrorKind::ServiceError) if the State Store returns an Error response
+    ///
+    /// [`struct@Error`] of kind [`UnexpectedPayload`](ErrorKind::UnexpectedPayload) if the State Store returns a response that isn't valid for this request
+    ///
+    /// [`struct@Error`] of kind [`AIOProtocolError`](ErrorKind::AIOProtocolError) if there are any underlying errors from [`rpc_command::Invoker::invoke`]
+    pub async fn next(&mut self) -> Result<Option<Vec<Vec<u8>>>, Error> {
```
cartertinney, line 90: I assume `Vec<Vec<u8>>` means that a key is `Vec<u8>`, and we are providing a `Vec` of them?

cartertinney, line 90: What's the cancel safety here?

rust/azure_iot_operations_services/src/state_store/resp3.rs
vaavva: sure, but have you been able to manually test yet?

vaavva (approved):
Approving, assuming this has been tested against the service, at least locally, just to make sure all of the serialization and such in reality matches the documentation

rust/azure_iot_operations_services/src/state_store/client.rs
olivakar: yes

olivakar: It's cancel safe. When we build the request we copy the continuation token, and we only update the token and the  complete  flag after the invoke comes back. So if the future gets dropped partway through , the pager is left exactly as it was, and calling  next()  again just asks for the same page.

cartertinney (approved):

rust/azure_iot_operations_services/src/state_store/client.rs
```diff
@@ -59,6 +59,73 @@ impl KeyObservation {
...
+/// internally, and the pager cannot outlive the [`Client`] that created it.
+pub struct FilterKeysPager<'a> {
+    client: &'a Client,
+    pattern: Vec<u8>,
+    timeout: Duration,
+    continuation_token: Option<Vec<u8>>,
+    complete: bool,
+}
+
+impl FilterKeysPager<'_> {
```
varunpuranik, line 74: Thinking through this a bit more, I feel Filter does not convey the meaning of the API properly. This is more of a "get" API than a "filter out or exclude" API. 
I would suggest we rename it to something to listKeys, iterateKeys, getKeys or scanKeys (my preference in order).

PR #1466, opened 2026-08-24
[Rust][MQTT][Connector] Allow AIO `metriccategory` metadata to be configured via API

cartertinney:
- API for MQTT and connector crates both allow the configuration of the `metriccategory` metadata that is included in the user properties on MQTT CONNECT.
- This allows for applications to be identified by AIO, e.g. REST Connector

### DIFF
--- a/rust/azure_iot_operations_connector/src/base_connector.rs
+++ b/rust/azure_iot_operations_connector/src/base_connector.rs
@@ -105,10 +105,10 @@ pub struct Options {
     #[builder(default = "Duration::from_secs(5)")]
     filemount_debounce_duration: Duration,
 
-    /// Value for the `metriccategory` that is used to categorize the internal traffic between 
-    /// the connector and other AIO services - Broker, Schema Registry, DSS. 
+    /// Value for the `metriccategory` that is used to categorize the internal traffic between
+    /// the connector and other AIO services - Broker, Schema Registry, DSS.
     /// See https://learn.microsoft.com/en-us/azure/iot-operations/reference/observability-metrics-mqtt-broker#category for details.
-    #[builder(default = "\"aiosdk-rust\".to_string()", setter(into))]
+    #[builder(default = "\"aiosdk-rust-connector\".to_string()", setter(into))]
     metric_category: String,
 
     /// Reconnect policy used by the MQTT Session.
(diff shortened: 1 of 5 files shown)


rust/azure_iot_operations_connector/src/base_connector.rs
```diff
@@ -104,6 +105,10 @@ pub struct Options {
     #[builder(default = "Duration::from_secs(5)")]
     filemount_debounce_duration: Duration,
 
+    /// Value for the `metriccategory` MQTT CONNECT user property
+    #[builder(default = "\"aiosdk-rust\".to_string()", setter(into))]
```
vadim-kovalyov: We just had a discussion with @vaavva that this might be used for new xregistry as well. So I think we should be more generic here and say something like:

```
Value for the metric category that is used to categorize the internal traffic between 
the connector and other AIO services - Broker, Schema Registry, DSS. 
See https://learn.microsoft.com/en-us/azure/iot-operations/reference/observability-metrics-mqtt-broker#category for details.
```

rust/azure_iot_operations_mqtt/src/aio.rs
```diff
@@ -13,4 +13,7 @@ pub struct AIOBrokerFeatures {
     /// Indicates if the Session should use AIO persistence
     #[builder(default = "false")]
     pub(crate) persistence: bool,
+    /// Value for the `metriccategory` MQTT CONNECT user property
+    #[builder(default = "\"aiosdk-rust\".to_string()", setter(into))]
+    pub(crate) metric_category: String,
```
vaavva, line 18: can you create a getter for this value on the SessionManagedClient like we have for client_id so that I can get it in the edge registry client and attach it to messages sent from there?

rust/azure_iot_operations_connector/src/base_connector.rs
vaavva: Should we have any format recommendations in the doc comment as well, like `<connectorType>-connector`? I don't know whether we want to recommend `aio-<connectorType>-connector` for non 1p

cartertinney: Updated the comment, and provided an alternate default for connectors specifically - `aiosdk-rust-connector`

vaavva (approved):

vaavva (approved):

PR #1471, opened 2026-08-27
Add typed Edge Registry APIs for listing schemas, thing descriptions, and thing models in the Rust and .NET SDKs.

bilalsellak:
Add typed Edge Registry APIs for listing schemas, thing descriptions, and thing models in the Rust and .NET SDKs.

## Changes

- Add `list_schemas` / `ListSchemasAsync`.
- Add `list_thing_descriptions` / `ListThingDescriptionsAsync`.
- Add `list_thing_models` / `ListThingModelsAsync`.
- Support default, all, and specific group selection.
- Support resource label filtering.
- Add .NET integration coverage for the new APIs.
- Add a Rust Edge Registry example with retry handling for eventual consistency.

### Local integration

- Ran the Rust SDK sample against a local MQTT broker and Edge Registry test host.
- Ran three .NET Edge Registry integration tests against the same host.
- Built and tested TinyKube and Azure-NBC using local SDK path dependencies.

### Standalone WASM end-to-end test

### DIFF
--- a/rust/azure_iot_operations_services/examples/edge_registry_client.rs
+++ b/rust/azure_iot_operations_services/examples/edge_registry_client.rs
@@ -1,0 +1,327 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use std::error::Error;
+use std::future::Future;
+use std::io;
+use std::time::Duration;
+
+use azure_iot_operations_mqtt::aio::connection_settings::MqttConnectionSettingsBuilder;
+use azure_iot_operations_mqtt::control_packet::{
+    PublishProperties, QoS, RetainOptions, SubscribeProperties, TopicFilter, TopicName,
+};
+use azure_iot_operations_mqtt::session::{
+    Session, SessionExitHandle, SessionManagedClient, SessionOptionsBuilder,
+};
+use azure_iot_operations_protocol::application::ApplicationContextBuilder;
+use azure_iot_operations_protocol::common::aio_protocol_error::AIOProtocolErrorKind;
+use azure_iot_operations_services::edge_registry::models::{
+    SchemaFormat, SchemaVersionAttributesBuilder, ThingDescriptionFormat,
+    ThingDescriptionVersionAttributesBuilder, ThingModelFormat, ThingModelVersionAttributesBuilder,
+};
+use azure_iot_operations_services::edge_registry::{
+    self, Client, GetVersionId, GroupId, GroupSelection, Label,
+};
+use bytes::Bytes;
+use env_logger::Builder;
+
+const TIMEOUT: Duration = Duration::from_secs(10);
+const LIST_RETRY_ATTEMPTS: u32 = 30;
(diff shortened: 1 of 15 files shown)


rust/azure_iot_operations_services/src/edge_registry/client.rs
```diff
@@ -983,6 +983,37 @@ impl Client {
         Ok(response.payload.into())
     }
 
+    /// List the XIDs of the xRegistry Schemas matching the provided constraints.
```
vaavva: Tiniest nit for consistency - can you put these at the top of their extension sections, since they are the operation for the highest layer (group/resource/version)? That way they stay in the same order as the Core APIs

rust/azure_iot_operations_services/examples/edge_registry_client.rs
```diff
@@ -0,0 +1,362 @@
...
+    if exit_handle.try_exit().is_err() {
+        exit_handle.force_exit();
+    }
+
+    operation_result?;
+    shutdown_result?;
+    Ok(())
+}
+
+async fn graph_round_trip_if_configured(
```
vaavva: I'm not sure this should be in an SDK sample - seems wasm specific? We want these to be as simple as possible to just show the SDK capabilities

rust/azure_iot_operations_services/examples/edge_registry_client.rs
```diff
@@ -0,0 +1,362 @@
...
+        || {
+            client.list_schema_versions(
+                GroupSelection::Default,
+                Some(schema_id.clone()),
+                None,
+                None,
+                TIMEOUT,
+            )
+        },
+        |versions| {
```
vaavva: For a sample, I would probably just have it print out the items in the list rather than treating this like a test and validating that the correct ones are there

rust/azure_iot_operations_services/examples/edge_registry_client.rs
```diff
@@ -0,0 +1,362 @@
...
+        "labeled Thing Model was not listed",
+        LIST_RETRY_TIMEOUT,
+    )
+    .await?;
+
+    log::info!("Listed Thing Description {thing_description_id} and Thing Model {thing_model_id}");
+    Ok(())
+}
+
+#[cfg(test)]
```
vaavva: similar here, examples in the SDK are for the customers, they're not intended to be our e2e tests (would be great to add some of this to an e2e network test, but I don't think we have those set up yet for edge registry)

vaavva (approved):
Reviewed Rust Portions, skimmed .net portions. Rust SDK code looks great, have some comments for adjusting the sample

rust/azure_iot_operations_services/examples/edge_registry_client.rs
bilalsellak: Ahh that makes sense. Ill update it

timtay-microsoft (approved):
.NET stuff LGTM

vaavva (approved):

PR #1472, opened 2026-08-27
[Rust][Services] Add resource id hashing and originalid label helper

RLeclair:
This PR:
- Adds `derive_resource_id` , an Edge Registry helper that derives an xRegistry- and cloud-conforming Resource identifier from an arbitrary identifier
- Uses lowercase hex SHA-256, matching `xregistry_core::compute_document_hash` in the Edge Registry service so identifiers stay reproducible across SDKs
- Records the original identifier verbatim in the caller's label vector under the new `ORIGINAL_ID_LABEL_KEY` (`originalid`) constant, replacing any existing entry so repeat calls don't accumulate duplicates

### DIFF
--- a/rust/azure_iot_operations_services/src/edge_registry.rs
+++ b/rust/azure_iot_operations_services/src/edge_registry.rs
@@ -259,15 +259,23 @@ impl From<client_gen::Label> for Label {
 /// The label key under which the original, pre-derivation identifier is recorded by
 /// [`derive_resource_id`].
 ///
-/// A lookup filters on this key with the original identifier as the value, not the derived
-/// Resource identifier.
+/// The label belongs on the Resource, and a lookup filters on this key with the original
+/// identifier as the value, not the derived Resource identifier.
 pub const ORIGINAL_ID_LABEL_KEY: &str = "originalid";
 
 /// Derives a conforming Resource identifier from an arbitrary identifier, recording the original
 /// in `labels` under [`ORIGINAL_ID_LABEL_KEY`].
 ///
+/// The derived identifier is the lowercase hex SHA-256 of `original_id`. It is always 64 characters
+/// drawn from `[0-9a-f]`, which satisfies both the xRegistry identifier rules and the stricter
+/// cloud rules.
+///
 /// Any existing [`ORIGINAL_ID_LABEL_KEY`] entry is replaced, so repeated calls against the same
 /// `labels` do not accumulate duplicates.
+///
+/// # Errors
+/// [`struct@Error`] of kind [`ValidationError`](ErrorKind::ValidationError) if `original_id` is
+/// empty. `labels` is left untouched in that case.
 ///
 /// # Example
 /// ```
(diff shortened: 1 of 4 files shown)


rust/azure_iot_operations_services/src/edge_registry.rs
```diff
@@ -251,3 +253,157 @@ impl From<client_gen::Label> for Label {
...
+/// assert_eq!(labels[0].key, ORIGINAL_ID_LABEL_KEY);
+/// assert_eq!(labels[0].value, "urn:azureiot:aio:dev:ep:opcua:asset:td.g");
+/// # Ok::<(), azure_iot_operations_services::edge_registry::Error>(())
+/// ```
+pub fn derive_resource_id(original_id: &str, labels: &mut Vec<Label>) -> Result<String, Error> {
+    if original_id.is_empty() {
+        return Err(ErrorKind::ValidationError("original_id must not be empty".to_string()).into());
+    }
+
+    labels.retain(|label| label.key != ORIGINAL_ID_LABEL_KEY);
```
vaavva: I would say that we should only overwrite exact matches (key and value)

rust/azure_iot_operations_services/src/edge_registry.rs
```diff
@@ -251,3 +253,157 @@ impl From<client_gen::Label> for Label {
...
+    labels.push(Label {
+        key: ORIGINAL_ID_LABEL_KEY.to_string(),
+        value: original_id.to_string(),
+    });
+
+    Ok(HEXLOWER.encode(&Sha256::digest(original_id.as_bytes())))
```
vaavva: don't we only want to encode if we need to?

rust/azure_iot_operations_services/src/edge_registry.rs
```diff
@@ -251,3 +253,157 @@ impl From<client_gen::Label> for Label {
...
+/// # };
+/// let mut labels = vec![];
+/// let resource_id = derive_resource_id("urn:azureiot:aio:dev:ep:opcua:asset:td.g", &mut labels)?;
+///
+/// assert_eq!(resource_id.len(), 64);
+/// assert_eq!(labels[0].key, ORIGINAL_ID_LABEL_KEY);
+/// assert_eq!(labels[0].value, "urn:azureiot:aio:dev:ep:opcua:asset:td.g");
+/// # Ok::<(), azure_iot_operations_services::edge_registry::Error>(())
+/// ```
+pub fn derive_resource_id(original_id: &str, labels: &mut Vec<Label>) -> Result<String, Error> {
```
vaavva: thoughts on this? So that folks can pass in "urn:azureiot:aio:dev:ep:opcua:asset:td.g", but if they had to `format!()` the value (likely), then they can use the already allocated String
```suggestion
pub fn derive_resource_id(original_id: impl Into<String>, labels: &mut Vec<Label>) -> Result<String, Error> {
```

vaavva: or just taking String, since the only reason you wouldn't use it is if it's empty

rust/azure_iot_operations_services/src/edge_registry.rs
```diff
@@ -251,3 +253,157 @@ impl From<client_gen::Label> for Label {
...
+/// [`struct@Error`] of kind [`ValidationError`](ErrorKind::ValidationError) if `original_id` is
+/// empty. `labels` is left untouched in that case.
+///
+/// # Example
+/// ```
+/// # use azure_iot_operations_services::edge_registry::{
+/// #     ORIGINAL_ID_LABEL_KEY, derive_resource_id,
+/// # };
+/// let mut labels = vec![];
+/// let resource_id = derive_resource_id("urn:azureiot:aio:dev:ep:opcua:asset:td.g", &mut labels)?;
```
vaavva: I was kind of thinking of this format, just since most people will probably start with an empty vec, and it makes it clearer to remember to use the output vec (I don't have a super strong opinion though)
```suggestion
/// let (resource_id, resource_labels) = derive_resource_id("urn:azureiot:aio:dev:ep:opcua:asset:td.g", vec![])?;
```

```
 fn derive_resource_id(original_id: String, labels: mut Vec<Label>) -> Result<(String, Vec<Label>), Error>
```

rust/azure_iot_operations_services/src/edge_registry.rs
RLeclair: Changed!

rust/azure_iot_operations_services/src/edge_registry.rs
RLeclair: Changed the signature and the one in of labels too

rust/azure_iot_operations_services/src/edge_registry.rs
RLeclair: Makes sense, I also changed the parameter to be called resource_labels to make it clear

rust/azure_iot_operations_services/src/edge_registry.rs
RLeclair: Done!

rust/azure_iot_operations_services/src/edge_registry.rs
```diff
@@ -310,22 +342,44 @@ mod tests {
...
+    #[test_case("abc-"; "trailing hyphen")]
+    #[test_case("Abc"; "uppercase")]
+    #[test_case("a_c"; "underscore")]
+    #[test_case("a\u{00f1}b"; "non ascii")]
     #[test_case("urn:azureiot:aio:dev:ep:opcua:asset:td.g"; "wot identifier")]
     #[test_case(&"x".repeat(512); "longer than an xregistry identifier")]
-    fn derived_identifier_conforms_to_the_cloud_naming_rules(original_id: &str) {
-        let resource_id = derive_resource_id(original_id, &mut vec![]).unwrap();
+    fn hashes_a_non_conforming_identifier(original_id: &str) {
+        let (resource_id, _) = derive_resource_id(original_id, vec![]).unwrap();
```
vaavva: might as well not `_` the labels and validate that the original was added there correctly without getting distorted or anything

vaavva (approved):

rust/azure_iot_operations_services/src/edge_registry.rs
RLeclair: Done!

vaavva (approved):

PR #1475, opened 2026-08-28
[.NET][Services] Add resource id hashing and originalid label helper

maximsemenov80:
Mirrors the Rust derive_resource_id helper (#1472). Identifiers that already satisfy the cloud naming rules are used verbatim; every other identifier becomes the lowercase hex SHA-256 of the original, which is recorded on the Resource under the originalid label so it stays queryable.

### DIFF
--- a/dotnet/test/Azure.Iot.Operations.Services.UnitTests/EdgeRegistry/ResourceIdTests.cs
+++ b/dotnet/test/Azure.Iot.Operations.Services.UnitTests/EdgeRegistry/ResourceIdTests.cs
@@ -14,6 +14,12 @@ public class ResourceIdTests
     /// the input. Every SDK must produce this value for identifiers to be reproducible across
     /// implementations.
     /// </summary>
+    /// <remarks>
+    /// This doubles as a cross-language test: the Rust <c>hashes_to_lowercase_sha256_hex</c> test in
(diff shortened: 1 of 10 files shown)


dotnet/test/Azure.Iot.Operations.Services.UnitTests/EdgeRegistry/ResourceIdTests.cs
```diff
@@ -0,0 +1,137 @@
...
+    /// Pins the hash to the one the Edge Registry service computes, the lowercase hex SHA-256 of
+    /// the input. Every SDK must produce this value for identifiers to be reproducible across
+    /// implementations.
+    /// </summary>
+    [Fact]
+    public void HashesToLowercaseSha256Hex()
+    {
+        (string resourceId, _) = ResourceId.Derive("Test-Document");
+
+        Assert.Equal("1ae8481659aaf8fe08cb58818b1793849756b0f526d835e5106cfb76e558cfdd", resourceId);
```
timtay-microsoft, line 28: Maybe stick comment in here that says that this is also sort of a cross-language test since we want to ensure that the same inputs to our helper hash functions yield the same results regardless of language.

timtay-microsoft (approved):

timtay-microsoft (approved):

PR #1477, opened 2026-08-28
[.NET][Services][EdgeRegistry] Add persistence user property to write  requests

maximsemenov80:
Mirrors Rust #1474. Every create operation now takes CreateOptions, which sends the aio-persistence user property. Unlike the state store, persistence defaults to on: registry entities should always survive a restart, so this is opt-out rather than opt-in.

### DIFF
--- a/dotnet/src/Azure.Iot.Operations.Services/EdgeRegistry/CreateOptions.cs
+++ b/dotnet/src/Azure.Iot.Operations.Services/EdgeRegistry/CreateOptions.cs
@@ -9,11 +9,6 @@ namespace Azure.Iot.Operations.Services.EdgeRegistry;
 public class CreateOptions
 {
     /// <summary>
-    /// The default options, which persist the created entity.
-    /// </summary>
-    public static CreateOptions Default => new();
-
-    /// <summary>
     /// Whether the Edge Registry should persist the created entity to disk. Defaults to <c>true</c>.
     /// </summary>
     /// <remarks>
--- a/dotnet/src/Azure.Iot.Operations.Services/EdgeRegistry/CreateOptions.cs
+++ b/dotnet/src/Azure.Iot.Operations.Services/EdgeRegistry/CreateOptions.cs
@@ -1,0 +1,24 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Azure.Iot.Operations.Services.EdgeRegistry;
+
+/// <summary>
+/// Options that control the behavior of a create operation.
+/// </summary>
+public class CreateOptions
+{
+    /// <summary>
+    /// The default options, which persist the created entity.
+    /// </summary>
+    public static CreateOptions Default => new();
+
+    /// <summary>
+    /// Whether the Edge Registry should persist the created entity to disk. Defaults to <c>true</c>.
+    /// </summary>
+    /// <remarks>
+    /// Persistence requires a broker deployed with persistence enabled; it is otherwise ignored. An
+    /// entity that has been persisted can't later be made non-persistent.
+    /// </remarks>
+    public bool Persist { get; set; } = true;
+}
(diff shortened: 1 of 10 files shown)


dotnet/src/Azure.Iot.Operations.Services/EdgeRegistry/ICoreClient.cs
```diff
@@ -60,10 +60,11 @@ public interface ICoreClient : IAsyncDisposable
...
     /// <param name="groupId">The Group. Use <see cref="GroupId.CloudDefault"/> for the cloud-default Group (the configured namespace).</param>
     /// <param name="attributes">The attributes of the Group to create.</param>
+    /// <param name="options">The <see cref="CreateOptions"/> that control the behavior of the create operation; when <see langword="null"/>, <see cref="CreateOptions.Default"/> is used.</param>
     /// <param name="timeout">The command timeout; when <see langword="null"/>, the client's default timeout is used.</param>
     /// <param name="cancellationToken">A token to cancel the operation.</param>
     /// <returns>A task whose result is the created <see cref="Models.CoreGroupEntity"/>.</returns>
-    Task<Models.CoreGroupEntity> CreateGroupAsync(string groupType, GroupId groupId, Models.CoreGroupAttributes attributes, TimeSpan? timeout = null, CancellationToken cancellationToken = default);
+    Task<Models.CoreGroupEntity> CreateGroupAsync(string groupType, GroupId groupId, Models.CoreGroupAttributes attributes, CreateOptions? options = null, TimeSpan? timeout = null, CancellationToken cancellationToken = default);
```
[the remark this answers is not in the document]
timtay-microsoft: Feel free to ignore this one since this feature is still in preview so we can make breaking changes like that

dotnet/src/Azure.Iot.Operations.Services/EdgeRegistry/CreateOptions.cs
```diff
@@ -0,0 +1,24 @@
...
+
+/// <summary>
+/// Options that control the behavior of a create operation.
+/// </summary>
+public class CreateOptions
+{
+    /// <summary>
+    /// The default options, which persist the created entity.
+    /// </summary>
+    public static CreateOptions Default => new();
```
timtay-microsoft: You probably don't need this since the default constructor would behave the same way.

```csharp
new CreateOptions();
```

```csharp
CreateOptions.Default();
```

maximsemenov80: ``Models.DeleteOptions`` has the identical ``public static DeleteOptions Default => new();``
 should I fix that too?

timtay-microsoft (approved):

dotnet/src/Azure.Iot.Operations.Services/EdgeRegistry/EdgeRegistryClient.cs
```diff
@@ -87,6 +91,20 @@ private static Dictionary<string, string> ExtensionVersionTopicTokens(string res
...
 
+    /// <summary>Builds the request metadata for a create request, signalling persistence when requested.</summary>
+    private static CommandRequestMetadata PersistenceMetadata(CreateOptions? options)
+    {
+        CommandRequestMetadata requestMetadata = new();
+
+        // Only sent when true, since sending "false" is equivalent to omitting the property.
+        if (options is null || options.Persist)
+        {
+            requestMetadata.UserData.TryAdd(PersistUserProperty, "true");
```
timtay-microsoft, line 102: It sounds like the fix here involves us adding a "retain" flag here as well so that when this command request metadata reaches the CommandInvoker class, the command invoker can set the retain flag appropriately:

```csharp
                MqttApplicationMessage requestMessage = new(requestTopic, MqttQualityOfServiceLevel.AtLeastOnce)
                {
                    ResponseTopic = responseTopic,
                    CorrelationData = requestGuid.ToByteArray(),
                    MessageExpiryInterval = (uint)reifiedCommandTimeout.TotalSeconds,
                    Retain = //TODO
                };
```

timtay-microsoft: You'll want to extend that fix to the other places where we could set this "aio-persistence" flag like in the DSS client and even the telemetrySender

PR #1478, opened 2026-08-28
[.NET][Services][Connector][EdgeRegistry] Report connector schemas through Edge Registry

maximsemenov80:
Mirrors Rust #1473. Implementing IEdgeRegistryMessageSchemaProvider makes the connector register message schemas as Edge Registry Schema Versions instead of Schema Registry schemas; either way the reported MessageSchemaReference is what the worker caches and puts on the CloudEvent. VersionXId.Parse recovers the registry namespace from the created Version's XID.

### DIFF
--- a/dotnet/src/Azure.Iot.Operations.Connector/ConnectorWorker.cs
+++ b/dotnet/src/Azure.Iot.Operations.Connector/ConnectorWorker.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System.Collections.Concurrent;
+using System.Globalization;
 using System.Text;
 using Azure.Iot.Operations.Connector.CloudEvents;
 using Azure.Iot.Operations.Connector.ConnectorConfigurations;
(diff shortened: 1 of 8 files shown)


dotnet/src/Azure.Iot.Operations.Connector/ConnectorWorker.cs
```diff
@@ -499,44 +465,22 @@ internal async Task ForwardReceivedEventAsync(string deviceName, Device device,
...
-                            eventMessageSchema.SchemaContent,
-                            eventMessageSchema.SchemaFormat,
-                            eventMessageSchema.SchemaType,
-                            eventMessageSchema.Version ?? "1",
-                            eventMessageSchema.Tags,
-                            cancellationToken: cancellationToken);
-
-                        _logger.LogInformation($"Registered message schema for event with name {assetEvent.Name} on asset with name {assetName} associated with device with name {deviceName} and inbound endpoint name {inboundEndpointName}.");
+                // This may register a message schema that has already been uploaded, but the registry service is idempotent
+                MessageSchemaReference? registeredSchemaReference = _edgeRegistryMessageSchemaProvider != null
```
timtay-microsoft, line 472: If I'm reading this correctly, your change makes is so that, if a user provides an edge registry message schema provider, we never send anything to the old schema registry service? If so, that sounds different from what @vaava described to us. 

What I understood from her was that we should continue sending the message schema to the older schema registry service regardless.

maximsemenov80: good call, .NET code matches Rust's PR#1473 behavior, I dug with Copilot into the question a bit and it looks like it might be a miss on the spec part, I would need help from someone with better understanding of the SR/ER flow.
Questions now:

1. Should the connector write to both registries during the transition, or does Edge Registry replace Schema Registry once available?
2. If both: ADR accepts only one MessageSchemaReference. Which one do we report (if both then likely new URI schema needed)?
3. What is aio-sr://{namespace}/{name}:{version} expected to resolve against? If it resolves via Schema Registry, an ER-only write leaves it dangling (if we don't add new URI schema).

timtay-microsoft: As discussed offline, the above approach does mirror what rust does. I'll resolve this thread.

timtay-microsoft (approved):

PR #1494, opened 2026-09-08
[.NET] Re-add persistence support for edge registry client, fix bug around not setting retain flag on persisted messages

timtay-microsoft:
#1477 but with a necessary bug fix for the case where a command or telemetry sets the aio-persistence flag. 

The main issue hidden in the original PR was that the AIO MQTT broker would reject any publish with the "aio-persistence" user property flag set, but without the publish's retain flag set. As a result, our session client would fail to deliver these kinds of messages (and would retry indefinitely)

Now, when the command request/outgoing telemetry metadata objects will use the "PersistCommand"/"PersistTelemetry" flag to communicate to the command invoker/telemetry sender to set both the "aio-persistence" user property and the retain flag on the corresponding MQTT publish packet. This addresses the first part of the issue described above. 

Fixing the session client to not retry on cases like the above will be a separate fix.

### DIFF
--- a/dotnet/src/Azure.Iot.Operations.Protocol/RPC/CommandRequestMetadata.cs
+++ b/dotnet/src/Azure.Iot.Operations.Protocol/RPC/CommandRequestMetadata.cs
@@ -82,6 +82,11 @@ namespace Azure.Iot.Operations.Protocol.RPC
         public ExtendedCloudEvent? ExtendedCloudEvent { get; internal set; }
 
         /// <summary>
+        /// The flag whether this command request will be persisted using the AIO-specific persistence mechanism or not. Only works for AIO MQTT broker.
+        /// </summary>
+        public bool PersistCommand { get; set; }
+
+        /// <summary>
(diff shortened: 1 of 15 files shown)


dotnet/src/Azure.Iot.Operations.Protocol/RPC/CommandRequestMetadata.cs
```diff
@@ -81,6 +81,11 @@ public class CommandRequestMetadata
         /// </remarks>
         public ExtendedCloudEvent? ExtendedCloudEvent { get; internal set; }
 
+        /// <summary>
+        /// The flag whether this command request will be persisted using the AIO-specific persistence mechanism or not. Only works for AIO MQTT broker.
+        /// </summary>
+        public bool PersistCommand { get; set; }
```
timtay-microsoft, line 87: We already have a corresponding "PersistTelemetry" flag in the outgoing telemetry metadata object, so this just mimics that

maximsemenov80 (approved):

maximsemenov80 (approved):

PR #1498, opened 2026-09-12
[NET] : Impl and happy path e2e

olivakar:


timtay-microsoft (approved):
Overall design looks good and seems consistent with the ADR you wrote. I'll happily sign off again after you address the copilot comments

timtay-microsoft (approved):

